PR #17, opened 2015-10-01
Use the symbolic name for the build directory, don't hardcode it

mr-c:


rob-p:
Thanks @mr-c !

mr-c:
Not a problem, thanks for the merge.

PR #23, opened 2015-10-29
Update CMakeLists.txt

lgautier:
xz-5.2.0 is no longer available on tukaani.org.

More generally, this is the source of future issues (next one with the next release of xz, since no archiving of previous releases on tukaani.org)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,11 +190,11 @@ if (NOT LIBLZMA_FOUND)
     message ("=======================================")
 ExternalProject_Add(liblzma
     DOWNLOAD_DIR ${CMAKE_CURRENT_SOURCE_DIR}/external
-    URL http://tukaani.org/xz/xz-5.2.0.tar.gz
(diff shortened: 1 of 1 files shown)


rob-p:
Thanks, @lgautier! I think that ideally, we will create a github repo for all dependencies, so that we can control such availability issues.

lgautier:
I just issued an other pull request: boost_1_57 is no longer available on the download site.

lgautier:
Oops. Just saw that `master` is already pointing at boost 1_59.

What branch should one use for a new install ? master, 0.5.0 ?

rob-p:
Hi Laurent,

  I typically like to keep master compilable, which is why I pushed the boost 1.59 change after 1.57 became available on the previous host.  If you're going to be compiling from source, v0.5.0 should be fine if you're using a system boost (and so don't have to fetch and compile one through CMake).  Otherwise, it should be OK to compile from master, which currently doesn't have any substantial changes from v0.5.0.

--Rob

Issue #24, opened 2015-10-31
Error compiling salmon on Ubuntu 14.04


ctb:
After starting a blank Ubuntu machine, and then executing:

```
sudo apt-get update && \
sudo apt-get -y install screen git curl gcc make g++ python-dev unzip \
         default-jre pkg-config libncurses5-dev r-base-core r-cran-gplots \
         python-matplotlib python-pip python-virtualenv sysstat fastqc \
         trimmomatic bowtie samtools blast2
sudo apt-get -y install cmake libboost-all-dev liblzma-dev

curl -O -L https://github.com/COMBINE-lab/salmon/archive/v0.5.0.tar.gz

tar xzf v0.5.0.tar.gz
cd salmon-0.5.0
cmake .
make
```

gives

```
In file included from /home/ubuntu/salmon-0.5.0/include/BAMQueue.hpp:146:0,
                 from /home/ubuntu/salmon-0.5.0/include/AlignmentLibrary.hpp:14,
                 from /home/ubuntu/salmon-0.5.0/src/SalmonUtils.cpp:13:
/home/ubuntu/salmon-0.5.0/include/BAMQueue.tpp: In function 'bool checkProperPairedNames_(const char*, const char*, uint32_t)':
/home/ubuntu/salmon-0.5.0/include/BAMQueue.tpp:247:33: error: 'BOOST_LIKELY' was not declared in this scope
     if (BOOST_LIKELY(nameLen > 1)) {
                                 ^
```

I'm not sure if this is a Boost version requirement, or what -- looks like I have boost 1.54 installed.

Suggestions welcome!

ctb:
Looks like it's not in 1.54 - http://www.boost.org/doc/libs/1_54_0/libs/config/doc/html/boost_config/boost_macro_reference.html

ctb:
Added

```
#ifndef BOOST_LIKELY
#define BOOST_LIKELY(x) (x)
#endif

#ifndef BOOST_UNLIKELY
#define BOOST_UNLIKELY(x) (x)
#endif
```

to include/SalmonMath.hpp and life is good (or at least salmon compiled ;).

rob-p:
Thanks, Titus!  I'll add this upstream.

ctb:
See https://github.com/COMBINE-lab/salmon/pull/25.

PR #25, opened 2015-10-31
fix for old Boost version

ctb:
Fixes https://github.com/COMBINE-lab/salmon/issues/24 - although this may not be the right fix, I'm dropping it in as a PR just in case it actually is :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@
 *.exe
 *.out
 *.app
+
+# Emacs save files
+*~
--- a/include/SalmonMath.hpp
+++ b/include/SalmonMath.hpp
@@ -1,6 +1,26 @@
 #ifndef SALMON_MATH_HPP
 #define SALMON_MATH_HPP
 
+// If we have built-ins, do as Boost does
+
+#ifndef BOOST_LIKELY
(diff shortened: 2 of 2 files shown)


rob-p:
This seems a reasonable thing to do, but I wonder if we should just re-introduce the actual Boost implementation here since it's so short.  That is:

```
// If we have built-ins, do as Boost does

#ifndef BOOST_LIKELY
#if defined(__has_builtin)
#if __has_builtin(__builtin_expect)
#define BOOST_LIKELY(x) __builtin_expect(x, 1)
#endif
#endif
#endif

#ifndef BOOST_UNLIKELY
#if defined(__has_builtin)
#if __has_builtin(__builtin_expect)
#define BOOST_UNLIKELY(x) __builtin_expect(x, 0)
#endif
#endif
#endif

// If we didn't have those built-ins fall back to this

#ifndef BOOST_LIKELY
#define BOOST_LIKELY(x) (x)
#endif

#ifndef BOOST_UNLIKELY
#define BOOST_UNLIKELY(x) (x)
#endif
```

ctb:
@rob-p, I added & this works for me on Ubuntu 14.04. So +1 from me.

Issue #58, opened 2016-05-11
Gibbs and Bootstrap issues


vals:
Hi, I was trying to investigate a thing using the Gibbs samples (and bootstraps).

I vaguely recall seeing both issues I'm reporting here somewhere, but I can't find those by searching.

I used Salmon 0.6.0, first with this command

    salmon quant \
    -i /nfs/research2/teichmann/reference/danio-rerio/salmon/zv9_cdna_ercc_gfp_index \
    -l IU \
    -p 4 \
    -1 SRS515267_1.fastq \
    -2 SRS515267_2.fastq \
    -o /tmp/SRS515267_salmon_out \
    --biasCorrect \
    --useFSPD \
    --numGibbsSamples 100 \
    && mv /tmp/SRS515267_salmon_out ."

Then I converted the Gibbs samples bootstrap files to a TSV with the script in this repository.

The problem is that I get rather nonsensical values in the Gibbs samples (attaching a screenshot of my notebook because that is the quickest way...)

![image](https://cloud.githubusercontent.com/assets/668803/15195757/2ccadcd4-17c1-11e6-8457-d8540cd59471.png)

What I vaguely recall is that the Gibbs samples does not work with one of the bias modelling options?

I think there should at least be a warning in the documentation entry for `--numGibbsSamples`

After this I ran the same command, but replaced `--numGibbsSamples` with `--numBootstraps`.

Now I get this:

![image](https://cloud.githubusercontent.com/assets/668803/15195925/f81cd64e-17c1-11e6-9c5b-d7a29e2e5bc9.png)

The values seem all right, except I'd prefer the values to be TPM's rather than counts. Is the transformation I'm doing in "line 78" fine for converting the counts back to TPM? Or is it more complicated than that?

rob-p:
Hi Valentine,

  Thanks for reporting this!  Do you have any recollection of what option might be incompatible with Gibbs?  Specifically, it's interesting that it doesn't work with Gibbs (it looks like you're getting `std::numeric_limits<double>::min()` everywhere) but works with Bootstrapping.  I will certainly look into this and will document (and make an explicit warning) any known incompatibilities in the next release.  I'd very much like to fix the issue with Gibbs sampling, as it's substantially faster than bootstrapping for a large number of samples.  Actually, there's a fairly substantial upgrade to the bias modeling that is just about ready and will be part of the next release (which is, in fact, what has delayed v0.7.0 for so long).

[...]

PR #59, opened 2016-05-14
Check sampling method in sampling file conversion script

vals:
As pointed out in #58 the format of Gibbs samples file and Bootstrap samples file is different. Here I check the `meta_info.json` file for which format should be read.


rob-p:
... and, I'm exceedingly embarrassed that I missed this in my own script :(.  Thanks for pointing this out and submitting a fix!

PR #68, opened 2016-08-05
Add EffectiveLength back to documentation.

mdshw5:
The EffectiveLength column is still in the output file, and it looks like it was included in [older documentation](http://salmon.readthedocs.io/en/nb/file_formats.html#quantification-file) but is now removed. Should it be added back?


rob-p:
Yup, this was probably a merge mistake.  The `EffectiveLength` should still be in there, thanks!

PR #92, opened 2016-09-15
Add CONDA_BUILD flag.

roryk:
Hi Rob,

I took a shot at adding a CONDA_BUILD flag to make it work with conda builds rather than editing the CMakeLists.txt with sed in the build script. I tested it and it seems to work, but it could use a review by someone who has used cmake before.

It might be that editing with sed is a cleaner solution, getting the flags in the right place made the CMakeLists.txt file a little ugly. Totally happy to do either solution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,7 +176,12 @@ endif ( DEFINED CUSTOM_BOOST_PATH )
 ##
 # We want static, multithreaded boost libraries
 ##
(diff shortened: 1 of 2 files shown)


rob-p:
Sorry it took me so long to get around to this.  Thanks for the PR!  I merged into develop instead of master since I'm working on a new release.

roryk:
Sweet, thanks Rob!

PR #158, opened 2017-09-28
Docker Improvements

kurtwheeler:
Fixes Dockerfile to be FROM ubuntu:16.04, fixes a checksum in CMakeLists.txt, and makes build_test.sh executable.

I tested this by changing the Dockerfile to copy the code into the image rather than pulling the code directly from github so that I could test the changes I made. I then built the image and ran:
```
$ docker run combinelab/salmon:0.8.2 salmon --version
version : 0.8.2
```


rob-p:
Ignore travis-ci --- it's drunk (has been broken for the past week or so, without any good reason; I think they updated images or some such).

PR #159, opened 2017-10-02
Factorization develop

mohsenzakeri:
The changes (other than adding the range factorization) are:

1- Number of minimum EM operation are set to 100. (works better for small samples of factorization paper)
2- The {:f} option for writing out "count" values are removed. This makes the Spearman correlation of RSEM simulated data 0.8245 in RF mode. (compared to 0.8232 when using the {:f} option)


rob-p:
After discussing my previous comment with @mohsenzakeri, I think I'm onboard with all of the changes from this branch.

PR #199, opened 2018-02-22
scripts fetchRapMap.sh: Fail if any command fails

sjackman:
In particular, fail when unzip is not available.

### DIFF
--- a/scripts/fetchRapMap.sh
+++ b/scripts/fetchRapMap.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -eu -o pipefail
 
 exists()
 {


sjackman:
I noticed that both Drone and Travis CI checks failed. All good?

I've used TravisCI and CircleCI, but I haven't used DroneCI. Do you like it?

rob-p:
Drone is awesome!  The problem seems to be related to the ancient image on which we do CI (holy build box) having a version of curl that, just now (in the last day?), became incompatible with github's OpenSSL protocol.  I guess they did an update, and now my version of curl is too old.  I think we need to update the image.

Issue #272, opened 2018-08-19
[Proposal] Travis CI to add test case easily


junaruga:
Dear salmon project.

I have a proposal.

Right now, the Travis CI test has 1 case for gcc-7.
I thought if we could add and remove test cases easily such as latest version gcc-8, minimal support version gcc-5, clang, and Mac OSX, we would be happy to contribute the source code.

Look at the RNA Sequence tool: Trinity's case. [1][2]
The structure of the `.travis.yml` is to add the test cases easily. It has gcc-N and Mac OSX case.
And I prepared it for this salmon project.
I would like to send pull-request soon after now.

[1] https://travis-ci.org/trinityrnaseq/trinityrnaseq
[2] https://github.com/trinityrnaseq/trinityrnaseq/blob/master/.travis.yml

PR #273, opened 2018-08-19
Update .travis.yml to add test case easily.

junaruga:
This is related to https://github.com/COMBINE-lab/salmon/issues/272

* Now there are 2 test cases on Travis CI, adding one case.
* Added code for speed up such as `git - depth: 1`, `make -j 4`.
* Refactoring. Removed tailing spaces.

Notes:
* I added `gcc-5`'s case as maybe it is minimal support version of gcc for this project.
* I want to replace current case `gcc-7` to latest version `gcc-8` if you like.
* Total running time ("Ran for" on the Travis page.) becomes faster than current situation. In my repository, total running time is "15 min 11 sec", seeing current master branch's test is around 17 min+. See [1]
* Removed `travis_wait`. Without the command, the default behavior is "when a long running command or compile step regularly takes longer than 10 minutes without producing any output". [2] I have not faced the situation when I  did debug. 
* There are commented out area at the bottom of `.travis.yml`. However as we can run `git log .travis.yml` to check past modification, shall we remove the commented out "whitelist" area?

[1] https://travis-ci.org/junaruga/salmon/builds/417976633
[2] https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,49 +1,49 @@
+dist: trusty
 sudo: false
-language: cpp
-
+git:
+  depth: 1
+language: generic
 matrix:
   include:
-    - os : linux
-      compiler : gcc-7
-      addons:
-        apt:
-          update: true
-          sources:
-              - sourceline: 'ppa:mhier/libboost-latest'
-              - ubuntu-toolchain-r-test
-          packages: ['g++-7', 'gcc-7', 'zlib1g-dev', 'libbz2-dev']
+    - env: CC=gcc-7 CXX=g++-7 AR=gcc-ar-7 NM=gcc-nm-7 RANLIB=gcc-ranlib-7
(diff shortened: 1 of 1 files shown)


rob-p:
Hi @junaruga,

  This is *phenomenal*, and I imagine it will be very useful in trying to determine which issues we might expect around a broader range of compilers.  Thank you so much for doing this!  

--Rob

rob-p:
I'm rebasing this since it's easier to put it directly into develop (which is merged into master prior to every release) than to pull into master and cherry pick.  However, since this is just relevant to a single file, I might cherry pick to master as well anyway ;P.

junaruga:
@rob-p thank you for the merging!

Okay, maybe when I want to send a pull-request, it looks better to `develop` branch in general.

How about below questions? Do you agree?

> * I want to replace current case gcc-7 to latest version gcc-8 if you like.

> * There are commented out area at the bottom of .travis.yml. However as we can run git log -p .travis.yml to check past modification, shall we remove the commented out "whitelist" area?

Issue #269, opened 2018-08-16
Alevin: Adding support for CEL-Seq


PeteHaitch:
I've just started working in a single-cell genomics core facility, **alevin** looks really useful for our 10X runs! 
But I also have a substantial number of samples processed with the CEL-Seq and CEL-Seq2 protocols (also 3'-tag protocols). 
I'm interested in adding support for these protocols to Alevin.

1. Is that something you'd be interested in incorporating?
2. Any hints on where to start messing around in the code to implement this would be much appreciated!

Cheers,
Pete

PeteHaitch:
Ah, I just saw https://github.com/COMBINE-lab/salmon/issues/247. I'll experiment with this first.

k3yavi:
Hi @PeteHaitch ,
Thanks for your interest in *Alevin*.
Although in current Alevin we have concentrated mainly on learning more about Droplet based 3'-tagged single cell protocols, especially 10x; we are very much interested in extending it towards other protocols like CEL-seq.
However, there are couple of challenges/difference which should be considered before incorporating it into the Alevin pipeline. Currently Alevin relies on the fact that the droplet based protocols use PCR amplification of the library and the UMI deduplication phase of Alevin assumes an exponential model, I am not sure how true is this with CEL-seq? Another issue is that CEL-seq is a Fluidigm based system while the current application for Alevin is for microfluidics based. In general we have observed that the 10x cell isolation step is pretty robust in reporting the Cellular Barcodes(CB) and although we have a probabilisitic model to handle the CB based uncertainty but the ambiguous case like that are very less frequent, (although not true for Drop-Seq). Having said that, we might have to do some analysis to actually figure out the right model for Barcode correction in Fluidigm based system.

Also, please do let us know of your experience in using the solution proposed in #247 . Looking forward to hearing back from you.

PeteHaitch:
Toying around with the solution in #247, I think I've found why it's not yet working for me.
The CEL-Seq read1 is UMI + CB whereas 10X is CB + UMI (to my understanding). Is there an easy way to tell **alevin** the order is reversed?

k3yavi:
Hey @PeteHaitch ,
[...]

PR #285, opened 2018-09-05
Tweaks to CEL-Seq2 support for Alevin

PeteHaitch:
A PR noting the differences in my approach to #269.
The one that I think needs addressing is in `writeFastq` (https://github.com/COMBINE-lab/salmon/compare/develop...PeteHaitch:develop?expand=1#diff-bf2f37cd9ea77a5c454a5bd860a924ee); without some change to this the `UMI` and `CB` are incorrectly extracted for CEL-Seq2. 
I simply commented out the original lines and modified it as needed for CEL-Seq2 in order to test my modified version.
I guess some sort of protocol-specific conditional is needed here.

The remainder are minor/cosmetic choices of variable names (please feel free to ignore!).

### DIFF
--- a/src/Alevin.cpp
+++ b/src/Alevin.cpp
@@ -549,9 +549,15 @@ bool writeFastq(AlevinOpts<ProtocolT>& aopt,
       for (size_t i = 0; i < rangeSize; ++i) { // For all the read in this batch
         auto& rp = rg[i];
         if(aopt.protocol.end == bcEnd::FIVE){
-          barcode = rp.first.seq.substr(0, aopt.protocol.barcodeLength);
-          umi = rp.first.seq.substr(aopt.protocol.barcodeLength,
-                                    aopt.protocol.umiLength);
+          //barcode = rp.first.seq.substr(0, aopt.protocol.barcodeLength);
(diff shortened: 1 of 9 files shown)


src/Alevin.cpp
```diff
@@ -549,9 +549,15 @@ bool writeFastq(AlevinOpts<ProtocolT>& aopt,
...
-          barcode = rp.first.seq.substr(0, aopt.protocol.barcodeLength);
-          umi = rp.first.seq.substr(aopt.protocol.barcodeLength,
-                                    aopt.protocol.umiLength);
+          //barcode = rp.first.seq.substr(0, aopt.protocol.barcodeLength);
+          //umi = rp.first.seq.substr(aopt.protocol.barcodeLength,
+          //                          aopt.protocol.umiLength);
+          // TODO: The above isn't valid for CEL-Seq2. I temporarily commented 
+          //       it out and replaced with a CEL-Seq2-specific format so that 
+          //       I could test my changes.
+          barcode = rp.first.seq.substr(aopt.protocol.umiLength,
```
k3yavi: Nice catch, I missed updating this part. You are right we should call the extract barcode and umi functions here.

PR #344, opened 2019-02-15
Add option for using quante.genes.sf and to choose default value

antpiron:
This small patch adds two options to quant merge:
- --genes to force use of quant.genes.sf instead of quant.sf ;
- --missing <arg> to change the default value in case of missing value.


rob-p:
Thanks for the PR, @antpiron! I'm just changing the base branch from `master` to `develop` before the merge.

PR #351, opened 2019-03-16
bug/typo in previous version

gaberoo:
Sorry, this was my mistake.


rob-p:
The Travis CI build is failing now.  It seems to not be able to find the 64-bit version of the divsufsort library.

rob-p:
w00t; all looks good here :).

PR #391, opened 2019-07-03
added --chromiumV3 to necessary input parameters

jdrnevich:
This doc didn't mention --chromiumV3 anywhere (I found it while looking at  salmon alevin -h  and then also in the additional tutorials. This also should be linked in the salmon.readthedocs.io somewhere...


k3yavi:
Thanks @jdrnevich for the heads up, I'll update the docs too.

PR #464, opened 2019-12-19
Lenght -> Length

mdshw5:
I know this PR may be controversial, but I believe it can be merged after careful review.


rob-p:
After much thought and consideration, I think this can be merged!

PR #515, opened 2020-04-28
Code for producing gibbs samples

hiraksarkar:
Implementing a simple Gibbs sampler within Alevin to accept the option `--numCellGibbsSamples` and produce the `mean`, `variance`, and Gibbs estimates in the same format as of `--numCellBootstraps`.

The default `thinning factor` is set to be `16`. 

Additionally, we add `--dumpCellEq` to dump the deduplicated gene-level equivalence classes per cell.


src/CollapsedCellOptimizer.cpp
```diff
@@ -449,6 +664,18 @@ void optimizeCell(std::vector<std::string>& trueBarcodes,
...
+        labelsEq.resize(salmonEqclasses.size());
+        countsEq.resize(salmonEqclasses.size());
+        for(size_t salEqId = 0 ; salEqId < salmonEqclasses.size(); ++salEqId){
+          labelsEq[salEqId] = salmonEqclasses[salEqId].labels;
+          countsEq[salEqId] = salmonEqclasses[salEqId].count;
+        }
+        gzw.writeDedupCellEQVec(trueBarcodeIdx, labelsEq, countsEq, true);
+      }
+
       if ( numBootstraps and noEM ) {
```
k3yavi, line 679: Add numGibbs here, just cross check if we have any other cases like this.
Another thing to check here or in alevinUtils would be that both gibbs and bootstaps are not provided simultaneously to the command line

hiraksarkar: This particular place is for dumping gene-level equivalence classes. Nothing to do with Gibbs. Sorry if I missed what you meant.

k3yavi: Huh ? I meant line 679. Basically checka if noEM is provided along with numBootstraps it gives error. We might wanna add numGibbs here and check similar such cases if exist.

hiraksarkar: Gotcha! done

PR #552, opened 2020-07-28
add missing include

mr-c:
This was necessary to build on Debian with gcc/g++ 10.2.0


rob-p:
Awesome. Thank you @mr-c; and thanks for basing off of develop :)!

PR #555, opened 2020-08-03
add missing <string> header

CreRecombinase:
In recent versions of gcc (10.1.0) <unordered_map> doesn't pull in <string>, so you need to add that.


CreRecombinase:
Sorry just saw that there's an identical outstanding pull request. Sorry!

PR #617, opened 2021-01-21
Ont

gmarcais:
Add an error model specific to Oxford Nanopore long sequencing reads. Added a switch `--ont` that triggers the use of this error model (this implies `--noLengthCorrection`).

The error model assumes IID error rate along the reads and a geometric length distribution of clipping at both ends of the reads. (valid after trimming the reads with Oxford's own trimmer).


rob-p:
Here goes nothing @gmarcais!

PR #703, opened 2021-09-16
Add inDrop v2 to alevin

Gaura:
Add [inDropv2 protocol](https://www.nature.com/articles/nprot.2016.154) to salmon alevin. To use this protocol pass the `--indropV2` flag. The details of library preparation can be seen [here](https://teichlab.github.io/scg_lib_structs/methods_html/inDrop.html). R1 is the biological read and R2 is the metadata read. Briefly, the reads have 8-11 bp barcode1, followed by a `W1` sequence which should be specified using `--w1` flag then 8bp barcode2 and 6 bp UMI sequence follows. 

Summary of changes:
1. `--indropV2` flag added to process sci-rna-seq3 reads
2. a local copy of single cell protocol is created before barcodes and UMI are extracted to allow share the positions and avoid searching twice
3. In case the barcode1 is less than 11bp, `A` is added after joining barcodes 1 and 2 to make the total length 19 bp
4. If no exact match for w1 is found, a search for w1 with a hamming distance <=2 is performed

It was tested for correlation on run SRR7165069 from GEO submission [GSM3141960](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM3141960).
Results:
image

```
  Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
 0.3634  0.8981  0.9655  0.9307  0.9874  0.9976 
```


k3yavi:
That's amazing @Gaura. This feature has been frequently requested by user but I never got a chance to work on this, thanks a lot the PR. Give me some time to go over the PR and if everything looks Ok we can merge it in into the next release cycle. May I ask previous version of inDrop had an issue with variable length bacrcodes, did they solved that issue in v2 ?

Gaura:
Thanks @k3yavi! I'm not aware of the issues with previous version. I tested it with the [test data](https://github.com/indrops/indrops/tree/master/test/seq_runs/run_v2_single_file) and the data I mentioned earlier and it worked fine. Let me know how the review goes and if I have missed something.

src/AlevinUtils.cpp
```diff
@@ -378,21 +378,35 @@ namespace alevin {
...
-      bc = read.substr(0, index);
-      if(bc.size()<8 or bc.size()>12){
+      bc = read2.substr(0, pt.w1Pos);
+      if(bc.size()<8 or bc.size()>11){
         return false;
       }
       uint32_t offset = bc.size()+pt.w1.size();
-      bc += read.substr(offset, offset+8);
+      bc += read2.substr(offset, 8);
+      bc += std::string(pt.barcodeLength - bc.size(), 'A');
```
rob-p, line 404: @Gaura : Do we want to consider "unique padding" just as we did with sci-seq3?  Specifically, we are lucky here in that there are 4 distinct lengths we could have so we could just pad with the 4 distinct nucleotides.  That is, make the barcode length 12 and then:

* If length is 11, add `A`
* If length is 10, add `AC`
* If length is 9, add `AAG`
* If length is 8, add `AAAA`

Or some such.  I'm not sure how much of a difference it would make, but it would be nice potentially to be able to say we don't introduce any collisions this way.

src/AlevinUtils.cpp
```diff
@@ -378,21 +378,35 @@ namespace alevin {
...
-      if (index == std::string::npos){
-        return false;
+                                     apt::InDropV2& pt, std::string& bc){
+      (void)read;
+      if(read2.length() >= (pt.w1Length + pt.barcodeLength + pt.umiLength)) {
+      pt.w1Pos = read2.find(pt.w1);
+      if (pt.w1Pos == std::string::npos){
+        bool found = false;
+        for( int i = 8; i <= 11; i++){
+          if (hammingDistance(pt.w1, read2.substr(i,pt.w1Length)) <= pt.maxHammingDist) {
```
rob-p, line 390: @Gaura :  I think this current behavior is OK, but we should be careful to document it.  Specifically, the approach to finding a barcode as implemented here seems to be:

Start at offset 8 and scan through offset 11, accepting the *first* substring whose Hamming distance to the target `pt.w1` is within the required bound.  This means if there are multiple indices that achieve this bound, we return at the first one found.  Likewise, if one has a distance of 2 but a later one has a distance of 1 or 0, we will return the 2.  This seems an unlikely scenario, but is possible and probably should, at least, be documented.

rob-p (commented):
Hi @Gaura :  This looks really good.  I left 2 comments to address that we can discuss if you'd like.

src/AlevinUtils.cpp
Gaura: Thanks! I have updated this accordingly.

src/AlevinUtils.cpp
Gaura: Great suggestion. I'll make a separate PR with the updated documentation which will include updates on other protocols as well.

rob-p:
Thanks @Gaura!

Issue #699, opened 2021-08-27
Identifying correct library type for SPLiT-seq/ParseBio reads


jeremymsimon:
Hey @rob-p and @k3yavi! Avi and I have chatted a bit about this offline, and [I've commented about this in the past](https://github.com/COMBINE-lab/salmon/discussions/629#discussioncomment-777414), but I'm currently attempting to extend alevin's use to the newer combinatorial indexing experiments like [SPLiT-seq](https://science.sciencemag.org/content/360/6385/176) or the commercial solution, [ParseBio](https://www.parsebiosciences.com/) (the biotech formerly known as SplitBio). Now that you support different barcode geometries and allow the user to specify where these barcodes should be expected, this _should_ be doable. 

I've downloaded the [raw data from the original SPLiT-seq paper](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM3017261), and we have some ParseBio data of our own; my goal is to help you all write a vignette on this once we get the details worked out. 

In short, the R1 file contains cDNA sequence, and R2 contains a 10bp UMI followed by 3 different barcodes in the following arrangements (note: ParseBio seems to shift BC1 up a few bp, to 79-86):
  - UMI: 1-10
  - BC3: 11-18
  - BC2: 49-56
  - BC1: 87-94

The intervening bases are a combination of common and variable sequences.

The other wrinkle with these approaches is that the cDNA can be amplified using either oligo-dT or random hexamer barcodes. The net effect of this is that BC1 can be one of two options, and some barcode combinations need to be pooled to represent the same cell. For example:

AACGTGAT-CTGTAGCC-ACACAGAA
GATAGACA-CTGTAGCC-ACACAGAA

may be the same cell, amplified by two different means. 

We know how the BC1s are paired, so Avi suggested pre-processing the R2 FASTQ file such that we locate random hexamer BC1 sequences and modify them to the matched oligo-dT sequence. In the above example, we'd correct some R2 reads that contain "GATAGACA" in the 87-94 position to "AACGTGAT". I [wrote a script](https://github.com/jeremymsimon/SPLITseq) (in perl because I'm old-school) that does this and allows for mismatches within 1 hamming distance. I'm sure someone could port this to python, rust, or C/C++ and speed this step up substantially, but it seems to work fine.

The next step is to run alevin. I've done so using salmon v1.5.2 with the following parameters (for a ParseBio run), reversing R1/R2 as per Avi's suggestion:
```
[...]

PR #726, opened 2021-12-08
Add split-seq V1 and V2 to alevin

Gaura:
This PR addresses issue #699. To use the protocol, pass the `--splitseqV2` or `--splitseqV1` flag. To test the implementation correlation analysis was done on [data](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE137941) submitted with the [article](https://www.sciencedirect.com/science/article/pii/S1934590920300552). Thanks @jeremymsimon for pointing to the data. Here are the results of correlation b/w the alevin output and published counts. This test run was done as mentioned in [here](https://github.com/COMBINE-lab/salmon/issues/699#issuecomment-988185505).

image

```
> summary(cr)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
 0.2966  0.7128  0.8690  0.8163  0.9448  0.9963 
```

### DIFF
--- a/include/SalmonDefaults.hpp
+++ b/include/SalmonDefaults.hpp
@@ -139,6 +139,7 @@ namespace defaults {
   constexpr const bool isCITESeq{false};
   constexpr const bool isCELSeq{false};
   constexpr const bool isCELSeq2{false};
+  constexpr const bool isSplitSeqV2{false};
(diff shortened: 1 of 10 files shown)


rob-p (approved):
These changes / fixes look good.

PR #771, opened 2022-04-27
minimize Docker image size by using multi-stage builds

kaczmarj:
This commit uses multi-stage builds to minimize the Docker image size.
The uncompressed size is down to 101 MB from 1.38 GB. The second 
stage of the build begins with a fresh ubuntu:18.04 image and copies in
the compiled salmon binaries. It also installs one system library that
is linked by one of the salmon shared libraries.


rob-p:
This is brilliant @kaczmarj — thanks!

Issue #813, opened 2022-11-28
Can't find left Boundary


cliftonlewis:
**Is the bug primarily related to salmon (bulk mode) or alevin (single-cell mode)?**
alevin
**Describe the bug**
I am running the following command on some sci-rna-seq3 samples and it seems to not work as expected. 
`salmon alevin -i af_splici/dm6_splici_idx/ -l ISR -1 data/SRR17122012_1.fastq -2 data/SRR17122012_2.fastq -o SRR17122012 --tgMap transcriptome_splici_fl52/transcriptome_splici_fl52_t2g.tsv -p 28 --sciseq3 --justAlign`
I then took the output into alevin-fry to create a generate-permit-list and it gives me the error that salmon hasn't added the extra bps to account for the chemistry
"thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `20`,
 right: `19`: found barcodes of different lenghts 20 and 19', src/cellfilter.rs:203:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace"
Thus I re-ran salmon alevin without the --justAlign flag and it seems to hit a different error
"### alevin (dscRNA-seq quantification) v1.9.0
### [ sciseq3 ] => { }

[2022-11-28 21:13:57.772] [alevinLog] [info] Found all transcripts to gene mappings
[2022-11-28 21:13:57.781] [alevinLog] [info] Processing barcodes files (if Present) 

 
processed 10 Million barcodes

[2022-11-28 21:14:01.454] [alevinLog] [info] Done barcode density calculation.
[2022-11-28 21:14:01.454] [alevinLog] [info] # Barcodes Used: 1 / 10285890.
[2022-11-28 21:14:01.455] [alevinLog] [error] Can't find left Boundary.
Please Report this issue on github."

Specifically, please provide at least the following information:

  * Which version of salmon was used? salmon (1.9.0); alevin-fry (0.8.0) 
  * How was salmon installed (compiled, downloaded executable, through bioconda)? Conda install
  * Which reference (e.g. transcriptome) was used? Drosophila melanogaster (BDGP6.32 (GCA_000001215.4))
  
**Desktop (please complete the following information):**
 - OS: Ubuntu Linux

Thanks in advance for all your help regarding this!

Gaura:
@cliftonlewis: could you tell us version of alevin-fry are you using? 
[...]

PR #817, opened 2022-12-09
counter artificial increase in barcode length

Gaura:
Subtract 2 to counter artificial increase in barcode length. Read1 length of 33 or 34 would result in error while being valid lengths.


Gaura:
fixes #813

rob-p (approved):
Hi @Gaura, I'm going to approve and merge this now because I need to cut a new release.  However, we should definitely loop back around and see if we can solve this more generally for all variable length barcode protocols.

PR #856, opened 2023-06-26
Adding <cstdint> includes to allow building with gcc13

A-N-Other:
As per https://gcc.gnu.org/gcc-13/porting_to.html, `gcc13` changes the way it handles includes. These three files needed patching to allow the build to work.


rob-p:
Thanks @A-N-Other!  I changed the base branch to develop, as we generally pull everything through that branch before it makes it to master.  Otherwise, these changes look good.  Hopefully we'll get around to pushing out a patch release with this change soon.

PR #868, opened 2023-08-23
ONT markov model

JuliaHolz:
Updated error model for ONT reads -- uses a markov model to model mismatches, indels, and homopolymer regions. Kmer size and step size for the markov model are set at 50 in ONTAlignmentModel.hpp, but could be adjusted (50 was found to be the best out of the values of k I tried).

### DIFF
--- a/include/ONTAlignmentModel.hpp
+++ b/include/ONTAlignmentModel.hpp
@@ -10,7 +10,7 @@
 
 #include "AlignmentCommon.hpp"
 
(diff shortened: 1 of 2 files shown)


JuliaHolz:
I think I meant to open a PR to Ocean's version of Salmon (oops. I opened a PR there as well.) This should probably eventually be merged into the regular version of Salmon as well, but I should probably wait for the other PR to be reviewed.

PR #916, opened 2024-03-15
update stadenio and htscodecs

pabloaledo:
This PR updates the version of stadenio_libs and htscodecs.
These versions introduce changes that make them more amenable to be compiled for ARM.


rob-p (approved):
Moved base from master to dev to allow following the normal release cycle if this works.  Thanks!

PR #967, opened 2024-10-23
Performance fix: don't let boost promote doubles to long doubles

dslarm:
This fix resolves performance issue where Boost::math unnecessarily promotes doubles to long double, which is not fully supported by hardware leading to slow-downs.

The change is to set a define during cmake process which prevents promotion ( -DBOOST_MATH_PROMOTE_DOUBLE_POLICY=false )


dslarm:
Argh.  I made the last commit on the wrong branch.  Hold on this pull request and I'll replace.

PR #968, opened 2024-10-23
Performance fix: don't let boost promote doubles to long doubles

### LINKED ISSUE #966: Performance issue in use of boost::math::digamma on aarch64 Linux
Due to a current default in the boost library (https://github.com/boostorg/math/issues/1211) in boost::math::digamma, there is a performance hit on aarch64.

This happens on v1.10.3 of Salmon, with GNU compiler 13 on Linux aarch64.

A 4-thread quantization of one of the Salmon tutorials DRR0* series files spends ~15% of time in this routine (called within CollapsedEMOptimizer). On a larger example, we see 7% performance hit over a run that takes 1300 seconds on 4 cores.  On x86 this time is small enough to be lost in the noise.

`salmon quant -i athal_index -l A 
         -1 DRR016125/DRR016125_1.fastq.gz
         -2 DRR016125/DRR016125_2.fastq.gz 
         -p $threads --validateMappings -o quants/DRR016125_quant`

There is a simple fix which is to ensure the CMake/Makefiles ensure salmon compiles with: 
`-DBOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS`
or to add that to any file that brings in boost::math via adding `#define BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS` at the start.

With that change, a 1300 second runtime drops to 1212 for the larger test case, and for the tutorial case is 48 seconds down to 40 on a 4-core r8g.xlarge (Graviton4).

Whilst Boost may fix the issue soon - it's likely that older versions of the library will be found installed for some time.  It would be helpful to add this define to cmake settings, or the sources.

dslarm:
Fixes #966 
This fix resolves performance issue where Boost::math unnecessarily promotes doubles to long double, which is not fully supported by hardware leading to slow-downs.

The change is to set a define during cmake process which prevents promotion ( -DBOOST_MATH_PROMOTE_DOUBLE_POLICY=false )

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -259,6 +259,9 @@ else()
   message(FATAL_ERROR "Your C++ compiler does not support C++14.")
 endif()
 
+
+list(APPEND TGT_COMPILE_FLAGS -DBOOST_MATH_PROMOTE_DOUBLE_POLICY=false)
+
 if(DO_QUIET_MAKE)
   set(QUIET_MAKE "--silent")
 else()


rob-p:
Closing as part of a housekeeping pass — with thanks; this was a real find (silent long-double promotion in Boost.Math is exactly the kind of thing nobody suspects until they profile), and the one-line policy define was the right minimal fix.

The reason for closing rather than merging: the repository's mainline is now the ground-up Rust implementation, which carries neither Boost nor the CMake build this patches, so there is no target for the change. The C++ 1.x line is frozen; anyone building it from source via the v1.10.x tags can apply `-DBOOST_MATH_PROMOTE_DOUBLE_POLICY=false` locally exactly as this PR describes — and this PR remains linkable as the reference for that.

PR #970, opened 2024-10-23
Only select LLVM's libc++ for APPLE platforms. 

### LINKED ISSUE #969: Compilation with LLVM fails on Linux Rocky9/Ubuntu-22.04 - stdlib incorrectly chosen
In CMakeLists.txt, `check_cxx_compiler_flag(-stdlib=libc++ HAVE_LIBCPP)` is used to determine whether to use LLVM's C++ stdlib - libc++, or the system's own GNU one (libstdc++).

Unfortunately, those distros do not ship with that library (per the clip below) - and yet the test succeeds - yielding failure later in the build process due to lots of missing header files.
```
[rocky@ip-10-11-36-33 ~]$ cat >test.cpp
#include <iostream>

int main() {
std::cout << "Helloworld";
}
[rocky@ip-10-11-36-33 ~]$ clang++ -stdlib=libc++ test.cpp 
test.cpp:1:10: fatal error: 'iostream' file not found
    1 | #include <iostream>
      |          ^~~~~~~~~~
1 error generated.
```

The cause is that CMake's `check_cxx_compiler_flag(-stdlib=libc++ HAVE_LIBCPP)` test which is used in Salmon only tests if the compiler accepts the flag (the test code it uses doesn't have any includes - just `main() {return 0;}`) rather than if it works.

I don't know the whys behind the missing library - the library is available on OS/X for example - however it seems this behaviour should be changed for now on Linux platforms.

dslarm:
Linux distros do not include it with their LLVM packages. Fixes #969.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,7 +234,7 @@ elseif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
     # If we have libc++, then try and use it
     include(CheckCXXCompilerFlag)
     check_cxx_compiler_flag(-stdlib=libc++ HAVE_LIBCPP)
-    if(HAVE_LIBCPP)
+    if(HAVE_LIBCPP AND APPLE)
       message("It appears that you're compiling with clang and that libc++ is available, so I'll use that")
       list(APPEND TGT_COMPILE_FLAGS -stdlib=libc++)
       set(BOOST_TOOLSET "clang")


dslarm:
@rob-p - can you merge this one in?

rob-p:
Closing as part of a housekeeping pass — thanks for the fix, and for the pair of build reports it came with. Scoping `-stdlib=libc++` to Apple platforms (where it's the system default) instead of anywhere Clang is detected was the correct call; Linux Clang builds against distro LLVM packages were genuinely broken by the old behavior.

The reason for closing rather than merging: the repository's mainline is now the ground-up Rust implementation, so the CMake toolchain selection this patches no longer exists on the active branch. The C++ 1.x line is frozen; for anyone building the v1.10.x tags from source with Clang on Linux, this PR stands as the documented workaround.

PR #1015, opened 2026-06-17
2.0.2: emit duplicate_clusters.tsv (--keepDuplicates) + mapper allocation/perf improvements

rob-p:
This bundles two independent, self-contained changes staged for 2.0.2.

## 1. Index: emit `duplicate_clusters.tsv` with `--keepDuplicates` (`de542658`)

Detect duplicate clusters in *both* index modes and always write `duplicate_clusters.tsv` (collapse only when `--keepDuplicates` is not set), matching C++ salmon's behavior. On the *T. urticae* decoy-aware test index, the `--keepDuplicates` build now writes the same 36 duplicate pairs C++ reports (identical set) while keeping all 20,675 refs; the non-keepDuplicates build still collapses to 20,639 and writes the same 36 pairs.

## 2. Mapper: reuse ksw2 scratch + two-MEM chaining fast path (`d7d975e8`)

Cut allocator traffic in the alignment/chaining hot path:
- A per-thread `KswScratch` (reusable `ksw2rs::Aligner` + DNA5 encode buffers) replaces per-call `Vec` encodes and `Extz::default()` workspaces across `ksw2_gap_global`, `ksw2_flank_extend`, and `ksw2_align_score` (the score-only ksw2 config is unchanged).
- Exact single-MEM and two-MEM fast paths in `chain_mems` skip the general DP's n-sized scratch buffers for the overwhelmingly common small cases; the two-MEM path defers (`None`) on tie cases so the caller falls back to the general implementation.

Both are score/result-preserving refactors.

## Checks
- `cargo fmt` clean; `cargo clippy --all-targets -- -D warnings` clean.
- `cargo test -p salmon-map -p salmon-index`: 53 tests pass (incl. alignment-validation and end-to-end mapping integration tests).

### DIFF
--- a/crates/salmon-index/src/lib.rs
+++ b/crates/salmon-index/src/lib.rs
@@ -249,8 +249,10 @@ struct RefHashes {
 struct PreprocessResult {
     /// non-ACGT bases replaced with pseudo-random ACGT
     replaced: u64,
-    /// `(retained_name, duplicate_name)` pairs in discovery order; the duplicate
-    /// was *not* written to the cleaned FASTA (so it is absent from the index).
+    /// `(retained_name, duplicate_name)` pairs in discovery order. Without
+    /// `--keepDuplicates` the duplicate was *not* written to the cleaned FASTA (so
+    /// it is absent from the index); with `--keepDuplicates` every transcript is
+    /// written and this just records which ones are exact duplicates of each other.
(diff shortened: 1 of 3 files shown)


rob-p:
Folded into #1017 (the 2.1.0 release branch): duplicate_clusters.tsv emission under --keepDuplicates and the mapper allocation/perf improvements are included there (commits f4cea342, 2eeebcee, etc.).

PR #1020, opened 2026-06-18
fix(quant): parallelize bias-corrected effective-length estimation and skip decoys

### LINKED ISSUE #1019: salmon 2.0.1: --gcBias/--seqBias abundance estimation is single-threaded and never completes on a large decoy-aware index (same input completes in ~2.75 min with 1.12.0)
## Summary

With **salmon 2.0.1** (mapping mode), `salmon quant` finishes a sample in ~1 minute without bias correction, but as soon as `--seqBias` and/or `--gcBias` is enabled the `estimating abundances` phase runs **single-threaded** (one core pinned at 100%, the other 11 idle) and **does not complete** after 45+ minutes, regardless of `-p`. Mapping itself is fast and multithreaded; only the bias-corrected abundance-estimation step is affected.

Crucially, the **same input and the same flags complete in ~2.75 minutes with salmon 1.12.0 (the final C++ release)**, so this looks like a regression in the 2.0 abundance-estimation path rather than an inherent cost of bias modeling.

## Environment

- salmon **2.0.1**, official prebuilt binary, native **arm64**
- Apple **M4 Max**, 16 cores, 128 GB RAM, **macOS 26.5.1**
- Index: GRCh38, Ensembl release 115, **decoy-aware** (gentrome = cdna + ncrna + primary-assembly genome as decoy); **532,811 references**, max_ref_len 248,956,422
- Paired-end bulk RNA-seq, ~36.6M fragment pairs/sample

## Command

```
salmon quant -i index_GRCh38_r115_decoy -l A -1 R1.fq.gz -2 R2.fq.gz -p 12 --seqBias --gcBias -o out
```

## Observed (salmon 2.0.1)

```
mapping reads (selective-alignment mode, 12 threads)
mapped 36286473 / 36602344 fragments (99.14%)      # ~2 min, multithreaded, excellent
estimating abundances (VBEM)                        # then ONE core at 100%, others idle
# ... no quant.sf after 45+ minutes; no further log output
```

- Same single-threaded stall with `--seqBias` alone (no `quant.sf` after >70 min), with `--gcBias` alone, with `--useEM` (instead of VBEM), and with `--biasSpeedSamp 100`.
- **Without** `--seqBias`/`--gcBias`: abundance estimation finishes in seconds, whole sample in ~1 min, 99.14% mapping.

## Comparison with salmon 1.12.0 (C++)

Built a 1.12.0 SSHash decoy-aware index from the **same** gentrome and ran the **same** sample/flags:

```
salmon-1.12.0 quant -i index_v112 -l A -1 R1.fq.gz -2 R2.fq.gz -p 12 --seqBias --gcBias -o out
# estimating abundances + writing output: quant.sf written in ~166 s total  (FINISHED)
```

So the bias-corrected abundance estimation that never completes in 2.0.1 completes in **~2.75 min** in 1.12.0. This is the main reason I suspect a 2.0 regression.

## Questions / request

1. Is the bias-corrected abundance-estimation / GC effective-length step expected to be **single-threaded** in 2.0.1, even with `-p > 1`? Mapping clearly uses all threads; this phase does not.
2. Can this phase be **parallelized** across the `-p` worker threads? It is the dominant cost for large decoy-aware indices and currently makes per-sample bias correction impossible at scale (we have 96 samples).
3. Does the GC effective-length update iterate over all references (decoys included)? If so, could it be restricted to expressed targets?

## Secondary observation (separate from the above)

On this same data, **1.12.0 maps only 35.95%** (2,926,845 fragments "discarded because best-mapped to decoys", plus further score-threshold discards) versus **99.14% in 2.0.1**. So we cannot simply fall back to 1.12.0 for bias-corrected quant: neither version gives both good mapping and a completing bias step. Happy to file this separately if useful.

I can provide full logs, the index `info.json`, or a profiled run.

BenjaminDEMAILLE:
## Summary

Fixes #1019: with `--seqBias` and/or `--gcBias`, the `estimating abundances` phase pinned a single core and never completed on large indexes. For comparison, the reporter measured C++ salmon 1.12.0 completing their original sample (a private ~36.6M fragment-pair dataset) with the same flags in ~2.75 min; that figure is from the issue and is a different dataset than the public reproducer below, which was not separately timed for 1.12.0.

## Root cause

Bias correction builds three "expected" background models from the burn-in abundance estimate. Only one of the three was parallelized:

| function | status before |
| --- | --- |
| `build_expected` (seqBias) | single-threaded `for tid in 0..num_refs` |
| `build_expected_pos` (posBias) | single-threaded `for tid in 0..num_refs` |
| `build_expected_gc` (gcBias) | already parallel |

Each is an O(total transcriptome length) sweep, so the serial ones dominate the abundance phase on a large target set. The reporter confirmed the stall reproduces even on a decoy-free transcriptome (~250k targets), which isolates single-threadedness (not decoy handling) as the load-bearing bug. Separately, building the expected models over `0..num_refs` includes decoy references; decoys are never expressed (the burn-in EM drives their alpha to 0, and the `MIN_ALPHA` guard already skips them), but a decoy chromosome slipping through would hand one rayon worker an O(ref_len) sweep over hundreds of Mbp.

## Fix

Three commits, one per change:

1. **perf(model): parallelize expected sequence-bias model construction** - rayon `fold`/`reduce` over transcripts with per-thread `SBModel` partials merged via the existing `combine_counts` (associative with `SBModel::new()` as identity), mirroring `build_expected_gc`.
2. **perf(model): parallelize expected positional-bias model construction** - same pattern. Positional masses are log-space, where `log_add`'s identity is `-inf` rather than `log(1)`, so per-thread partials use a new `SimplePosBias::new_empty` (no pseudocount) and the single per-bin pseudocount is injected once at the end. This reproduces the serial accumulation exactly (an empty bin still collapses to `log(1)`).
3. **fix(quant): exclude decoy references from expected bias models** - bound all three builds to quantification targets via `num_targets = first_decoy_index` (decoys are always the contiguous tail). Parity-preserving, since a decoy's weight `alpha/effLen` is ~0.

The global rayon pool is already sized by `-p`, so the parallel sweeps scale with the requested thread count automatically.

## Tests

- New unit tests in `seqbias.rs` and `bias.rs` assert the `num_targets` bound is respected (a target beyond it does not contribute) and that a zero-alpha target is skipped either way.
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --release` all pass on the pinned toolchain (1.91.1).

## Verification on the issue's public reproducer

The Ensembl r115 decoy-aware index + Himes airway `SRR1039508`, run with `-p 12 --seqBias --gcBias`, should now complete (write `quant.sf`) and use all threads during `estimating abundances`, where it previously hung.

### DIFF
--- a/crates/salmon-model/src/bias.rs
+++ b/crates/salmon-model/src/bias.rs
@@ -8,7 +8,7 @@
 //! with the conditional fragment-length distribution.
 
 use crate::gcbias::{GcContext, GcFragModel, GcView};
-use crate::posbias::{length_class_index, SimplePosBias, NUM_LENGTH_CLASSES};
+use crate::posbias::{length_class_index, SimplePosBias, NUM_LENGTH_CLASSES, NUM_POS_BINS};
 use crate::seqbias::{
     conditional_cdf, log_bias, revcomp_bytes, SBModel, CONTEXT_LEFT, CONTEXT_LENGTH, MIN_ALPHA,
     MIN_CDF_MASS,
@@ -223,7 +223,7 @@ pub fn corrected_effective_length_full(
 /// reverse density = fragments that can end here). Models are finalized.
 #[allow(clippy::too_many_arguments)]
 pub fn build_expected_pos<FL>(
-    num_refs: usize,
+    num_targets: usize,
     ref_len_of: FL,
     alphas: &[f64],
     eff_lens: &[f64],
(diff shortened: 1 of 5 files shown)


rob-p:
Folded into #1017 (the 2.1.0 release branch): the parallelized bias-corrected effective-length estimation and decoy exclusion are cherry-picked there (commits ea0dffc5, 69265e1d, d18a5796). Thanks @BenjaminDEMAILLE.

PR #1026, opened 2026-06-20
fix(quant): count num_mapped after the strand-compatibility filter (stranded over-count, #1025)

### LINKED ISSUE #1025: Stranded library inflates num_mapped / percent_mapped (Σ NumReads ≠ num_mapped); unstranded and C++ salmon agree
## Summary

In **selective-alignment / sketch (reads) mode**, when the library is **stranded**, salmon 2.1.0 over-reports `num_mapped` (and therefore `percent_mapped`, plus `num_compatible_fragments` / `num_assigned_fragments`, which are aliased to it). The reported mapped count includes fragments whose every mapping is strand-incompatible and is then dropped before quantification, so it no longer matches the quantified mass: `Σ NumReads (quant.sf) ≠ num_mapped`.

This was found while validating the #1019 fix on real data (it is **independent of #1019** and of bias correction). The same data quantified as **unstranded (IU)** conserves mass exactly, and **C++ salmon 1.12.0** conserves mass exactly even when it detects the library as stranded.

## Reproduction

GRCh38 Ensembl r115 decoy-aware gentrome, one real paired-end RNA-seq sample (first 3M pairs), `-p 12`, no bias flags. Same reads through every run:

| run | detected lib | `num_mapped` | `Σ NumReads` | gap |
| C++ salmon 1.12.0 `-l A` | ISF | 275565 | 275565 | **0.00%** |
| salmon 2.1.0 `-l A` | ISF | 311967 | 295597 | **5.25%** |
| salmon 2.1.0 `-l IU` (forced) | IU | 311967 | 311967 | 0.00% |
| salmon 2.0.1 `-l A` | IU (mis-detected) | 488857 | 488822 | 0.01% |

`lib_format_counts.json` for the 2.1.0 ISF run reports `compatible_fragment_ratio: 1.0` and `num_assigned_fragments: 311967`, yet `Σ NumReads = 295597`, so the counters are internally inconsistent. Note 2.1.0 and C++ both correctly detect the library as ISF (the `-l A` detection fix in 2.1.0 is working); only the counting differs.

## Root cause

`record()` in `crates/salmon-quant/src/processor.rs` increments `num_mapped` as soon as a fragment has any mapping, **before** the strand-compatibility filter. A fragment whose every mapping is strand-incompatible is then dropped (`compat.is_empty() -> return`) and contributes no mass, but it has already been counted as mapped.

For an unstranded library the filter never empties `compat`, so the invariant holds there. C++ salmon counts a fragment as mapped only once it has a compatible mapping, which is why C++ stays at `Σ NumReads == num_mapped` even for stranded libraries.

## Impact

Reported mapping rate is inflated on stranded data (here 10.40% reported vs ~9.85% actually quantified), and the `Σ NumReads == num_mapped` invariant downstream tools may rely on is broken. Quantification of the compatible fragments themselves is unaffected; this is a counting/reporting bug. Reads (SA/sketch) mode only; alignment-based (`-a`) mode uses a separate counter and is not affected.

A PR with a one-line-of-logic fix (count `num_mapped` after the compatibility filter, matching C++) and a regression test follows.

BenjaminDEMAILLE:
## What

Count a fragment as mapped only after it passes the strand-compatibility filter, matching C++ salmon. Fixes #1025.

## Why

In reads (SA/sketch) mode, `record()` incremented `num_mapped` for every fragment with any mapping, **before** the strand-compatibility filter. On a stranded library, a fragment whose every mapping is strand-incompatible is then dropped (`compat.is_empty() -> return`) and contributes no mass, but it had already been counted as mapped. This inflated `num_mapped` / `percent_mapped` / `num_compatible_fragments` / `num_assigned_fragments` and broke the invariant `Σ NumReads == num_mapped`. Unstranded libraries are unaffected (the filter never empties `compat`), and C++ salmon counts mapped post-filter, so it keeps the invariant.

Same reads (GRCh38 r115 decoy gentrome, one real PE sample, no bias flags):

| run | detected lib | `num_mapped` | `Σ NumReads` | gap |
| C++ salmon 1.12.0 `-l A` | ISF | 275565 | 275565 | 0.00% |
| 2.1.0 `-l A` (before) | ISF | 311967 | 295597 | 5.25% |
| 2.1.0 `-l IU` | IU | 311967 | 311967 | 0.00% |

## Change

`crates/salmon-quant/src/processor.rs`:
- Move the `num_mapped` increment (and the orphan classification) to **after** the `compat.is_empty()` check.
- Library-type auto-detection still samples observed formats from the raw mappings, before the filter (unchanged behavior).
- Orphan classification now keys off the best **compatible** mapping (identical for unstranded libraries, where the filter keeps every mapping).

Alignment-based (`-a`) mode uses a separate counter and is untouched.

## Testing

- New regression test `stranded_num_mapped_matches_quantified_mass` (`crates/salmon-quant/tests/quant_end_to_end.rs`): asserts `Σ NumReads == num_mapped` and that strand-incompatible fragments (inward FR reads quantified under the opposite stranded type `ISR`) are not counted as mapped. Fails before this change (`num_mapped = 1000 / 1000`), passes after.
- `cargo test --workspace`: 152 passing.
- `cargo fmt --all --check`: clean.
- `cargo clippy` on the changed crate: no findings in the changed files. (Two pre-existing `clippy` 1.95.0 lints in `salmon-model`/`salmon-map`, unrelated to this change, are left untouched.)

### DIFF
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -737,7 +737,7 @@ where
         let mut workers = Vec::with_capacity(cfg.nthreads);
         for _ in 0..cfg.nthreads {
             let rx = rx.clone();
-            workers.push(scope.spawn(move || -> (Local, u64) {
+            workers.push(scope.spawn(move || -> (Local, u64, u64) {
                 let mut local = Local::new(
                     cfg.use_error_model,
                     cfg.error_bins,
@@ -748,6 +748,7 @@ where
                     cfg.pos_bias,
                 );
                 let mut count = 0u64;
+                let mut mapped = 0u64;
                 let mut frags: Vec<FragRecord> = Vec::new();
                 while let Ok((log_fm, raw_batch)) = rx.recv() {
                     let ctx = FragCtx {
@@ -769,6 +770,7 @@ where
                         log_fm,
                     };
                     let mut since_flush = 0usize;
+                    let mut batch_mapped = 0u64;
                     for raw_group in &raw_batch {
                         frags.clear();
                         for r in raw_group {
@@ -776,7 +778,9 @@ where
                                 frags.push(f);
                             }
                         }
-                        process_fragment(&frags, &ctx, &mut local);
+                        if process_fragment(&frags, &ctx, &mut local) {
+                            batch_mapped += 1;
+                        }
                         // Publish the error-model delta into the shared model
                         // every FLUSH_INTERVAL fragments so other workers' `basis`
                         // sees fresh-enough training. The update granularity is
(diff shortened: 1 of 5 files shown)


rob-p:
Thanks for the precise diagnosis and fix, @BenjaminDEMAILLE! Merging this. I pushed one follow-up commit (`b2190cdb`) extending the same fix to **alignment (`-a`) mode**, which had the identical accounting bug but larger: `num_mapped` was the count of every aligned fragment in the BAM (before the per-fragment strand-compatibility filter), so on a stranded library it over-reported by ~49% on FR data and broke `Σ NumReads == num_mapped` there too. `process_fragment` now reports whether a fragment was assigned, and the alignment-mode console/log distinguish *aligned fragments* (`num_processed`) from *strand-compatible & quantified* (`num_mapped`) so a <100% rate on a stranded BAM reads as strand-incompatibility rather than lost alignments. Added a SAM-based regression test. This will ship in 2.1.1.

PR #1027, opened 2026-06-20
feat(quant): per-fragment-seeded FLD acceptance (parallel determinism, stage 1)

BenjaminDEMAILLE:
## What

Seed the fragment-length-distribution (FLD) acceptance sampling per fragment instead of from a thread-local PRNG, so the set of fragments that train the FLD is independent of thread scheduling. This is **stage 1** of the parallel-determinism follow-up named in the 2.1.0 notes ("per-fragment-seeded FLD acceptance + order-independent accumulation").

## Why

salmon trains the FLD online by accepting each concordant mapping's length with probability equal to its abundance-aware posterior. The acceptance coin was drawn from a `thread_local!` xorshift PRNG (`processor.rs`), so *which* fragments trained the FLD depended on how reads were distributed across worker threads. That is the primary documented source of the small multi-threaded run-to-run wobble (~0.26% of assigned reads on SRR1039508).

## Change

`crates/salmon-quant/src/processor.rs`:
- Replace the thread-local `FLD_RNG` / `fld_rng_u01()` with a per-fragment `FldSampler`:
  - `for_read(read_id)` seeds from `xxh3_64(read_id) | 1` (xxh3 is the workspace hash already used for eq-class labels; `| 1` keeps the seed nonzero for any input).
  - `u01(i)` derives the draw for compatible-mapping index `i` statelessly from `(seed, i)` via the splitmix64 finalizer, so it does not depend on how many earlier mappings of the fragment were concordant.
- Thread the read id into `record()` and build the sampler once per fragment, only inside the training window (`online_post.is_some()`), so there is no added cost after the model freezes.

The acceptance decision becomes a pure function of `(read_id, mapping_index, posterior)`. Reads (SA/sketch) mode only; the alignment-based (`-a`) path trains the FLD unconditionally (no RNG gate) and is untouched.

## Scope / honest claim

This makes the FLD-training set thread-independent and substantially tightens run-to-run agreement. It does **not** by itself make `-p N` byte-identical: the order-dependent online-mass and bias-model accumulation are a separate, larger stage (planned as a follow-up PR using order-independent integer-count FLD accumulation + a fixed-order reduction). Output changes bit-for-bit vs 2.1.0 even at `-p 1` (a different but statistically equivalent sampled FLD-training set; still U[0,1) draws, still accept with probability `post[i]`), which is expected and accuracy-neutral.

## Testing

- New unit tests (`processor.rs`): acceptance is order/thread-independent (same accept vector under forward/reversed/shuffled evaluation), deterministic per read id, seed-robust (incl. empty id), and in `[0,1)`.
- `cargo test -p salmon-quant`: green (unit + existing end-to-end accuracy tests).
- `cargo fmt --all --check`: clean. `cargo clippy` on the changed crate: no findings in the changed file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2452,6 +2452,7 @@ dependencies = [
  "serde_json",
  "tempfile",
  "tracing",
+ "xxhash-rust",
 ]
 
 [[package]]
--- a/crates/salmon-quant/Cargo.toml
+++ b/crates/salmon-quant/Cargo.toml
@@ -17,6 +17,7 @@ salmon-eqclass = { workspace = true }
 salmon-infer = { workspace = true }
 piscem-rs = { workspace = true }
 paraseq = "0.4"
+xxhash-rust = { workspace = true }
 anyhow = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/crates/salmon-quant/src/processor.rs
+++ b/crates/salmon-quant/src/processor.rs
@@ -236,22 +236,45 @@ fn frag_log_prob(
     }
 }
 
-thread_local! {
-    /// Per-thread PRNG state for stochastic FLD-sample acceptance (mirrors
-    /// salmon's per-thread RNG used when training the fragment-length model).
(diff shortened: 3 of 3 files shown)


rob-p:
superseded by #1039.

PR #1028, opened 2026-06-20
feat(model): order-independent FLD accumulation (parallel determinism, stage 2a)

BenjaminDEMAILLE:
## What

Make the fragment-length-distribution (FLD) training **order-independent**, so the trained FLD, the effective lengths derived from it, and (with stage 1) the resulting `quant.sf` are byte-identical regardless of `-p N`. This is **stage 2a** of the parallel-determinism follow-up.

> Stacks on #1027 (stage 1). The first commit here is #1027's; it disappears once that merges.

## Why

The FLD was trained by spreading a binomial kernel into a shared log-space histogram with lock-free `log_add_assign` CAS loops. Log-space addition is not associative, so concurrent observations committed in a thread-scheduling-dependent order left the trained histogram (and the effective lengths / online snapshots derived from it) differing in the low bits run-to-run.

## Change

`crates/salmon-model/src/fld.rs`:
- Every observation contributes the *same* kernel (`add_val(len, 0.0)`), so the trained histogram is exactly `prior + Σ_bin counts[bin]·kernel`. Add an order-independent per-bin observation counter (`counts: Vec<AtomicU64>`) and a `counts_hist()` that rebuilds the histogram from `prior + counts` in a fixed bin order.
- The deterministic reads-mode paths (`refresh_online`, `cache`, `mean`) now derive from `counts_hist()` instead of the live histogram, so the per-fragment online snapshot, the cached PMF/CMF (hence effective lengths), and `frag_len_mean` are byte-identical regardless of observation interleaving.
- The live `hist`/`tot_mass` log accumulation is kept for the alignment-mode (`-a`) live `pmf` reads, which do not require determinism. The previously write-only `sum` field is dropped (mean now derives from the counts).

## Scope / honest claim

Combined with stage 1, this makes `-p N` `quant.sf` byte-identical for datasets within the auxiliary-model burn-in window (the common case; default `numPreAuxModelSamples` = 5M fragments), where eq-class weights are score-only. The remaining sources for *all* dataset sizes and for bias-corrected runs (after-burn-in eq-class-weight snapshot timing, fractional multimapping eq-class weight summation order, and the bias online-mass accumulation) are addressed by stage 2b (a fixed-order reduction).

Accuracy is unchanged: the trained distribution is identical, only its accumulation order is fixed.

## Testing

- `fld.rs` unit test `accumulation_is_order_independent`: the cached PMF/CMF, the online snapshot, and the mean are byte-identical when the same multiset of lengths is added in forward / reversed / shuffled order. (Fails before this change — `log_add_assign` is not associative.)
- `quant_end_to_end.rs` test `quant_is_byte_identical_across_thread_counts`: build an index, quantify the same reads at `-p 1` and `-p 4` (twice), assert `quant.sf` byte-identical.
- `cargo test -p salmon-model -p salmon-quant -p salmon-align`: green. `cargo fmt --all --check`: clean. `cargo clippy` on the changed crate: no new findings (one pre-existing `explicit_counter_loop` in `add_val`, unchanged).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2452,6 +2452,7 @@ dependencies = [
  "serde_json",
  "tempfile",
  "tracing",
+ "xxhash-rust",
 ]
 
 [[package]]
--- a/crates/salmon-model/src/fld.rs
+++ b/crates/salmon-model/src/fld.rs
@@ -12,7 +12,7 @@
 use salmon_core::atomic::AtomicF64;
 use salmon_core::math::{log_add, LOG_0, LOG_EPSILON};
 use statrs::distribution::{Binomial, ContinuousCDF, Discrete, Normal};
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, RwLock};
 
 /// Tracks the observed distribution of fragment lengths.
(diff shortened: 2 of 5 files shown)


rob-p:
superseded by #1039.

PR #1029, opened 2026-06-20
feat(quant): add --writeBam (BGZF BAM alignment output)

BenjaminDEMAILLE:
## What

Add `--writeBam <path>`, the binary counterpart of `--writeMappings`: the same spoofed-CIGAR alignment records, written as BGZF-compressed **BAM** via `noodles-bam` instead of text SAM.

## Why

`--writeMappings` emits SAM text; many downstream tools want BAM. BAM is also the pragmatic choice over CRAM here: CRAM is reference-compressed, so `noodles-cram`'s writer requires a `fasta::Repository` containing every transcript and `.expect("missing reference sequence")`s without it (a full second copy of the transcriptome in memory, ~hundreds of MB on GRCh38, plus reference-compression of spoofed alignments). BAM stores read sequences verbatim, so it needs no reference and is robust.

## Change

New `crates/salmon-quant/src/bam.rs`, mirroring `sam.rs`:
- `BamWriter`: a mutex-guarded `noodles_bam::io::Writer<bgzf::io::Writer<BufWriter<File>>>` plus the `sam::Header` built from the index references (one `@SQ` per reference, decoys included). `write_records` writes a worker's buffered records under the lock; `finish` writes the BGZF EOF block.
- `build_records`: the analogue of `sam::write_fragment`, returning owned `noodles_sam::alignment::record_buf::RecordBuf`s with the same flags, overhang-clipped CIGAR, SEQ (reverse-complemented for reverse-strand mappings), and `NH`/`HI`/`XT`/`AS` tags, for paired / orphan / single-end mappings.

Wired through `QuantOptions::write_bam`, the `--writeBam` CLI flag, and a per-thread `bam_buf` in the processor that is flushed alongside the existing SAM buffer at batch/thread boundaries (same contention profile).

## Testing

- New round-trip test `writes_readable_bam` (`tests/quant_end_to_end.rs`): quantify with `write_bam`, reopen the file with `noodles-bam`, and assert the header lists every reference and the records parse.
- `cargo test -p salmon-quant`: green. `cargo fmt --all --check`: clean. `cargo clippy` on the changed crates: no findings in the changed files.

## Notes

Record ordering across threads is nondeterministic (unsorted output; downstream tools that need coordinate/name order should sort). A CRAM writer could be added later behind the same `build_records` if the in-memory reference-repository cost is judged acceptable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -297,6 +297,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "63044e1ae8e69f3b5a92c736ca6269b8d12fa7efe39bf34ddb06d102cf0e2cab"
 dependencies = [
  "memchr",
+ "regex-automata 0.4.14",
  "serde",
 ]
 
@@ -2647,9 +2648,14 @@ name = "salmon-quant"
 version = "2.3.1"
 dependencies = [
  "anyhow",
+ "bstr",
  "flate2",
  "jiff",
  "libradicl",
+ "noodles-bam",
+ "noodles-bgzf",
+ "noodles-core",
+ "noodles-sam",
  "paraseq",
  "piscem-rs",
  "rayon",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,16 +42,12 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 # parallelism
 rayon = "1"
-crossbeam = "0.8"
 crossbeam-channel = "0.5"
 # concurrent map for equivalence classes
 scc = "2"
 dashmap = "6"
 # fast (non-DoS-resistant) hasher for hot per-read integer-keyed maps
 ahash = "0.8"
-# numerics
-nalgebra = "0.33"
-ndarray = "0.16"
 statrs = "0.18"
 # rng (PCG, matching salmon's pcg usage)
 rand = "0.8"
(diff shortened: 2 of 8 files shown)


rob-p:
Thanks for this, @BenjaminDEMAILLE — `--writeBam` is a clean, self-contained addition, and choosing BAM over CRAM (avoiding the in-memory reference repository for reference compression) is the right call here.

**Queued for the release after the current one.** The pending release (#1039, targeted 2.2.0 — RAD `--writeRad`/`--rad` decoupling, `--deterministic`, deterministic + faster bias models, chunk compression) is finalized, QA'd, and CI-green, and `--writeBam` is orthogonal to that theme. Folding it in now would reopen a validated release and conflict in our most sensitive files (a merge dry-run hits `salmon-quant/src/{lib.rs,processor.rs}` and `salmon-cli/src/main.rs` — exactly the `QuantOptions`/per-thread-flush/CLI surface the RAD work touches). Rather than carry that, we'll land this as a fast-follow once #1039 merges — at which point the rebase onto the new `master` dissolves those conflicts, since our changes will be in the base.

A couple of small things to fold in when it's rebased (not blockers):
- Share `build_records` between `sam.rs` and `bam.rs` rather than duplicating — both build identical records, so a single record-builder emitting either sink keeps them from drifting.
- Add `--writeBam` to the docs (`website/.../reference/cli` and `output-formats`) and a release-notes line.
- Note the nondeterministic record order (same as the existing SAM output) in the docs.

Appreciate it — this is genuinely wanted downstream, just sequencing it behind the determinism release.

BenjaminDEMAILLE:
Rebased onto current `master` (v2.3.1): the earlier conflict with the `--writeRad` / `--deterministic` work is resolved, the full workspace builds clean (no new warnings), and the round-trip `writes_readable_bam` test passes. Ready for review whenever you have a moment, @rob-p 🙏

BenjaminDEMAILLE:
@rob-p Gentle ping so this one doesn't slip 🙂

Back on 2026-06-28 you queued `--writeBam` for "the release after the current one" (then 2.2.0). Since then 2.3.0, 2.3.1 and 2.3.2 have all shipped, so the queued slot has come and gone.

Status is still green: rebased onto current `master`, workspace builds clean with no new warnings, and the `writes_readable_bam` round-trip test passes. Ready to merge whenever you have a slot, and happy to rebase onto `develop` if that's where it should land now. Thanks @rob-p!

rob-p:
Closed in favor of #1065, which this inspired. See the reasoning and differences there.

PR #1030, opened 2026-06-20
feat(eqclass): deterministic per-class weight reduction (parallel determinism, stage 2b)

BenjaminDEMAILLE:
## What

Make the equivalence-class **weight reduction deterministic**, so the EM inputs (and thus `quant.sf`) no longer drift run-to-run on multimapping data. Stage 2b of the parallel-determinism follow-up.

> Draft: stacks on #1027 (stage 1) and #1028 (stage 2a). Marked draft because of the memory trade-off and the remaining sources below — I'd like your read before treating it as merge-ready.

## Why

Eq-class weights were summed live into a `DashMap` as fragments mapped, so a multimapping class with ≥3 fractional contributions was summed in a thread-scheduling-dependent order. Float addition is not associative, so those weights (and the abundances derived from them) drifted in the low bits run-to-run. (Uniquely-mapping classes sum exactly-1.0 contributions, so they were already stable — which is why stages 1+2a already give byte-identical `-p N` on unique-mapping data.)

## Change

`crates/salmon-eqclass`:
- Buffer each fragment's contribution `(key, group, weights, count)` into a **key-sharded** buffer (64 shards, low contention) instead of summing in place. The key is a stable per-fragment id (`xxh3_128(read_id)`, passed from the reads path; alignment mode / single-threaded callers fall back to a monotonic sequence key).
- `finish()` sorts contributions by `(txps, bins, key)` and sums each class's weights in that fixed order. The summed weights **and** the class ordering are now a pure function of the input, independent of thread count. (This also removes a latent tie: the old sort ordered classes by `txps` only, leaving same-`txps`/different-`bins` classes in `DashMap` iteration order.)

`crates/salmon-quant`: the reads path passes `xxh3_128(read_id)` as the key.

## Scope / honest trade-offs

- Combined with stages 1 + 2a, the eq-class inputs to the EM are order-independent for the reads path.
- **Two sources remain** for fully bit-exact `-p N` on *large* / bias-corrected runs, out of scope here: (1) the after-burn-in FLD **snapshot timing** a fragment reads, and (2) the online-mass accumulation behind `--seqBias/--gcBias` effective lengths. Both need the fixed-order reduction extended to the online phase.
- **Memory:** contributions are buffered until `finish()` (bounded by the mapped-fragment count) rather than collapsed in place — a real increase on large inputs. This is the trade-off to flag.

## Testing

- New unit test `weight_sum_is_insertion_order_independent`: a class's summed weights are byte-identical when the same keyed contributions are added in forward / reversed / shuffled order. **Verified it fails for arrival-order accumulation** (off by 1 ULP) and passes only with the keyed sorted reduction.
- `cargo test --workspace`: 157 passing (incl. the stages 1+2a determinism guards). `cargo fmt --all --check`: clean. `cargo clippy` on the changed crates: no findings in the changed files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2452,6 +2452,7 @@ dependencies = [
  "serde_json",
  "tempfile",
  "tracing",
+ "xxhash-rust",
 ]
 
 [[package]]
--- a/crates/salmon-eqclass/src/lib.rs
+++ b/crates/salmon-eqclass/src/lib.rs
@@ -12,6 +12,8 @@
 //! into a flat [`CollapsedEqClasses`] for inference.
 
 use std::hash::{Hash, Hasher};
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Mutex;
 use xxhash_rust::xxh3::Xxh3;
 
 /// The label of an equivalence class: the sorted set of transcript ids a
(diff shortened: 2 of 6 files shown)


rob-p:
superseded by #1039.

PR #1031, opened 2026-06-20
feat(quant): opt-in --deterministic two-pass inference (full -p N reproducibility)

BenjaminDEMAILLE:
## What

Add an **opt-in** `--deterministic` mode that makes `salmon quant` produce **byte-identical `quant.sf` regardless of `-p N`**, including bias-corrected and large (after-burn-in) runs. Off by default: the default path is the historical inline, fully-parallel inference, unchanged.

> Stacks on #1027 / #1028 / #1030. Reworked from "deterministic by default" to opt-in per @rob-p's review.

## Why

Stages 1 (#1027), 2a (#1028) and 2b (#1030) made the FLD acceptance, FLD accumulation, and eq-class weight reduction order-independent. The remaining sources all trace to the **online masses**, accumulated via lock-free CAS during the parallel loop: the FLD-acceptance threshold, the after-burn-in FLD snapshot a fragment reads, and the bias-model weights (hence effective lengths) all depend on them. Those are only deterministic if fragments go through the online phase in a fixed order, which is what the opt-in pass provides.

## Change

A new `--deterministic` flag (`QuantOptions::deterministic`, default `false`) selects between two inference strategies:

- **Default (inline, parallel):** each worker maps and runs the per-fragment inference (`record`) immediately, accumulating the online posterior, FLD, bias models, and eq-classes concurrently. No fragment store, no serial phase, no added overhead. This is the existing behaviour, restored unchanged.
- **`--deterministic` (two-pass):** pass 1 maps in parallel and only buffers `(key, mappings)` (`key = xxh3_128(read_id)`) into a shared sink; pass 2 (`run_inference_serial`, single-threaded) sorts by key and replays the online posterior, FLD training, bias collection, and eq-class accumulation in that fixed order. The online masses, FLD, bias models, and eq-class weights become a pure function of the input.

The burn-in gate reads the shared `num_processed` counter inline (as before) and the online inference's progressive count in the deterministic pass, so `numPreAuxModelSamples` is honored in either path.

## Trade-offs

- **Default unchanged:** with the flag off there is no fragment store and no serial phase. The cost is paid only when the flag is set.
- **`--deterministic` perf:** the serial inference pass is a small slice of wall time on real data, where the (still-parallel) mapping pass dominates; the ~20% seen on trivial synthetic reads is an artifact of their near-instant mapping.
- **`--deterministic` memory:** the inter-pass buffer holds `(u128 key, mappings)` per mapped fragment (the `read_id` bytes are not retained). This is inherent to a by-construction determinism guarantee, and the natural next step (see discussion) is a RAD-backed on-disk store so very large inputs stream from disk in pass 2.

## Testing

- Accuracy unchanged: `selective_alignment_quantification_tracks_truth` / `pseudoalignment_…` (default inline path) still recover the true fractions.
- Determinism (opt-in): `quant_is_byte_identical_across_thread_counts` and the full-precision `multimapping_counts_bit_identical_across_threads` (20k multimapping fragments, `--seqBias --gcBias`, after-burn-in forced, `-p 1` vs `-p 8`) set `deterministic = true` and pass, guaranteed **by construction**.
- `cargo test --workspace`: 158 passing. `cargo fmt --all --check`: clean. `cargo clippy`: no findings in the changed files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2452,6 +2452,7 @@ dependencies = [
  "serde_json",
  "tempfile",
  "tracing",
+ "xxhash-rust",
 ]
 
 [[package]]
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -568,6 +568,13 @@ struct QuantArgs {
     /// salmon's default is 1000000, this port's is 5000.)
     #[arg(long = "numPreAuxModelSamples", default_value_t = 5000)]
     num_pre_aux_model_samples: u64,
+    /// Produce byte-for-byte identical `quant.sf` regardless of thread count
+    /// (`-p`). Splits quantification into a parallel mapping pass and a
+    /// single-threaded, key-sorted inference pass, so every order-dependent
+    /// accumulation is reproducible. Costs some memory (mapped fragments are held
+    /// between passes) and serializes the inference phase; off by default.
+    #[arg(long = "deterministic")]
+    deterministic: bool,
     /// [accepted; not yet implemented] disable fragment-length-distribution
     /// concordance in the per-fragment probability.
     #[arg(long = "noFragLengthDist")]
@@ -893,6 +900,7 @@ fn run_quant(args: QuantArgs, quiet: bool) -> Result<()> {
     opts.cond_gc_bins = args.conditional_gc_bins;
     opts.skip_quant = args.skip_quant;
     opts.num_pre_aux_model_samples = args.num_pre_aux_model_samples;
+    opts.deterministic = args.deterministic;
     opts.map_config.seed_mode = seed_mode(args.unimems, args.refmems, args.sparse_seeds);
     // alignment scoring (selective alignment)
     opts.map_config.align.match_score = args.ma as i8;
--- a/crates/salmon-eqclass/src/lib.rs
+++ b/crates/salmon-eqclass/src/lib.rs
@@ -12,6 +12,8 @@
 //! into a flat [`CollapsedEqClasses`] for inference.
 
 use std::hash::{Hash, Hasher};
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Mutex;
 use xxhash_rust::xxh3::Xxh3;
 
 /// The label of an equivalence class: the sorted set of transcript ids a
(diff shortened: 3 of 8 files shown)


rob-p:
Hi @BenjaminDEMAILLE,

  Thank you for suggesting this featured hinted at in the release notes (and stacking the separately reviewable PRs to implement them)!

  One suggestion I would like to make, since I've been thinking about byte-for-byte reproducibility in `salmon` for a *long* time: I'm definitely open to such a feature, but I have a strong preference that it be _opt-in_ rather than the default, triggered by an appropriately named flag. Specifically, I think the easiest way to do this, as the PRs suggest, is with a two pass approach. Yet, it adds runtime overhead to the hot path and, on very large inputs, creates a mapping store that should either be held in memory (increasing RAM usage) or written to disk (increasing I/O).  I *do* think there are scenarios where byte-for-byte reproducibility may be important, which is why I'm open to it, but I'd prefer to make this _opt-in_, at least for several releases.
  
   One other thought. If we decide to go the route of writing out mappings in one pass for later ingestion and inference in a second pass, then I have a _strong_ preference that instead of a bespoke binary format, we write out the mapping in [RAD format](https://github.com/COMBINE-lab/libradicl/tree/develop), as used by our other tools `piscem`, `piscem-infer`, and `alevin-fry`. In fact, one feature that I have been thinking about adding to salmon 2 (now that it's in Rust) is RAD format input, so that users could choose to map the reads using `piscem` (with it's bulk read mapping command) and then perform inference with `salmon`.
   
   Anyay, thanks again for these PRs, and I'm happy to discuss any of the above with you, to get your thoughts, suggestions, and feedback on the best path forward.
   
 --Rob

BenjaminDEMAILLE:
Hi @rob-p, thanks for the thoughtful review. Both points make sense, and opt-in is clearly the right call. I've reworked this PR accordingly.

**Opt-in (done).** The two-pass is now behind a `--deterministic` flag, default off. With the flag off the inference runs inline during the parallel mapping pass exactly as before: no fragment store, no serial phase, no extra overhead. With it on, pass 1 buffers `(key, mappings)` and pass 2 replays the inference single-threaded in key-sorted order. `cargo test --workspace` stays green (158); the determinism guards now opt in via the flag, and the accuracy/truth-recovery tests still exercise the default inline path. Happy to rename the flag if `--deterministic` isn't your preferred convention (`--reproducible` / `--byteReproducible` also read fine).

Two things I'd value your steer on:
- The two cheap, order-independence pieces (per-fragment-seeded FLD acceptance #1027 and integer-count FLD accumulation #1028) add no runtime/memory cost and tighten the existing `-p N` wobble even without the two-pass. I'm inclined to keep those always-on as plain improvements, but if you'd rather the default output stay byte-identical to today's for several releases, I'll gate them behind the same flag too.
- The eq-class keyed reduction (#1030) carries a small buffering cost like the two-pass, so I can gate it behind `--deterministic` as well for consistency. Tell me whether you'd prefer that or keep it standalone.

**RAD.** I really like this and it resolves the bespoke-format smell I was uneasy about. Writing pass 1's mappings as RAD means the deterministic-quant store and the piscem-map -> salmon-quant path share a single ingestion route, and it gives the natural answer to the large-input RAM concern (stream RAD from disk in pass 2 instead of holding everything in memory). I'll read through the spec / `libradicl` (and how piscem-infer and alevin-fry consume it) and come back with a concrete design.

On sequencing, would you prefer I:
1. land the opt-in flag now with the current in-memory store (small, gets the feature available behind the flag), then do the RAD-backed store + RAD input as a focused follow-up; or
2. go straight to a RAD-backed store (pass 1 writes RAD, pass 2 reads it) so the first version of the flag already uses the format you'd want long-term?

I lean slightly toward (2) since it avoids shipping a store we'd immediately replace, but it's more work up front and I'd rather match your roadmap. Either way RAD *input* to salmon-quant sounds like a great feature on its own and I'd be glad to help with it.

Benjamin

rob-p:
superseded by #1039.

Issue #1032, opened 2026-06-20
Tracking: byte-identical multi-threaded quantification (parallel-determinism follow-up) + related PRs


BenjaminDEMAILLE:
Tracking issue for the **parallel-determinism follow-up** named in the 2.1.0 release notes ("per-fragment-seeded FLD acceptance + order-independent accumulation"), plus the other open PRs from the same round of testing. Opening this so the determinism stack can be reviewed/discussed as a unit rather than four disconnected PRs.

## Goal

Make multi-threaded `salmon quant` produce **byte-identical `quant.sf` regardless of `-p N`**, by construction. 2.1.0 was already "more reproducible than C++" but not bit-exact across thread counts; the wobble traced to thread-scheduling-dependent accumulation in the online phase.

## Determinism stack (staged, each independently reviewable; stacked in order)

- #1027 — **stage 1**: per-fragment-seeded FLD acceptance. Removes the primary documented source (a thread-local RNG decided which fragments trained the FLD). Small, low-risk.
- #1028 — **stage 2a**: order-independent FLD accumulation via integer per-bin counts (every observation adds the same kernel, so the histogram is a count convolved with the kernel; counts are commutative).
- #1030 — **stage 2b**: deterministic per-class eq-class weight reduction (buffer keyed contributions, sum each class in a fixed key-sorted order; float addition is not associative).
- #1031 — **deterministic two-pass inference (completes the goal)**: the remaining sources (FLD-acceptance threshold, after-burn-in FLD snapshot timing, bias-model weights) all trace to the online masses. Pass 1 maps in parallel and buffers `(key, mappings)`; pass 2 replays the online posterior / FLD training / bias collection / eq-class accumulation single-threaded in key-sorted order. Deterministic by default.

### Open design point (#1031)

#1031 serializes the inference phase and buffers mapped fragments between passes. Measured trade-offs: the serial pass is a low single-digit % of wall time on real data (mapping, which stays parallel, dominates — the ~20% on trivial synthetic reads is an artifact); the buffer holds `(u128 key, mappings)` per mapped fragment. This is the architectural choice that wants a maintainer call — happy to switch it to opt-in (`--deterministic`, default = the current inline path) if you'd rather not change the default behaviour or pay the buffer on every run.

[...]

PR #1033, opened 2026-06-20
feat(quant): RAD-format deterministic mapping store + external merge sort

BenjaminDEMAILLE:
## What

Write the `--deterministic` mapping store in **RAD format** (the format used by piscem, piscem-infer, and alevin-fry) instead of an in-memory `Vec`, and stream it through an **external merge sort** in pass 2 so the deterministic path stays memory-bounded on large inputs.

> Stacks on #1031 (the opt-in two-pass). The 5 commits unique to this branch are the RAD store; the earlier commits in the diff belong to #1027 / #1028 / #1030 / #1031 and will drop out once those merge. Implements the direction discussed in #1032.

## Why

In #1032 the suggestion was: if the two-pass persists mappings between passes, write them in RAD rather than a bespoke binary format (sharing an ingestion route with piscem / piscem-infer / alevin-fry, and laying the groundwork for RAD-format input to salmon), and avoid holding the whole store in memory on very large inputs. This does both.

## Change

- **`crates/salmon-quant/src/rad.rs`** — a RAD-conformant writer + reader. The prelude (header + file/read/alignment `TagSection`s) is written via `libradicl`'s type primitives; the per-record hot loop is hand-written little-endian (as piscem's own RAD writer is). A salmon-specific alignment-level tag schema carries the full-fidelity per-mapping fields selective-alignment inference needs (`tid`, `weight`, `status`, orientation, `fragment_len`, `read_len`, the positions, `format`), a superset of the bulk profile. Only the fields pass-2 inference reads are stored.
- **Streaming store** — pass 1 appends each mapping batch as a RAD chunk (serialized off-lock, appended under a short-held mutex); the per-thread buffer is freed each batch, so pass 1 no longer accumulates a shared growing `Vec`.
- **External merge sort (`ExternalSort`)** — pass 2 spills the store into key-sorted runs of `RUN_SIZE` fragments and k-way merges them via a min-heap keyed by `(frag_key, run_index)` (total, reproducible order); run files are removed on drop. `run_inference_serial` now consumes the already-sorted stream one mini-batch at a time, so pass-2 peak memory is ~one run regardless of input size.
- **Conformance** — the 128-bit sort key is declared as two `u64` read-level tags rather than one `u128`: published `libradicl` 0.10 encodes but does not decode the `u128` type id (9), so a single-`u128` tag would make the prelude unreadable by the canonical reader. The bytes are identical to the `u128` little-endian layout.

## Testing

- Unit: record + tag-section + `format=None` round trips; **`ExternalSort` multi-run merge yields globally key-sorted output and cleans up its run files**; the prelude is parsed by **`libradicl`'s own `RadPrelude::from_bytes`** (proving conformance, this caught the `u128` decode gap).
- End-to-end determinism (all through the on-disk RAD round trip): byte-identical `quant.sf` / full-precision counts across `-p 1` vs `-p 8` for paired and single-end, with `--seqBias`/`--gcBias` and after burn-in; and a multi-run guard showing the result is identical across thread counts **and** identical to the single-run result (many small sort runs == one big run).
- `cargo test --workspace`: 165 passing. `cargo fmt --all --check`: clean. `cargo clippy`: no findings in the changed files.

## Note

This is the deterministic mapping **store**. RAD-format **input** (`piscem map` -> `salmon quant`) is a natural follow-up that reuses this reader, tracked separately.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,6 +187,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "222fb4925a15bea6a68075021910e03d6aa2d04951d71ff1d956190a551d738f"
 
 [[package]]
+name = "bio-types"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4dcf54f8b7f51450207d54780bab09c05f30b8b0caa991545082842e466ad7e"
+dependencies = [
+ "derive-new",
+ "lazy_static",
+ "regex",
+ "strum_macros",
+ "thiserror 1.0.69",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
 name = "bit-vec"
 version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


rob-p:
superseded by #1039.

PR #1034, opened 2026-06-20
feat(quant): RAD-format input (--rad) and --writeRad (decouple map/quant)

BenjaminDEMAILLE:
## What

Make the RAD mapping store a real interchange format, decoupling mapping from inference:

- **`--writeRad <file>`** writes salmon's mapping store to a path and keeps it.
- **`--rad <file>`** quantifies from a salmon-written RAD store instead of FASTQ reads.

So you can run mapping once and re-quantify without re-mapping, and it is the foundation for the `piscem map -> salmon quant` path discussed in #1032.

> Stacks on #1033 (the RAD store) and #1031. The commit unique to this branch is the input/output wiring.

## Change

- `--rad` skips mapping and runs the same deterministic, key-sorted pass-2 inference (`ExternalSort` + `run_inference_serial`) over the store, then the usual effective-length / EM / bias-correction / output. The index (`-i`) still supplies reference lengths and sequences (for effective lengths and bias). paired-vs-single is read from the RAD header; `processed == mapped` since the store holds only mapped fragments.
- `--writeRad` produces the store (implies the deterministic store is generated) and keeps it instead of deleting it after pass 2.
- `rad.rs`: `read_is_paired` + `verify_salmon_profile`, which **rejects a non-salmon RAD** (e.g. a piscem bulk RAD, whose reduced records this reader does not yet decode) with a clear error rather than silently misreading it.
- CLI: `--rad` satisfies the input requirement and is mutually exclusive with read inputs (`-1/-2/-r`).

## Testing

- New end-to-end test: map once at `-p 4` with `--writeRad`, then quantify from the store with `--rad` at `-p 8`; the resulting `quant.sf` is **byte-identical** to the store-producing run (with `--seqBias --gcBias`). This exercises the whole decoupled path and confirms it is thread-count independent.
- `cargo test --workspace`: 166 passing. `cargo fmt --all --check`: clean. `cargo clippy`: no findings in the changed files.

## Scope

This reads salmon's **own** full-fidelity RAD profile (carrying the per-mapping coverage weight and positions, so selective-alignment fidelity is preserved). Consuming **piscem's bulk** RAD profile (reduced, uniform-weight) is the natural next step and is cleanly gated today (a bulk RAD currently produces an explanatory error). The reader will reuse libradicl's `PiscemBulkReadRecord` decoder so the orientation/position decoding stays canonical.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,6 +187,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "222fb4925a15bea6a68075021910e03d6aa2d04951d71ff1d956190a551d738f"
 
 [[package]]
+name = "bio-types"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4dcf54f8b7f51450207d54780bab09c05f30b8b0caa991545082842e466ad7e"
+dependencies = [
+ "derive-new",
+ "lazy_static",
+ "regex",
+ "strum_macros",
+ "thiserror 1.0.69",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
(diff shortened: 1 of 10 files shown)


rob-p:
superseded by #1039.

PR #1037, opened 2026-06-21
feat(quant): accept piscem bulk/sketch-mode RAD as --rad input

BenjaminDEMAILLE:
## What

Make `--rad` accept a **piscem `map` (bulk/sketch-mode) RAD** in addition to salmon's own `--writeRad` store, so `piscem map -> salmon quant` works without re-mapping. This is the interop path discussed in #1032.

> Stacks on #1034 (the `--rad` / `--writeRad` wiring). The commit unique to this branch is the piscem-profile reader + the profile-dispatch.

## Change

- **`rad.rs`**: `detect_profile` replaces `verify_salmon_profile`, returning `RadProfile::{Salmon, PiscemBulk}` from the prelude tags (a `frag_map_type` read-level tag is piscem's signature; salmon's own 10-tag alignment schema is the store). `PiscemRadReader` streams a piscem RAD as `(key, mappings)` for the existing `run_inference_serial`, decoding each record via **libradicl's own `PiscemBulkReadRecord`** (no bespoke parser; matches piscem by construction) into uniform-weight sketch mappings, then drops decoy references via `filter_sketch_decoys`, exactly as salmon's native sketch path does. The file-level tag *values* (piscem's `ref_lengths`) are consumed between the prelude and the first chunk.
- **`lib.rs`**: `--rad` branches on the detected profile. salmon store -> the deterministic, key-sorted `ExternalSort` replay (unchanged). piscem -> stream in file order (the only deterministic order available, since bulk RAD carries no per-read hash key).
- **CLI**: `--rad` help now notes it also accepts a piscem RAD.

The index (`-i`) still supplies reference lengths/sequences for effective lengths and bias; positions/orientations/fragment lengths are present in the bulk record, so the bias models remain driveable.

## Testing

- New end-to-end test: `salmon quant --rad` over a hand-written piscem-profile RAD (canonical prelude written via `libradicl`; records emit the 2-bit orientation the reader expects), asserting read mass is conserved and fragments are assigned exactly to their transcripts.
- `cargo test --workspace`: all green. `cargo fmt --all --check`: clean. `cargo clippy`: no findings in the changed files.

## Notes / open points (see #1032)

- Reading uses the canonical `libradicl` decoder, so the format matches what piscem produces by construction; no port of `piscem-infer` was needed.
- For an orphan / single-end record the `frag_length` field is currently read as the read length (no template); it only feeds the ambiguous-fragment length term, not assignment, but worth confirming against piscem's writer.
- Minor: `libradicl 0.10`'s `PiscemBulkReadRecord::write` encodes orientation with a 3-bit scheme its own reader (2-bit `from_u32_paired_status`) does not round-trip. The reader is the canonical side, so this is harmless here; the test emits the 2-bit code piscem writes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,6 +187,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "222fb4925a15bea6a68075021910e03d6aa2d04951d71ff1d956190a551d738f"
 
 [[package]]
+name = "bio-types"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4dcf54f8b7f51450207d54780bab09c05f30b8b0caa991545082842e466ad7e"
+dependencies = [
+ "derive-new",
+ "lazy_static",
+ "regex",
+ "strum_macros",
+ "thiserror 1.0.69",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
 name = "bit-vec"
 version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


rob-p:
superseded by #1039.

PR #1039, opened 2026-06-27
feat(quant): deterministic RAD map/quant decoupling — --writeRad/--rad, --deterministic, bias-aware RAD, chunk compression

rob-p:
## Summary

This PR decouples mapping from quantification in the Rust salmon port via the **RAD format**, end to end, and makes the **bias models deterministic and faster**. Targeted as **salmon 2.2.0** (see `docs/release-notes-2.2.0.md`).

- a new **`salmon-rad`** crate (RAD schema, record types, writer) built on **libradicl 0.14**;
- **`--writeRad`** to emit per-fragment mappings, and **`--rad`** to quantify a RAD file directly (salmon-written *or* piscem `map-bulk`/sketch output), decoupling the map and quant phases;
- a **deterministic, information-adaptive** RAD quantifier that replaces the online (dual-phase) estimate in RAD mode and is **byte-identical across thread counts**;
- a **`--deterministic`** reads mode (map FASTQ once → intermediate RAD → requant) for reproducible results from FASTQ input;
- **bias-aware RAD** quantification (sequence/GC/positional), now also **deterministic** and substantially **faster**;
- optional **LZ4/zstd chunk compression** (`--radCompress`, on by default for RAD output).

It depends on the published **`libradicl` 0.14.0** (chunk compression + backpatchable file tags).

## Relationship to @BenjaminDEMAILLE's PRs (#1033, #1034, #1037)

This work builds directly on the RAD direction @BenjaminDEMAILLE opened, and **supersedes** these PRs:

- **#1033** — RAD-format deterministic mapping store + external merge sort
- **#1034** — RAD-format input (`--rad`) and `--writeRad` (decouple map/quant)
- **#1037** — accept piscem bulk/sketch-mode RAD as `--rad` input

**What it incorporates from his work:** the core design — decouple mapping from quantification through the RAD format; a `--writeRad` producer and a `--rad` consumer; treating **piscem bulk/sketch RAD as a first-class input**; and the goal of **deterministic** RAD-based quantification. The schema, the map/quant split, and the piscem-input handling all follow the path he laid out, and the prior commits credit him as co-author.

**Where it differs:**

- **Determinism mechanism.** #1033 achieves order-independence by writing a RAD *mapping store* and performing an **external merge sort** of records. This PR instead makes the *computation itself* order-independent — fixed-point (`u128`) equivalence-class weight accumulation and integer-count `DiscreteFld` histograms built once in a fixed order — and quantifies from a **fixed FLD**. No sort, no spill-to-disk, no external-merge pass: quantification is a single in-memory streaming pass that is deterministic *and* fast.
- **Information-adaptive pass count.** salmon-written RAD bakes an order-independent empirical FLD (+ initial abundances + resolved library format) into the **header** via libradicl 0.14 backpatchable file tags, so it quantifies in **1 pass**; piscem RAD (nothing baked) derives a unique-fragment FLD in a first pass, then quantifies — **2 passes**. A `baked_flags` marker drives the dispatch; readers also use the baked library format for `-l A` concordance filtering without re-inference.
- **Scope added beyond the original PRs:** bias-aware RAD quant (seq/GC/positional, sharing the BAM bias-correction tail); **deterministic + faster bias models**; optional **LZ4/zstd chunk compression**; and a FASTQ-side **`--deterministic`** mode.

## What's in this PR

**`salmon-rad` crate** — RAD schema/prelude builder, salmon bulk record type, and a thread-safe `RadOutputWriter` (one chunk per worker flush) on libradicl 0.14.

**RAD output (`--writeRad`)** — writes per-fragment mappings (sketch or selective-alignment profile); combine with `--skipQuant` to map only. Bakes an order-independent FLD, initial abundances, and library format into the header at finalize.

**RAD input (`--rad`)** — parallel quantification of a RAD file via libradicl's `ParallelRadReader`. Auto-detects salmon vs piscem-bulk/sketch profiles; consumes baked header values when present.

**Deterministic info-adaptive quant** — online inference removed from RAD mode; fixed-FLD quant; 1 pass (baked) or 2 passes (derive). Byte-identical at `-p 1` vs `-p 16`.

**`--deterministic` reads mode** — map FASTQ once to an intermediate RAD (order-independent FLD + library format baked), then requant; intermediate is compressed and removed by default.

**Bias-aware RAD** — collect sequence/GC/positional bias from RAD placements and apply the shared bias-correction tail. `--deterministic` bias no longer requires `-t` (reference sequences are sourced from the index for the second pass).

**Deterministic bias models** — sequence/GC/positional observed models accumulate per-fragment mass in **fixed-point integer** buffers and materialize once, so the trained models (and the bias-corrected effective lengths) are **byte-identical across thread counts**. This is the only behavioral change vs 2.1.2 for bias runs; accuracy vs truth is unchanged.

**Faster bias correction** — profiled the effective-length sweep and sped up the dominant work:
- the seqBias per-position factor build folds obs/exp into one precomputed `obs − exp` log-bias table (one encode + one sweep instead of two) → **~1.9× on the seqBias correction sweep**;
- the length-sweep convolution for any non-GC bias (seq and/or positional) is an FFT cross-correlation, `O(L log L)` instead of `O(L · n_len)`.
Both are reassociation-level (deterministic, accuracy-vs-truth unchanged).

**Chunk compression** — `--radCompress=lz4|zstd|none` (default **lz4**) and `--noCompressRad`; on by default for `--writeRad` and the `--deterministic` intermediate. Transparent and lossless (decompressed in libradicl's reader thread; consumers unchanged; codec-absent RADs read as uncompressed). On real 36 M-read RAD: lz4 ≈ 1.25×, zstd ≈ 1.9× smaller; lz4 write is *neutral-to-faster*.

**Supporting improvements** (bundled here):
- default minimizer length **m = 19** (matches piscem);
- **lazy reference-sequence load** — skip the multi-GB decoy `refseq` in sketch mode without seq/GC bias (~3 GB saved on a human + genome-decoy index);
- reusable **per-thread mapping scratch** (`map_*_into`) to cut per-read allocation;
- promote diagnostic env vars to **hidden flags** (`--biasSeedEMIters`, `--dumpBiasModels`); add **`--keepFixedFasta`**;
- depend on published **libradicl 0.14** (drops the local-path patch).

## Validation

Full pre-release QA on simulated and 36 M-read human data, all green:

- **Determinism:** RAD quant and `--deterministic` produce **byte-identical** `quant.sf` across thread counts (`-p 1` ≡ `-p 16`), **with and without bias**.
- **Compression:** requant of `none`/lz4/zstd RAD is **byte-identical** (sims + 36 M human); a test (`compressed_rad_quantifies_identically`) exercises the full reader-decompression path, plus a `chunk_codec` tag-presence test. Sizes (sim): none 454 MB → lz4 273 MB → zstd 193 MB.
- **Backward compat:** real piscem `map-bulk` RAD (**no `chunk_codec` tag**) quantifies all 31.5 M fragments with **mass conserved** (Σ NumReads = fragment count), decoy and no-decoy.
- **Alignment (BAM/SAM) mode:** core + `--noErrorModel` + seq/GC/pos bias (shared tail) + bootstraps + Gibbs all run with mass conserved.
- **Reads mode:** SA / sketch / single-end / stranded (`-l A` & explicit) + bias + bootstraps/Gibbs + `--skipQuant`.
- **Accuracy / concordance:** sim accuracy vs truth and 36 M cross-method concordance match the direct (reads-mode) path; piscem→RAD tracks salmon sketch.
- `cargo test --workspace`, `cargo fmt`, `clippy` all clean.

## Notes

- **Draft / not for merge yet** — opening for review; small follow-ups may be pushed.
- Targeted release **2.2.0**; the version bump + crates.io publish are handled by the release tooling (`bump_and_publish.sh`) at release time. Release notes: `docs/release-notes-2.2.0.md`.
- Builds against the published `libradicl` 0.14.0 (no local path patch).

https://claude.ai/code/session_01B7JMur5DmDpECddErpi2JS


rob-p:
Dear @BenjaminDEMAILLE,

  Please take a look at this when you have a chance. It fully implements RAD writing and reading and `--deterministic` mode. It incorporates some of your ideas / PRs, but also makes some modifications, as I think we found a way to achieve this even more efficiently. In general (depending on the dataset and hardware, etc.), the `--deterministic` inference is *even faster* than the standard one-pass online phase + offline phase.  To your earlier point, this may warrant reassessing if it should potentially be made the default mode or not (probably not in the initial release, so there is some time for users to test it, but maybe in the near future).
  
  In addition to some other optimizations (e.g. collecting FLD statistics during the mapping pass concurrent with RAD output, and building orientation aware equivalence classes that can be filtered at the end of mapping when we know the FLD), the big idea in the approach of this PR is that we quantize the range factorization bins and bias models in such a way that we can replace `f64` summation (which is not associative) with integer summation (which is), giving us an order independent output under (1) a *fixed* FLD and (2) *fixed* rough abundance estimates (which come from the naive equivalence classes and are relevant only in the bias-correction path).  These completely obviate the need for the RAD sort, serial inference, and thus, also, the name hash.
  
  Let me know if you have any specific thoughts or feedback. I plan to pull the trigger on this soon (maybe even today?), as it's undergone quite thorough testing. But if you have any thoughts, I'm happy to discuss.  Also, while the implementation ultimately evolved substantially, I'll definitely credit you for helping with this push in the release.
  
Best,
Rob

BenjaminDEMAILLE:
Thanks for pulling this together, @rob-p, and for the credit. The fixed-point integer accumulation is a clean way to get order-independence: it removes the `f64` associativity problem at the source instead of papering over it.

On the default-mode question, I'd actually lean toward making the deterministic mode the **default**, on FAIR grounds, specifically the "R" (Reusable) and the provenance expectation (R1.2). Today, with non-determinism as the default, the same input and the same parameters can produce different quantifications depending on thread count and scheduling. That makes every downstream result (DE calls, pipelines, published numbers) implicitly environment-dependent, which is exactly what undermines reusability and clean provenance for anything built on salmon's output.

A few reasons I think default-on is defensible:

- **Reproducibility should be the floor, not an opt-in.** Users don't expect `-p 8` vs `-p 16` to change their abundances; requiring them to ask for determinism inverts the principle of least surprise.
- **The cost looks modest.** Since order-independence comes from integer summation rather than a heavy re-architecture, the overhead should be small. If that holds in your benchmarks, the reproducibility win clearly outweighs it.
- **It future-proofs FAIR pipelines.** Archives and tools that cite salmon outputs can then assume reproducibility without each user knowing to flip a flag.

Your testing-window instinct is reasonable: ship it non-default in the initial release as you suggest, but with an explicit plan to flip it to default once the perf numbers are confirmed, and keep a `--non-deterministic` (or `--fast`) opt-out for users who knowingly trade reproducibility for raw speed. That way the *default* upholds FAIR and power users keep the escape hatch.

Happy to help benchmark the deterministic vs. non-deterministic path if that would help make the call.

PR #1041, opened 2026-06-30
feat(gpu): optional GPU backend for selective alignment (--gpu, deterministic, off by default)

BenjaminDEMAILLE:
## Summary

Adds an **optional GPU backend for selective-alignment validation** (the banded
Smith-Waterman that scores each candidate placement), behind a `gpu` cargo
feature and a `--gpu` flag. **Off by default**, so the standard binary is
unchanged and pulls in no GPU toolchain.

The headline property: `--gpu` produces a `quant.sf` **byte-identical** to the
CPU `--fullLengthAlignment` run. It changes *where* alignment runs, not the
result.

## How it works

- **New `salmon-gpu` crate** (feature-gated): a banded affine-gap DP expressed
  once as a pure-Rust reference and once as a WGSL compute shader, run via
  [`wgpu`](https://github.com/gfx-rs/wgpu) — **one portable shader on Metal
  (Apple) or Vulkan (Linux/NVIDIA)**, no CUDA/Metal SDK at build.
- An **`Aligner` trait** abstracts the alignment backend in `salmon-map`
  (`map_read_pair` is split into `collect_pair` → `PendingPair::{align_tasks,
  finalize}`); the quant pass collects every fragment in a mini-batch and scores
  all their candidate alignments in **one GPU dispatch**, then finalizes. When no
  aligner is set (the default), the existing in-place CPU path runs verbatim.
- `--gpu` implies `--fullLengthAlignment` (the batchable mode) and falls back to
  a CPU full-length backend if no GPU adapter is present.

## Determinism (the key guarantee)

Alignment scores are integer-valued, so the GPU reproduces the CPU backend
exactly. Proven by a chain of differential tests:

- the pure-Rust reference == `ksw2rs` (salmon's CPU aligner) on every
  score-affecting alignment, and
- the WGSL kernel == the reference, **bit-for-bit, verified on Metal**, and
- end-to-end, the batched path yields a `quant.sf` **byte-identical** to the
  inline full-length path (`batched_alignment_quant_is_identical_to_inline_full_length`).

So salmon's "identical output regardless of where the work runs" guarantee holds,
and the **default (non-GPU) path is untouched** (all existing tests green).

## Performance (honest)

Correct and deterministic today; **throughput is workload- and hardware-
dependent**. Each short-read banded DP is tiny, so the win comes from batching a
whole mini-batch into one dispatch. On Apple Silicon (unified memory, Metal) a
50k-alignment batch runs ~6.7× a single CPU thread; against a many-core CPU it is
**not yet a guaranteed net speedup**. This is an early, correctness-first backend
— base packing, buffer reuse, and wavefront parallelism are the next throughput
steps.

## Build & use

```sh
cargo build -p salmon-cli --release --features gpu
salmon quant -i idx -l A -1 r1.fq.gz -2 r2.fq.gz -p 16 -o out --gpu
```

Without the feature, `--gpu` is a clear error; without an adapter, it falls back
to CPU.

## CI

A macOS (Apple Silicon) job builds/tests the `gpu` feature (the GPU differential
test runs on real Metal); the default matrix never touches it, so the standard
binary stays GPU-toolchain-free.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,3 +85,33 @@ jobs:
             -a sample_data/sample_alignments.bam \
             -o target/sample_quant_alignment
           test -s target/sample_quant_alignment/quant.sf
+
+  # GPU alignment backend. Builds the wgpu/WGSL backend and runs its tests; on
+  # the Apple-Silicon runner Metal is present so the GPU differential test
+  # actually exercises the kernel (it skips gracefully if no adapter is found).
+  # The default build above never touches this feature, so the standard binary
+  # stays GPU-toolchain-free.
+  gpu-backend:
+    runs-on: macos-14
+    timeout-minutes: 60
+    steps:
+      - name: Checkout salmon
+        uses: actions/checkout@v4
+
+      - name: Install Rust ${{ env.RUST_TOOLCHAIN }}
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_TOOLCHAIN }}
+          components: clippy
+
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@v2
+
+      - name: Clippy (gpu feature)
+        run: cargo clippy -p salmon-gpu --features gpu --all-targets -- -D warnings
+
+      - name: Build CLI with --features gpu
+        run: cargo build -p salmon-cli --features gpu --release
+
+      - name: Test salmon-gpu (gpu feature)
+        run: cargo test -p salmon-gpu --features gpu --release
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,6 +55,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "anes"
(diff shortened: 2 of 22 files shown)


BenjaminDEMAILLE:
Ping when you get a chance, @rob-p. This one is fully opt-in (`--gpu`, off by default, deterministic output) and targets `develop`; CI is green including the `gpu-backend` job. No rush, just flagging it's ready whenever you'd like to take a look.

rob-p:
Marking this **postponed** rather than closed, and converting to draft — deliberately not a rejection.

The blocker is evidence, not design: a GPU backend is ~3k lines of new surface (kernel code, a device-selection path, a second implementation of the alignment inner loop that must be kept semantically locked to the CPU one forever after), and what would justify carrying that is a strong *end-to-end* empirical case, which the PR doesn't yet make. We've been burned before by kernel-level speedups that dissolve at the workload level — the recent hasher swap (#1126) showed 5–8% in isolation and 0% end-to-end — and an accelerator claim needs to clear a higher bar than a hasher: measured wall-clock on full real-data quantification runs, against the CPU path at realistic thread budgets on the same machine, across more than one GPU generation, with the determinism guarantee demonstrated (byte-identical `quant.sf` across launch configurations), and honest accounting of the H2D/D2H transfer and index-residency costs that mapping workloads are notorious for hiding.

If those measurements materialize and show a decisive win on hardware people actually run salmon on, this becomes very interesting — the deterministic two-phase structure we're converging on is, if anything, *friendlier* to a GPU mapping pass than the online path was. Until then it stays parked here (now labeled `postponed`, rebased against `develop`) rather than merged speculatively.

Issue #1038, opened 2026-06-25
Range start index error


o-william-white:
**Describe the bug**

When trying to run salmon quant for a small number of my fastq files, I get the following error:
```
2026-06-25T11:32:19.222965Z  INFO piscem_rs::index::reference_index: Loading SSHash dictionary from /mnt/shared/projects/nhm/owhite/krill_transcriptome_assembly_Esup/results/salmon_index/index.ssi
2026-06-25T11:32:20.011135Z  INFO piscem_rs::index::reference_index:   k=31, 15972783 strings, canonical=true
2026-06-25T11:32:20.011148Z  INFO piscem_rs::index::reference_index: Loading contig table from /mnt/shared/projects/nhm/owhite/krill_transcriptome_assembly_Esup/results/salmon_index/index.ctab
2026-06-25T11:32:20.268848Z  INFO piscem_rs::index::reference_index:   15972783 contigs, 70243308 entries, entry_width=41 bits
2026-06-25T11:32:20.268866Z  INFO piscem_rs::index::reference_index: Loading reference info from /mnt/shared/projects/nhm/owhite/krill_transcriptome_assembly_Esup/results/salmon_index/index.refinfo
2026-06-25T11:32:20.711185Z  INFO piscem_rs::index::reference_index:   2856569 references, max_ref_len=232365
2026-06-25T11:32:20.711409Z  INFO piscem_rs::index::reference_index: Loading equivalence class table from /mnt/shared/projects/nhm/owhite/krill_transcriptome_assembly_Esup/results/salmon_index/index.ectab
2026-06-25T11:32:20.909356Z  INFO piscem_rs::index::reference_index:   15972783 tiles, 12291065 ECs, 61707773 label entries
2026-06-25T11:32:20.909969Z  INFO piscem_rs::index::reference_index:   ref_shift=19, pos_mask=0x3ffff
2026-06-25T11:32:22.068380Z  INFO salmon_quant: mapping reads (selective-alignment mode, 8 threads)

thread '<unnamed>' (2346620) panicked at crates/salmon-map/src/align.rs:243:26:
range start index 245 out of range for slice of length 240
stack backtrace:
   0:     0x561078ff6a7a - <<std[52919eca6bce4da3]::sys::backtrace::BacktraceLock>::print::DisplayBacktrace as core[18c8dd30382e7099]::fmt::Display>::fmt
   1:     0x561078a9ad3a - core[18c8dd30382e7099]::fmt::write
[...]

PR #1042, opened 2026-06-30
perf(align): reuse query buffer and LUT-encode DNA5 (~14% faster full-length, byte-identical)

BenjaminDEMAILLE:
## Summary

Two **profiling-guided, behavior-preserving** optimizations to `align_chain`,
salmon's per-candidate selective-alignment validation (called millions of times).
Output is **byte-identical** (all map/quant tests pass, including the end-to-end
determinism test). Independent of any other change.

## What & why

Profiling a single-threaded reads run (1M pairs, Apple M4 Max, macOS `sample`)
showed `--fullLengthAlignment` is dominated by alignment (~65%), with the
per-candidate query allocation and DNA5 encoding a measurable share:

- `align_chain` built its oriented query with a fresh `read.to_vec()` /
  `revcomp()` **per candidate** → now reuses a thread-local buffer via a new
  `revcomp_into` helper (one fewer heap allocation per candidate).
- `dna5_encode` did a per-base `match` (a compare chain) → now uses a 256-entry
  lookup table, since encoding the read + reference window is on the inner loop
  of every full-length score.

## Result (measured, `-p 1`, 1M read pairs, M4 Max)

| Mode | before | after | |
|---|---|---|---|
| `--fullLengthAlignment` | ~6.55s | **~5.64s** | **~14% faster** |
| anchored (default) | ~1.72s | ~1.70s | unchanged (within noise) |

The default anchored path is unchanged on purpose: profiling shows its cost is
dominated by **chaining (~30%) and index hit-search (~28%)** — the latter in
external crates (`piscem-rs` / `sshash-lib`) — so there is little allocation
headroom there (alignment is only ~10% of the default path). This PR targets the
part that pays off.

## Tests

`cargo test -p salmon-map` and `cargo test -p salmon-quant` green; `quant.sf`
byte-identical before/after.

### DIFF
--- a/crates/salmon-map/src/align.rs
+++ b/crates/salmon-map/src/align.rs
@@ -74,18 +74,28 @@ pub struct Alignment {
     pub ref_window_start: i32,
 }
 
+/// Reverse-complement a DNA base (ACGT, case-insensitive; anything else → N).
+#[inline]
+fn comp(b: u8) -> u8 {
+    match b {
+        b'A' | b'a' => b'T',
+        b'C' | b'c' => b'G',
+        b'G' | b'g' => b'C',
+        b'T' | b't' => b'A',
+        _ => b'N',
+    }
+}
+
 /// Reverse-complement a DNA byte slice (ACGT, case-insensitive output upper).
 pub(crate) fn revcomp(s: &[u8]) -> Vec<u8> {
-    s.iter()
-        .rev()
-        .map(|&b| match b {
-            b'A' | b'a' => b'T',
-            b'C' | b'c' => b'G',
-            b'G' | b'g' => b'C',
-            b'T' | b't' => b'A',
-            _ => b'N',
-        })
-        .collect()
+    s.iter().rev().map(|&b| comp(b)).collect()
+}
+
+/// Reverse-complement `s` into `out` (cleared first), reusing its allocation.
+#[inline]
+fn revcomp_into(s: &[u8], out: &mut Vec<u8>) {
+    out.clear();
+    out.extend(s.iter().rev().map(|&b| comp(b)));
 }
 
 /// The perfect (all-match) score for a read of length `read_len`.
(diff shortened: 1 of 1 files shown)


rob-p:
Thanks @BenjaminDEMAILLE — this is a clean, well-scoped, profiling-guided change, and the shape is exactly right (thread-local query reuse + a DNA5 LUT, byte-identical). Two findings from reviewing and benchmarking it on our end:

**1. Base branch.** This targets `develop`, but the active Rust-rewrite line is `master` (`develop` is an older snapshot, ~89 commits behind). The patch doesn't cherry-pick onto `master` as-is — could you retarget/rebase onto `master`?

**2. Most of the win is already on `master`.** `develop`'s `dna5_encode` allocates a fresh `Vec` per call, which is the dominant cost your patch removes. `master` already replaced that with in-place `dna5_encode_into` / `dna5_encode_rev_into` (encode into a reused buffer), so that allocation is gone there. When I ported your two ideas onto `master` and benchmarked on x86_64 (`-p 1`, 1M read pairs), output stayed byte-identical but timing was neutral:

| mode | before | after |
|---|---|---|
| anchored (default selective alignment) | 23.7s | 23.0s (within noise) |
| `--fullLengthAlignment` | 84.3s | 84.1s |

So the ~14% is almost certainly real against `develop` / Apple silicon, but largely pre-captured on `master` for x86_64. The `ALIGN_QUERY` reuse in `align_chain` *is* still genuinely new on `master` (it still does a per-candidate `read.to_vec()`/`revcomp`), so rebasing onto `master` and re-measuring — especially the query-reuse part, and the LUT on ARM — is worth doing; there may still be something there.

Small heads-up if you rebase: `master`'s `dna5_encode_rev_into` only **reverses** the codes (it does not complement), so the LUT must be applied there without a complement step.

Thanks again for digging into the hot path!

BenjaminDEMAILLE:
Thanks Rob, both points addressed.

**Retargeted onto `master`.** The branch is now a single commit on top of current `master` (v2.3.0); the PR diff is just `crates/salmon-map/src/align.rs`. Since `master` already has the in-place `dna5_encode_into` / `dna5_encode_rev_into`, I dropped the develop-era allocation change and kept only the two ideas that are genuinely new here:

- **Query reuse in `align_chain`** — the per-candidate `read.to_vec()` / `revcomp(read)` now builds into a reused thread-local `ALIGN_QUERY` buffer (new `revcomp_into` helper). The #1038 out-of-bounds MEM guard is preserved, just evaluated against the reused buffer's length.
- **DNA5 LUT** — a 256-entry table replaces the per-base `match` in both encoders. Per your heads-up, `dna5_encode_rev_into` only *reverses* (no complement), so the per-base LUT drops in unchanged there; I added a comment noting that invariant.

**Byte-identical.** All `salmon-map` tests pass, and every per-case score checksum in `examples/bench_align` matches `master` exactly.

**Re-measured on ARM (Apple M4 Max)** via `examples/bench_align`, median of 3 interleaved runs, ns/aln (lower is better):

| case | path | master | this PR | speedup |
|---|---|--:|--:|--:|
| read_len=51  | full-window DP (`--fullLengthAlignment`) | 1445 | 1247 | ~14% |
| read_len=100 | full-window DP | 2732 | 2364 | ~13% |
| read_len=150 | full-window DP | 4057 | 3479 | ~14% |
| read_len=150, 2 mism | full-window DP | 4245 | 3561 | ~16% |
| read_len=51  | anchored (default) | 117 | 101 | ~13% |
| read_len=150 | anchored (default) | 188 | 178 | ~6% |

So your read is right: the develop-vs-master win was largely pre-captured on x86_64, but on Apple silicon the remaining query-reuse + LUT is a real, repeatable double-digit-% gain on the changed function (larger on the full-length path where the DNA5 encode is on the inner loop, and larger at short reads on the anchored path where the removed allocation is a bigger fraction), byte-identical. This is an `align_chain` microbenchmark isolating the touched code, not end-to-end wall time — happy to run an end-to-end `-p 1` pass on ARM too if you'd like that number for the record.

BenjaminDEMAILLE:
For context, this is part of a small set of companion hot-path perf PRs across the COMBINE-lab Rust stack that salmon builds on — same profiling pass, same "kill per-read/per-candidate setup cost, byte-identical output" approach:

- **COMBINE-lab/piscem-rs#3** — perf(hit_searcher): halve `ReadKmerIter` per-read setup cost
- **COMBINE-lab/sshash-rs#2** — perf(offsets): avoid recomputing boundary offsets in `locate_with_end`

They're independent and can be reviewed/merged in any order; flagging them together since they came out of the same investigation.

rob-p:
Merged and released in **salmon 2.3.1** (crates.io + tag `v2.3.1`). Folded together with the two companion PRs (piscem-rs#3, sshash-rs#2) — byte-identical output; on our x86_64 default selective-alignment path the combined mapping win is ~2.5% at `-p1` (neutral at `-p16`, memory-bound at scale), with the larger gains on ARM as you measured. Thanks @BenjaminDEMAILLE!

PR #1051, opened 2026-07-09
perf(infer): profiling harness + optional SQUAREM EM/VBEM acceleration

BenjaminDEMAILLE:
# Profiling harness + optional SQUAREM EM/VBEM acceleration

Two additive, **opt-in** performance changes to the EM/VBEM inference path, plus a reproducible profiling harness. **No default behaviour changes**: the shipped binary and default `salmon quant` output are unaffected.

This PR is accompanied by a short literature review, because the acceleration it adds (SQUAREM) is a specific, well-motivated choice from the numerical-optimization literature, and the same review frames the follow-up (#1052, DAAREM).

## 1. Motivation

Salmon estimates transcript abundances with an EM (or VBEM under `--useVBOpt`) over equivalence classes [Patro et al. 2017]. That step is a **fixed-point iteration** `x <- F(x)` that converges only **linearly**, and slowly when the fraction of "missing information" is large, which is exactly the multi-mapping structure of transcript quantification. The cost is paid again for every bootstrap/Gibbs replicate. The goal here is to accelerate the inference **without touching the model** (same fixpoint, same accuracy), plus give us the instrumentation to measure it.

## 2. Literature review: accelerating EM-like fixed-point iterations

**SQUAREM** [Varadhan & Roland 2008] is the accelerator this PR adopts. Treating one EM update as a fixed-point map `F(x)`, from two successive maps it forms difference vectors `r`, `v` and takes a squared extrapolation step `x' = x - 2*alpha*r + alpha^2*v`, then one stabilizing EM map. It is genuinely off-the-shelf (you supply only `F`), adds negligible memory, is typically superlinear, and commonly needs 1-2 orders of magnitude fewer EM evaluations on high-missing-information problems. It is also what the original C++ `CollapsedEMOptimizer` used and what the Rust port had left deferred, so this restores parity.

For context on where SQUAREM sits: it uses only the last two iterates. History-based multi-secant methods, **Anderson acceleration** [Anderson 1965; Walker & Ni 2011], its hardened **DAAREM** form [Henderson & Varadhan 2019], and **quasi-Newton EM** [Zhou, Alexander & Lange 2011], can converge faster on higher-dimensional, ill-conditioned problems, at a higher per-iteration cost. A recent survey [Kuroda 2023] finds no universal winner: SQUAREM's advantage is simplicity and near-zero overhead; the multi-secant methods win on the hardest problems. SQUAREM is therefore the right first accelerator to land (safe, cheap, default-off); DAAREM is offered as a follow-up in #1052.

*(The standard model-level levers, equivalence-class collapse [Bray et al. 2016], stochastic collapsed VB/VBEM [Patro et al. 2017], and range-factorization [Zakeri et al. 2017], are already inside salmon; the remaining algorithmic headroom for the inference kernel is the accelerator itself.)*

### Profiling harness (`build(profiling)`)
- New `[profile.profiling]` (release + line-table debug info, unstripped) so `samply`/`flamegraph` can attribute samples; the shipped `release`/`dist` binaries and the `.cargo/config.toml` target-cpu floor are untouched.
- Per-phase wall-clock via a `PhaseTimer` on a dedicated `salmon::timing` tracing target (index_load / mapping / eff_length_collapse / em_bias / posterior / output). Pure logging, never touches output or determinism.
- A `criterion` micro/convergence benchmark for the EM/VBEM M-step (`crates/salmon-infer/benches/em_step.rs`) over a seeded power-law equivalence-class fixture.
- `scripts/profile.sh` (samply harness; a `sample_data.tgz` smoke tier plus a documented GIAB tier).

### Optional SQUAREM acceleration (`feat(infer)`)
- New `--emAccel <none|squarem>` (default `none`). `squarem` drives an SqS3 cycle in `run_em_counts` that **reuses the existing sharded M-step kernels** as the fixed-point map `F` (two M-steps + a stabilizing third; steplength clamped to `a <= -1`; non-negativity backoff to the plain double-step).
- The plain path is refactored to share one `F` closure but is **byte-identical** to before. The norm/vector math is sequential for run-to-run determinism regardless of thread count.
- Benefits the point estimate and every bootstrap replicate; Gibbs is untouched (it uses no EM M-step).

## 4. Correctness and measurements

Validated on GIAB **NA12878 / HG001** (ENA `ERR356372`) against a GENCODE v50 index:

- **Same fixpoint** as plain EM: on identical (deterministic-mapping) equivalence classes, TPM Pearson ~ 0.99999 with identical total TPM; residual differences are at the convergence-tolerance level on negligible-abundance transcripts. This is why SQUAREM is **not** byte-identical and is therefore opt-in / default-off, existing goldens and `--deterministic` output are unchanged. Unit tests assert same-fixpoint agreement (EM and VBEM), mass conservation, and a slow-mixer case.
- **Speed** (measured on a 4M-read-pair GIAB subset): offline EM ~7% faster and a 30x bootstrap ~7% faster. The bigger gains show up on slow-mixing / tight-tolerance problems; the warm-started default EM already converges quickly, so the everyday gain is modest but free.

## 5. Verification

Green under the pinned CI toolchain (1.91.1): `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --release`.

Happy to split the two commits into separate PRs if you prefer to review the profiling harness and the SQUAREM change independently.

## References

All confirmed against publisher DOIs; none included on memory alone.

- Anderson, D. G. (1965). Iterative procedures for nonlinear integral equations. *Journal of the ACM* 12(4):547-560. https://doi.org/10.1145/321296.321305
- Bray, N. L., Pimentel, H., Melsted, P., & Pachter, L. (2016). Near-optimal probabilistic RNA-seq quantification. *Nature Biotechnology* 34:525-527. https://doi.org/10.1038/nbt.3519
- Henderson, N. C., & Varadhan, R. (2019). Damped Anderson acceleration with restarts and monotonicity control for accelerating EM and EM-like algorithms. *Journal of Computational and Graphical Statistics* 28(4):834-846. https://doi.org/10.1080/10618600.2019.1594835
- Kuroda, M. (2023). Acceleration of the EM algorithm. *WIREs Computational Statistics* 15(4):e1618. https://doi.org/10.1002/wics.1618
- Patro, R., Duggal, G., Love, M. I., Irizarry, R. A., & Kingsford, C. (2017). Salmon provides fast and bias-aware quantification of transcript expression. *Nature Methods* 14:417-419. https://doi.org/10.1038/nmeth.4197
- Varadhan, R., & Roland, C. (2008). Simple and globally convergent methods for accelerating the convergence of any EM algorithm. *Scandinavian Journal of Statistics* 35(2):335-353. https://doi.org/10.1111/j.1467-9469.2007.00585.x
- Walker, H. F., & Ni, P. (2011). Anderson acceleration for fixed-point iterations. *SIAM Journal on Numerical Analysis* 49(4):1715-1735. https://doi.org/10.1137/10078356X
- Zakeri, M., Srivastava, A., Almodaresi, F., & Patro, R. (2017). Improved data-driven likelihood factorizations for transcript abundance estimation. *Bioinformatics* 33(14):i142-i151. https://doi.org/10.1093/bioinformatics/btx262
- Zhou, H., Alexander, D., & Lange, K. (2011). A quasi-Newton acceleration for high-dimensional optimization algorithms. *Statistics and Computing* 21(2):261-273. https://doi.org/10.1007/s11222-009-9166-3

*AI-assisted-research disclosure: the literature search underlying this review was conducted with AI research agents; every reference was verified independently against its publisher DOI, with no fabrication or mashups.*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2607,6 +2607,7 @@ dependencies = [
 name = "salmon-infer"
 version = "2.3.2"
 dependencies = [
+ "criterion",
  "rand 0.8.6",
  "rand_distr",
  "rand_pcg",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,6 +102,16 @@ lto = "thin"
 codegen-units = 1
 panic = "abort"
 
+# Like `release`, but keeps line-table debug info and does not strip, so
+# `samply`/`cargo flamegraph` can attribute samples to functions/lines. The
+# shipped `release`/`dist` binaries are unaffected (they stay stripped), and the
+# same portable `target-cpu` floor from `.cargo/config.toml` still applies — we
+# profile the binary users actually run. Build: `cargo build --profile profiling`.
+[profile.profiling]
+inherits = "release"
+debug = "line-tables-only"
+strip = false
+
 [profile.bench]
 lto = "thin"
 
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -22,7 +22,7 @@ use salmon_align::{
     GenomeProjectOptions,
 };
 use salmon_index::{build as build_index, IndexBuildOptions};
-use salmon_quant::{quantify, ChunkCodec, ProgressCounters, QuantOptions};
+use salmon_quant::{quantify, ChunkCodec, EmAccel, ProgressCounters, QuantOptions};
 
 use std::io::IsTerminal;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -149,6 +149,25 @@ struct Cli {
     /// `SALMON_NO_VERSION_CHECK` environment variable) is a no-op.
     #[arg(long = "no-version-check", global = true)]
     no_version_check: bool,
+}
+
+/// CLI surface for [`EmAccel`]: the EM/VBEM convergence acceleration scheme.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, clap::ValueEnum)]
+enum EmAccelArg {
+    /// Plain fixed-point iteration; output unchanged from historical salmon.
+    #[default]
+    None,
+    /// SQUAREM acceleration (same fixpoint, far fewer M-steps; not byte-identical).
+    Squarem,
+}
+
+impl From<EmAccelArg> for EmAccel {
+    fn from(a: EmAccelArg) -> Self {
+        match a {
+            EmAccelArg::None => EmAccel::None,
+            EmAccelArg::Squarem => EmAccel::Squarem,
+        }
+    }
 }
 
 // clap subcommand args structs differ in size (IndexArgs vs the larger QuantArgs);
@@ -497,6 +516,10 @@ struct QuantArgs {
     /// VBEM per-feature Dirichlet prior weight.
     #[arg(long = "vbPrior", default_value_t = 1e-2)]
     vb_prior: f64,
+    /// EM/VBEM convergence acceleration. `squarem` reaches the same abundances in
+    /// far fewer M-steps but is not byte-identical to the default `none`.
+    #[arg(long = "emAccel", value_enum, default_value_t = EmAccelArg::None)]
+    em_accel: EmAccelArg,
     /// Mean of the fragment-length distribution prior.
     #[arg(long = "fldMean", default_value_t = 250.0)]
     fld_mean: f64,
@@ -1127,6 +1150,7 @@ fn run_quant(args: QuantArgs, quiet: bool) -> Result<()> {
         opts.incompat_prior = args.incompat_prior;
         opts.em.vb_prior = args.vb_prior;
         opts.em.per_nucleotide_prior = args.per_nucleotide_prior;
+        opts.em.accel = args.em_accel.into();
         opts.sig_digits = args.sig_digits;
         opts.fld_mean = args.fld_mean;
         opts.fld_sd = args.fld_sd;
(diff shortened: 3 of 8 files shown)


rob-p:
Hi @BenjaminDEMAILLE : I'm a bit confused here. The building and feat commits seem to make sense, but the description here refers to building `hisat2`. Was this PR message intended for another repo / tool?

BenjaminDEMAILLE:
Apologies for the confusion, Rob — that was a copy-paste error: the original description was pasted from an unrelated HISAT2 ARM-build PR. I've corrected the PR body; it now matches the two commits here: (1) a profiling harness (a `profiling` cargo profile, per-phase `salmon::timing` spans, a criterion EM benchmark, and `scripts/profile.sh`), and (2) an opt-in SQUAREM accelerator for the EM/VBEM optimizer (`--emAccel squarem`, default off, same fixpoint as plain EM, not byte-identical so gated behind the flag). No default-behavior changes. Thanks for catching it.

rob-p:
Follow-up: the `PhaseTimer` / `salmon::timing` instrumentation added here originally lived only in the reads quant driver (`salmon-quant`), so `--deterministic`, `--rad`, and `-a` (alignment) — which quantify via `salmon-align`'s `quantify_rad` / `quantify_alignments` — emitted no phases. In particular the `em_bias` (EM/VBEM) phase was invisible in exactly the modes used to benchmark the SQUAREM/DAAREM accelerators from this PR and #1052.

I've extended it so all quant paths are covered: `PhaseTimer` is hoisted into `salmon-core` (shared by `salmon-quant` and `salmon-align`) and both align-side drivers are instrumented — `quantify_rad` (rad_read / em_bias / posterior / output) and `quantify_alignments` (mapping / em_bias / posterior / output). Pure instrumentation, no output/determinism change; verified all four paths (reads online, reads `--deterministic`, `-a` online, `-a --deterministic`) emit under `RUST_LOG=salmon::timing=info`.

Merged into `develop` (commit 6c6f93f3). Flagging here since it's a direct extension of this PR's profiling harness.

PR #1056, opened 2026-07-11
fix(infer): thread-count-independent EM reduction on the deterministic paths (fixes flaky thread_count_invariant)

rob-p:
## Problem

The parallel EM/VBEM M-step shards the equivalence classes by
`rayon::current_num_threads()` and sums each transcript's mass across the shards.
f64 addition is non-associative, so the summation **grouping** — and the
converged estimate to the last bit — depends on the runtime thread count. On
x86_64 the groupings round identically; on **aarch64** they occasionally don't,
which made the RAD path's `thread_count_invariant` test (asserts bit-exact
counts across `-p`) **flaky in CI** (e.g. the v2.3.3 release-bump run failed on
`ubuntu-24.04-arm` while the identical commit passed on x86_64).

## Fix

Make the M-step reduction **thread-count-independent** — but only where a
cross-thread byte-identity guarantee actually exists, and without a perf hit.

**1. Confine to the deterministic paths.** New `EmOptions.deterministic_reduction`
selects a **fixed** shard count (`nclasses.clamp(1, 64)`) instead of one derived
from the thread count, so the reduction grouping (and the result) is identical at
any `-p`. It's set by `quantify_rad` and `quantify_alignments` (`--deterministic`
/ `--rad` / alignment mode). The **default online path is untouched** — its
output isn't thread-count-identical anyway (the FLD is built with a
thread-dependent online estimator), so it keeps the fast thread-adaptive shards.

**2. Keep it fast (sparse M-step).** A fixed high shard count with the old dense
per-shard buffers zeroed and reduced `O(nshards × num_txps)` per iteration — up
to **3.2× slower** EM. The new `ShardCtx` precomputes each shard's touched
transcript ids **once** (the eq-classes are constant across EM iterations) and
zeroes/reduces only those → `O(incidences)`. This is **value-identical** (adding
the skipped zeros cannot change an f64 sum), so it's a pure speedup.

## Results (5M-read human sample, `em_bias` phase)

| | naive fixed-64 | **fixed-64 + sparse (this PR)** |
|---|---|---|
| deterministic-path EM, `-p 16` | +96% | **+4%** |
| deterministic-path EM, `-p 1` | +222% | **+19%** |
| **default online EM** | (n/a) | **unchanged** (±noise) |

- **Determinism:** `--deterministic` output is now byte-identical across `-p 1/8/16`
  **by construction** (so aarch64-safe, not luck); `thread_count_invariant` passes
  20/20 locally (was flaky).
- **Output unchanged:** `quant.sf` is identical to before (the last-ULP grouping
  differences round away at `sigDigits`); the default online path is unchanged.
- Full workspace tests (incl. all EM/VBEM correctness tests) pass; `fmt` +
  `clippy -D warnings` clean.

The remaining ~19% at `-p 1` is inherent to a fixed 64-shard count on a single
thread (uncommon for real quant, and only in opt-in reproducible mode); it can be
tuned via `EM_MAX_SHARDS` if desired.

https://claude.ai/code/session_01B7JMur5DmDpECddErpi2JS

### DIFF
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -1461,6 +1461,14 @@ pub fn quantify_alignments(opts: &AlignQuantOptions) -> Result<AlignQuantResult>
     let start_time = asctime_now();
     let run_timer = std::time::Instant::now();
     let mut timer = PhaseTimer::new();
+    // Alignment-mode eq-classes are order-independent; run the parallel M-step
+    // with a thread-count-independent reduction so the point estimate is
+    // byte-identical across `-p`.
+    let det_em = {
+        let mut e = opts.em.clone();
+        e.deterministic_reduction = true;
+        e
+    };
     let header = read_alignment_header(&opts.bam)?;
 
     // Reject coordinate-sorted input up front (header-only check, no per-record cost):
@@ -1676,12 +1684,12 @@ pub fn quantify_alignments(opts: &AlignQuantOptions) -> Result<AlignQuantResult>
             dropped_mass: 0.0,
         }
     } else if opts.init_uniform {
-        optimize(&collapsed, num_refs, &opts.em, Some(&eff_lengths))
+        optimize(&collapsed, num_refs, &det_em, Some(&eff_lengths))
     } else {
         optimize_with_init(
             &collapsed,
             num_refs,
-            &opts.em,
+            &det_em,
             init_alphas.as_deref(),
             Some(&eff_lengths),
         )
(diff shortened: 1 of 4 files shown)


rob-p:
Closing as superseded rather than merged. The flakiness this fixed was real, but the property it restructured the reduction to obtain — bit-identical counts across thread counts — is invisible at the precision `quant.sf` reports, and the test now asserts what consumers can actually observe.

Measured on a deliberately hard fixture (800 transcripts, 60k fragments, 1–4-way ambiguity, 2–32 threads, 50 runs): maximum cross-thread deviation **4.5e-13**, against a reporting step of 5e-4 for `NumReads` and 5e-7 for TPM. Per-iteration reordering error is bounded by `nshards·ε·Σcounts` and cannot compound because each EM iteration renormalises. `thread_count_invariant` now gates at 1e-9 — six orders stricter than anything visible in the output, ~2000× above measured associativity noise, so a genuine determinism regression still fails loudly. Repeated single-threaded runs are still required to be bit-identical, and a measurement probe (`measure_thread_count_fp_deviation`, behind `--ignored`) stays in the tree for re-running if the reduction strategy changes.

Landed in #1119 (`657ccf46`); 15/15 repeated runs pass. Thanks for the investigation here — it correctly identified the mechanism (shard-count- and stealing-dependent addition order in `reduce_shards`), which is exactly what made the bounded-tolerance argument possible.

PR #1063, opened 2026-07-26
perf(align): parallelize bias effective-length correction

Hugo-Polloli:
## Summary

Alignment and RAD quantification apply bias-corrected effective lengths with the same per-transcript `corrected_effective_length_full` calculation as reads mode, but `salmon-align::apply_bias_correction` ran that sweep serially. On a large transcript catalog this left the abundance phase pinned to one core after the bias models had been built.

This PR:

- runs the independent per-transcript corrections through Salmon's existing Rayon pool;
- leaves the arithmetic within each transcript and the ordering of `eff_lengths` unchanged;
- preserves the dedicated GC-only `ref_seqs` regression; and
- adds a separate all-bias test requiring finite effective lengths and exact 1-thread/8-thread invariance for effective lengths, counts, TPMs, and all dumped bias models.

There is no public API, file-format, model, or default-option change.

## Scope relative to earlier bias work

The reads-mode driver already uses `into_par_iter()` for this correction sweep. This does not revisit the expected-model construction work discussed in #1020 and folded into #1017; it changes only the separate serial sweep in the later alignment/RAD `apply_bias_correction` path.

## Why the parallelism is output-preserving

Each iteration reads shared immutable bias models and reference sequence, performs the existing calculation for one transcript, and writes one disjoint `eff_lengths[tid]` slot. There is no cross-transcript reduction, so thread scheduling cannot change floating-point evaluation order within a transcript.

## Performance

Measured with a release build on a deterministic real-data RAD subset containing 471,017 records while retaining the full 655,022-reference catalog and reference/bias workload:

| threads | correction sweep |
| ---: | ---: |
| 1 | 42.306 s |
| 2 | 22.638 s |
| 4 | 13.438 s |
| 8 | 8.508 s |
| 16 | 6.381 s |

That is 6.6x scaling for the corrected-effective-length sweep at 16 threads.

In an interleaved A-B-B-A comparison at 16 threads:

| measurement | serial baseline | parallel candidate | change |
| --- | ---: | ---: | ---: |
| bias/abundance phase mean | 86.98 s | 51.35 s | -41.0% |
| end-to-end wall mean | 124.95 s | 92.35 s | -26.1% |

Peak RSS was effectively unchanged (approximately 10.46 million KiB in both cases). `quant.sf` and `quant.genes.sf` were byte-identical; the 1/2/4/8/16-thread candidate runs also produced the same quantification SHA-256.

As a larger supporting check, one 17.4M-fragment run decreased from 7:31.6 to 5:10.0 and produced identical quantification and bias-model artifacts. I treat that timing only as supporting evidence because unrelated background services made the host load less controlled than the A-B-B-A run above.

## Verification

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace --release` (all tests pass; one profiling benchmark remains ignored as intended)
- `git diff --check origin/master...HEAD`
- the all-bias regression requires finite effective lengths and exact 1-thread/8-thread equality
- exact real-data output parity across thread counts and against the serial baseline

AI-assisted implementation disclosure: OpenAI Codex assisted with implementation, review, testing, and benchmarking; the reported commands and artifacts were executed and checked locally.

### DIFF
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -1431,44 +1431,53 @@ fn apply_bias_correction(
         bias_dump.exp3_pos = masses(erc);
     }
 
-    for tid in 0..num_refs {
-        if alphas[tid] < 1e-8 {
-            continue;
-        }
-        let s = ref_bytes[tid].as_slice();
-        let pos_vecs: Option<(Vec<f64>, Vec<f64>)> =
-            pos_models.as_ref().map(|(ofw, orc, efw, erc)| {
-                let lc = length_class.expect("positional bias requires length classes")[tid];
-                let rl = s.len();
-                let (mut o5, mut e5) = (vec![0.0; rl], vec![0.0; rl]);
-                let (mut o3, mut e3) = (vec![0.0; rl], vec![0.0; rl]);
-                ofw[lc].project_weights(&mut o5);
-                efw[lc].project_weights(&mut e5);
-                orc[lc].project_weights(&mut o3);
-                erc[lc].project_weights(&mut e3);
-                (
-                    salmon_model::positional_factor(&o5, &e5),
-                    salmon_model::positional_factor(&o3, &e3),
-                )
-            });
-        let bias = salmon_model::BiasInputs {
-            seq: seq_tab.as_ref().map(|(f, r)| (f, r)),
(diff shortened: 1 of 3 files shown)


Hugo-Polloli:
I'll verify the failing job asap
(fixed)

rob-p:
Thanks!

PR #1072, opened 2026-08-01
feat(quant): decode gzipped FASTQ in parallel with rapidgzip

BenjaminDEMAILLE:
## What

Reads-mode FASTQ input is gunzipped by flate2's `MultiGzDecoder`, one serial inflate stream per file. On a many-core run that decoder becomes the producer bottleneck well before the mapping threads saturate.

This routes `.gz` FASTQ through [`rapidgzip-core`](https://github.com/COMBINE-lab/rapidgzip-rust) (0.1), the COMBINE-lab Rust port of rapidgzip. It decodes gzip, concatenated members and BGZF across a worker pool and returns a `Read + Send` stream that paraseq owns directly, so nothing downstream changes. Every member is still CRC- and size-verified.

## Thread budget

`--gzipDecompressThreads <N>` (new). The budget is a **whole-run** ceiling split across the input files, since paraseq opens them all up front and decodes them concurrently; a per-file ceiling would multiply by the file count.

Unset, it derives from `--threads`, bounded by the cores the process can actually run on, and capped at 8: one worker decodes far more FASTQ than one mapping thread consumes, so past that the cores do more good mapping reads. rapidgzip creates workers lazily and retires them under consumer backpressure, so the value is a cap and not an allocation. An explicit value is used as given.

`--gzipDecompressThreads 0` selects the old serial decoder for the whole run.

## Fallback

The parallel decoder needs positional (`pread`) access to a stable file, so it is used only for regular files. Gzipped input that is not a regular file (process substitution, a FIFO, `/dev/stdin`) falls back to `MultiGzDecoder`, as does input whose framing rapidgzip rejects up front, which lets flate2 produce the same error it always did for malformed input. Failures found mid-stream (bad CRC or member footer) surface as read errors and cannot fall back.

Uncompressed FASTQ and every non-reads path (alignment mode, RAD, reference FASTA, SAM.gz) are untouched.

## Tests

- `gz_budget_is_split_across_concurrently_read_files` — budget derivation: split across files, floor of 1, cap at 8, bounded by cores, explicit value passed through, 0 = opt out.
- `parallel_and_serial_gzip_decode_identically` — both decoders yield identical bytes across a concatenated (multi-member) gzip boundary.
- `non_regular_gzip_input_falls_back_to_serial` — a FIFO decodes instead of failing.
- `gzipped_reads_quantify_identically_to_plain_reads` (end-to-end) — plain, rapidgzip and flate2 runs of the same fixture give identical `num_mapped` and per-transcript counts; mapping is single-threaded so any difference could only come from decompression.
- `gzip_decompress_threads_is_optional_and_accepts_zero` — CLI parsing, including that 0 is accepted here (unlike `--bamCompressThreads`).

`cargo fmt`, `cargo clippy --workspace --all-targets` and `cargo test --workspace` are clean.

## Docs

`--gzipDecompressThreads` documented in `website/.../reference/cli.mdx` and `docs/cli-flag-status.md`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1650,6 +1650,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "libz-rs-sys"
+version = "0.6.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "50474818739ccab820cd57bca432d6b02d090b47f9e85501d963cd05851f82eb"
+dependencies = [
+ "zlib-rs",
+]
+
+[[package]]
 name = "linux-raw-sys"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2467,6 +2476,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "rapidgzip-core"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1c8915419d173342f3f9f104024f8f67a378944fdef494958d2ebdd756fe4e8a"
+dependencies = [
+ "crossbeam-deque",
+ "libz-rs-sys",
+]
+
+[[package]]
 name = "rapidhash"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2770,6 +2789,7 @@ dependencies = [
  "noodles-sam",
  "paraseq",
  "piscem-rs",
+ "rapidgzip-core",
  "rayon",
  "salmon-core",
  "salmon-eqclass",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,8 @@ xxhash-rust = { version = "0.8", features = ["xxh3", "xxh64"] }
 # compression
 flate2 = "1"
 zstd = "0.13"
+# parallel gzip decoding for FASTQ input (COMBINE-lab port of rapidgzip)
+rapidgzip-core = "0.1"
 # logging
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -484,6 +484,20 @@ struct QuantArgs {
         conflicts_with = "write_mappings"
     )]
     bam_compress_threads: Option<std::num::NonZeroUsize>,
+    /// Maximum worker threads used to decompress gzipped FASTQ input, which
+    /// salmon decodes in parallel with rapidgzip. The budget covers the whole
+    /// run and is split across the input files, since they are read
+    /// concurrently. The default derives from --threads, bounded by the cores
+    /// this process can run on and capped at 8: one worker already decodes far
+    /// more FASTQ than one mapping thread consumes, so beyond that the cores do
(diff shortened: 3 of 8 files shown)


rob-p:
Thanks for this — the analysis is right, the fallback matrix is right, and two of its ideas are now upstream. Closing it as an implementation, and I want to explain why that isn't a judgement on the work.

## Why not merge

salmon already does `use piscem_rs::io::fastx::{reader_with_batch_size, Collection, CollectionType}` — it uses piscem-rs for readers and only opens the file itself. As of piscem-rs 0.8.0, that crate exposes `io::fastx::{open_input, plan_thread_budget}` and the `io::calibrate` module, which together are the same decision this PR makes, already built and measured. So the choice is not between two designs; it is between one implementation and two. Wiring salmon to the upstream one is roughly a thirty-line change against 336 lines of a second implementation that would then need to track the first.

That is the whole reason, and it only became true after this PR was opened — piscem-rs 0.7.1 exposed those APIs specifically so salmon could use them, and 0.8.0 corrected them.

## What was taken from here

**The `available_parallelism` bound.** Your point that the budget should be "bounded by the cores the process can actually run on" was a real gap upstream: piscem-rs bounded by `-t` alone, so under a cpuset or CPU quota it would budget decode threads against cores the process does not hold and oversubscribe what is left for mapping. Fixed in 0.7.1 and credited to this PR in the 0.8.0 release notes.

**The whole-run budget framing.** "A per-file ceiling would multiply by the file count" is exactly the failure mode piscem-rs hit and only fixed after an oscillation bug; you got it right first time. Upstream already had that shape — a run-wide total the supervisor enforces, with per-file ceilings deliberately above fair share so a starved decoder can borrow from idle peers — so nothing changed, but it is worth saying that this PR independently arrived at the right structure.

## Where the approaches genuinely differ

The one substantive difference is the cap at 8, derived from `--threads`. It is a fitted constant, and upstream had the same problem in a different place — a threads-per-file threshold justified as a logical bound but which quietly embedded a measured consumer rate. Both are fit to one machine and neither transfers.

piscem-rs 0.8.0 removes its version and measures instead: a probe runs the pipeline's shape in miniature — one producer feeding a bounded queue, consumers draining it — and reports which side blocks. It samples in windows and stops when the verdict is unambiguous or has stopped moving, so a clear-cut input pays ~76 ms and only genuinely marginal ones approach the 1 s ceiling. Validated at 6 of 8 on a measured crossover surface, correct in both regimes that matter and conservative in the one that is a wash.

That matters for salmon specifically, and in a way a constant cannot express: selective alignment consumes roughly a quarter of what sketch mode does per unit time — 256 s of mapping-only CPU against 60 s for the same 8M reads — so the same input and thread count can be decode-bound in one mode and comfortably fed in the other. The probe takes the mapping closure, so whichever mode is about to run is the consumer it measures. No mode plumbing, no classification.

## What happens next

salmon's own wiring is in progress on `develop`: `open_input` in place of `MultiGzDecoder`, a `--decoder` flag matching piscem's spelling, and a measurement of whether SA and sketch actually land on different sides. If they do not, that will be reported too.

If you want to keep going here, the SA-versus-sketch measurement is the open question, and `--dict auto`'s threshold is a genuinely unresolved one: the tiny/sshash ranking inverts by workload, so the current fixed k-mer cutoff is wrong in one regime or the other and nobody has measured which matters more in practice.

Issue #1075, opened 2026-08-06
Empty quant.genes.sf with Ensembl cDNA + GTF: version-mismatched IDs match nothing, and the summary line reports success


BenjaminDEMAILLE:
With an Ensembl cDNA index and an Ensembl GTF — the reference combination for human bulk RNA-seq — `quant.genes.sf` comes out containing nothing but its header, and the run exits 0.

```
wc -l quant.sf         # 515983
wc -l quant.genes.sf   #      1
```

salmon 2.4.1, official `aarch64-apple-darwin` binary. Reference: Ensembl release 115, human (cDNA FASTA + GTF), GTF decompressed (see the companion issue for why it has to be).

## Cause: versioned vs unversioned transcript IDs

The two files disagree on identifier format, and the join is a literal string comparison.

| Source | Identifier |
|---|---|
| `quant.sf`, from the Ensembl cDNA FASTA | `ENST00000622028.1` |
| Ensembl GTF attributes | `transcript_id "ENST00000622028"` + `transcript_version "1"` |

Ensembl keeps the version in a separate GTF attribute while the cDNA FASTA header carries it inline. [`read_transcript_gene_map`](https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-core/src/genemap.rs#L38) takes `transcript_id` alone, so every key is unversioned; [`write_gene_quant`](https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-core/src/genemap.rs#L106) then does `gene_map.get(name)` against versioned names. Nothing matches, ever — not a subset, all 515,982.

Reproduced against the current reader with a two-transcript GTF:

```
map_entries=2  unmapped=2  lines_in_quant.genes.sf=1
file_content="Name\tLength\tEffectiveLength\tTPM\tNumReads\n"
```

## The diagnostics actively mislead

I want to correct something I assumed before reading the code: salmon is **not** silent here. [`main.rs:878`](https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-cli/src/main.rs#L878) does warn:

```
warning: 515982 transcript(s) had no entry in the gene map and were omitted from quant.genes.sf
```

The problem is the line printed immediately after it, [`main.rs:883`](https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-cli/src/main.rs#L883):

```rust
let n_genes = map.values().collect::<std::collections::HashSet<_>>().len();
...
[...]

PR #1076, opened 2026-08-06
fix(genemap): accept gzip-compressed GTF/GFF for --geneMap

### LINKED ISSUE #1074: --geneMap rejects gzip-compressed GTF/GFF, and only fails after the run completes
`--geneMap` rejects a gzip-compressed GTF, and only after the whole run has finished.

Gzip is the default distribution format for both Ensembl and GENCODE annotations, so passing `Homo_sapiens.GRCh38.115.gtf.gz` straight from the download is the obvious thing to do. It costs a full run to find out it is not supported.

## Reproducing

salmon 2.4.1, official `aarch64-apple-darwin` binary. Reference: Ensembl release 115, human (cDNA FASTA + GTF).

```
salmon quant -i <index> -l A -1 R1.fq.gz -2 R2.fq.gz -o out \
  --geneMap Homo_sapiens.GRCh38.115.gtf.gz
```

```
Error: reading gene map <path>/Homo_sapiens.GRCh38.115.gtf.gz

Caused by:
    stream did not contain valid UTF-8
```

The exit is non-zero, so in a pipeline the step is a failure even though every expensive artifact was already produced.

## Why it lands at the end of the run

Gene-level aggregation is the last thing `salmon quant` does. In [`crates/salmon-cli/src/main.rs:1612`](https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-cli/src/main.rs#L1612), `write_gene_level` is called *after* mapping, EM and `quant.sf` have all completed — and `write_gene_level` opens the gene map as its first action ([`main.rs:865`](https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-cli/src/main.rs#L865)).

So the gene map is never validated up front. On a human dataset that is several minutes of mapping and inference thrown away for a filename the tool could have rejected — or handled — immediately.

## Root cause

[`read_transcript_gene_map`](https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-core/src/genemap.rs#L16) in `crates/salmon-core/src/genemap.rs` opens the path as plain text and iterates lines:

```rust
let reader = io::BufReader::new(std::fs::File::open(path)?);   // genemap.rs:23
...
for line in reader.lines() {
    let line = line?;                                          // genemap.rs:27  ← InvalidData here
```

`BufRead::lines()` validates UTF-8, so the gzip magic bytes fail on the first line with `ErrorKind::InvalidData` and the message `stream did not contain valid UTF-8`. The message names neither gzip nor the file, and suggests no remedy; the path only appears because of the `.with_context()` at the call site.

## Not a regression — but worth stating explicitly

I checked the C++ implementation on the `cpp` branch before filing, because "this used to work" would change the priority. It did not work there either, and it failed *worse*.

`generateGeneLevelEstimates` in `src/util/SalmonUtils.cpp` dispatches on exactly the same extension set:

```cpp
std::set<std::string> validGTFExtensions = {".gtf", ".gff", ".gff3",
                                            ".GTF", ".GFF", ".GFF3"};
auto extension = geneMapPath.extension();
```

For `annotation.gtf.gz` the extension is `.gz`, so C++ fell through to the two-column branch, `readTranscriptToGeneMap`, which does `while (ifile >> transcript >> gene)` over an `std::ifstream`. C++ streams do not validate UTF-8, so that loop consumed the compressed bytes as whitespace-delimited tokens and built a **garbage map with no error at all**.

So the honest framing is: gzip has never been supported, and 2.x turned a silent-garbage path into a loud failure. That is an improvement. What is left to fix is that the failure is late, the message is opaque, and the format is one users legitimately expect to work.

## A trap for whoever fixes this

The extension dispatch at [`genemap.rs:17-22`](https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-core/src/genemap.rs#L17-L22) sets `is_gtf` from the *last* extension:

```rust
let is_gtf = matches!(ext.as_str(), "gtf" | "gff" | "gff3");
```

For `annotation.gtf.gz` that is `gz`, so `is_gtf == false`. **Adding decompression alone is not enough** — the stream would decompress correctly and then be parsed by the two-column TSV branch, whose first two whitespace-separated fields on a GTF line are the sequence name and the source.

I confirmed this against the current code by handing it a GTF named `annotation.txt`:

```
parsed_as_tsv entries=1 content={"1": "havana"}
```

One bogus entry, no error, and `map.is_empty()` is false so the existing guard at `genemap.rs:51` does not catch it. That would convert this loud failure into the silent one in the companion issue. Any fix has to decide the GTF-vs-TSV question on the name with the compression suffix removed (or on content), not on the final extension.

## Suggested direction

- Detect gzip from the leading `1f 8b` magic bytes rather than the file name, so a mis-named `.gtf` that is actually compressed also works, and wrap in `flate2::read::MultiGzDecoder` (multi-member — concatenated members are legal and Ensembl files are not guaranteed single-member). `flate2` is already a workspace dependency used by `salmon-quant`, `salmon-align` and `salmon-model`, so this adds nothing to the dependency tree; `salmon-core` just has to opt in.
- Strip a trailing `.gz` before the GTF/TSV extension decision, per the trap above.
- Name the file in the error and, on an invalid-UTF-8 failure, say what the likely cause is.

Validating `--geneMap` before mapping starts would also be worth doing, but it is a separate change and I have left it out of the PR.

I have a PR ready for the reader itself.

BenjaminDEMAILLE:
Closes #1074.

`--geneMap Homo_sapiens.GRCh38.115.gtf.gz` failed with `stream did not contain valid UTF-8`, and only after mapping, EM and `quant.sf` had all finished. Gzip is how Ensembl and GENCODE ship annotations, so this is the invocation people reach for first, and it cost a full run to discover.

## Root cause

[`read_transcript_gene_map`](https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-core/src/genemap.rs#L16) in `crates/salmon-core/src/genemap.rs` opened the path as plain text:

```rust
let reader = io::BufReader::new(std::fs::File::open(path)?);
for line in reader.lines() {
    let line = line?;          // ← BufRead::lines() validates UTF-8; gzip magic fails here
```

It surfaces at the end of the run because gene-level aggregation is the last step: `write_gene_level` is called from `main.rs:1612` after `quant.sf` is written, and opens the gene map as its first action.

This is **not a regression.** I checked the `cpp` branch before writing any code, because "it used to work" would have changed the fix. `generateGeneLevelEstimates` in `src/util/SalmonUtils.cpp` dispatches on the same extension set (`.gtf/.gff/.gff3` and uppercase), so `.gtf.gz` fell through to the two-column branch, where `while (ifile >> transcript >> gene)` consumed compressed bytes as tokens without complaint. C++ built a garbage map silently. 2.x already turned that into a loud failure; this PR makes the format work instead.

## The part that is easy to get wrong

Adding decompression alone would have been a bug. `is_gtf` came from the *last* extension:

```rust
let is_gtf = matches!(ext.as_str(), "gtf" | "gff" | "gff3");
```

For `annotation.gtf.gz` that is `gz`, so the decompressed GTF would have been handed to the TSV branch, whose first two whitespace-separated fields on a GTF line are the sequence name and the source. Measured against the reader as it stands, by giving it a GTF named `annotation.txt`:

```
parsed_as_tsv entries=1 content={"1": "havana"}
```

One bogus entry. Non-empty, so the `map.is_empty()` guard does not fire; no transcript matches; `quant.genes.sf` comes out empty and the run exits 0. That is exactly the silent failure in #1075 — so the naive fix would have traded a loud bug for a quiet one. `has_gtf_extension` strips a trailing `.gz` before deciding, and `reads_gzipped_gtf` pins it: the assertion is on the parsed pairs, so a TSV misparse fails the test.

## What changed

- **Detection by magic bytes, not by name.** `BufRead::fill_buf` peeks the leading `1f 8b` without consuming, so a compressed annotation that kept a bare `.gtf` name also works.
- **`MultiGzDecoder`, not `GzDecoder`.** Concatenated gzip members are legal; a single-member decoder stops at the first member's end and returns a *partial* map with no error. Covered by `reads_multi_member_gzip`.
- **`.gz`-aware GTF/TSV detection**, per above.
- **Errors name the file and hint at the cause.** `File::open`, the sniff and the line loop are all annotated:

  ```
  reading gene map /path/Homo_sapiens.GRCh38.115.gtf.bz2: stream did not contain
  valid UTF-8 (expected plain text or gzip; annotations compressed with anything
  else, such as bzip2 or zstd, must be decompressed first)
  ```

  The underlying message is preserved rather than replaced, so a truncated gzip still reports as a gzip failure. The caller's `.with_context("reading gene map {path}")` in salmon-cli is dropped, since it would now print the path twice.
- **Docs**: `--geneMap` help text, the CLI reference table, and the quick-start bullet.

## Before / after

| Input | Before | After |
|---|---|---|
| `annotation.gtf` | 2 pairs | 2 pairs |
| `t2g.tsv` | 2 pairs | 2 pairs |
| `annotation.gtf.gz` | `Error: stream did not contain valid UTF-8`, after the full run | 2 pairs |
| `t2g.tsv.gz` | same error | 2 pairs |
| gzip content named `.gtf` | same error | 2 pairs |
| multi-member `.gtf.gz` | same error | 2 pairs (not truncated) |
| `.gtf.bz2` | `stream did not contain valid UTF-8` | same, plus file name and what was expected |
| missing file | bare errno | `reading gene map <path>: No such file or directory` |

`salmon-core` goes from 20 to 30 tests.

## Commits

Four, each self-contained, each with its test in the same commit:

1. `test(genemap): cover plain, gzip and TSV gene-map inputs` — the gzip case is `#[ignore]`d with a pointer to #1074 so the commit is green and the series bisectable. I checked it fails for the documented reason, not incidentally: `called Result::unwrap() on an Err value: Error { kind: InvalidData, message: "stream did not contain valid UTF-8" }`.
2. `feat(genemap): detect gzip by magic bytes and decompress transparently` — removes the `#[ignore]` in the same commit.
3. `fix(genemap): surface the file path and a hint in reader errors`
4. `docs: state that --geneMap accepts gzip-compressed GTF/GFF`

Verified per commit, not just at the tip — `cargo fmt --all --check`, `cargo clippy -p salmon-core -p salmon-cli --all-targets` and `cargo test` on each of the four:

```
4ca1355f test(genemap): cover plain, gzip and TSV  | fmt=OK clippy_errors=0 test_failures=0
abc46086 feat(genemap): detect gzip by magic bytes | fmt=OK clippy_errors=0 test_failures=0
63b26d71 fix(genemap): surface the file path       | fmt=OK clippy_errors=0 test_failures=0
648f6b70 docs: state that --geneMap accepts gzip   | fmt=OK clippy_errors=0 test_failures=0
```

`cargo test --workspace` is green at the tip. The pre-existing clippy warnings in salmon-model, salmon-align and salmon-map are untouched and also present on `develop`.

## Dependency

`flate2` moves into `salmon-core`'s `[dependencies]`. It is already a workspace dependency used by `salmon-quant`, `salmon-align` and `salmon-model`, so the dependency tree does not change — `salmon-core` only opts in.

## Deliberately not addressed

- **`--geneMap` is still not validated before mapping starts.** A bad path or an unreadable annotation still costs the whole run before it is reported. Worth fixing, but it is a change to the CLI's startup ordering rather than to the reader, so I left it out; happy to do it separately if you want it.
- **Other compression formats** (bzip2, zstd). The error now names them as things to decompress rather than supporting them; adding codecs seemed like scope creep for a bug report about the format Ensembl actually ships.
- **The empty-`quant.genes.sf` problem** is #1075 and is independent — different cause, different fix, no overlap with this branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2683,6 +2683,7 @@ dependencies = [
 name = "salmon-core"
 version = "2.4.1"
 dependencies = [
+ "flate2",
  "serde",
  "thiserror 2.0.19",
  "tracing",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2683,7 +2683,7 @@ dependencies = [
 name = "salmon-core"
 version = "2.4.1"
 dependencies = [
- "flate2",
+ "niffler",
  "serde",
  "thiserror 2.0.19",
  "tracing",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,9 @@ xxhash-rust = { version = "0.8", features = ["xxh3", "xxh64"] }
 # compression
 flate2 = "1"
 zstd = "0.13"
+# format-sniffing decompression (gzip/BGZF, bzip2, xz, zstd) for user-supplied
+# annotation files; already in the tree via paraseq and piscem-rs.
+niffler = "3"
 # logging
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -401,7 +401,8 @@ struct QuantArgs {
     #[arg(short = 'o', long = "output", required = true)]
     output: PathBuf,
     /// Transcript-to-gene map (GTF/GFF, or a 2-column TSV), optionally
-    /// gzip-compressed; also writes gene-level estimates to `quant.genes.sf`.
+    /// compressed (gzip, BGZF, bzip2, xz, zstd); also writes gene-level
+    /// estimates to `quant.genes.sf`.
     #[arg(short = 'g', long = "geneMap", value_name = "FILE")]
     gene_map: Option<PathBuf>,
     /// Worker threads (0 = all cores). salmon also runs one auxiliary thread for
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -400,8 +400,8 @@ struct QuantArgs {
     /// Output directory.
     #[arg(short = 'o', long = "output", required = true)]
     output: PathBuf,
-    /// Transcript-to-gene map (GTF/GFF, or a 2-column TSV); also writes
-    /// gene-level estimates to `quant.genes.sf`.
+    /// Transcript-to-gene map (GTF/GFF, or a 2-column TSV), optionally
+    /// gzip-compressed; also writes gene-level estimates to `quant.genes.sf`.
     #[arg(short = 'g', long = "geneMap", value_name = "FILE")]
     gene_map: Option<PathBuf>,
     /// Worker threads (0 = all cores). salmon also runs one auxiliary thread for
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -862,8 +862,9 @@ fn write_gene_level(
     tpm: &[f64],
     counts: &[f64],
 ) -> Result<()> {
-    let map = salmon_core::genemap::read_transcript_gene_map(gene_map)
-        .with_context(|| format!("reading gene map {}", gene_map.display()))?;
+    // `read_transcript_gene_map` names the file in every error it returns, so
+    // wrapping it here would print the path twice.
+    let map = salmon_core::genemap::read_transcript_gene_map(gene_map)?;
     let unmapped = salmon_core::genemap::write_gene_quant(
         &out_dir.join("quant.genes.sf"),
         names,
--- a/crates/salmon-core/Cargo.toml
+++ b/crates/salmon-core/Cargo.toml
@@ -12,8 +12,8 @@ description = "Shared core types for the Rust port of salmon (transcripts, libra
 thiserror = { workspace = true }
 serde = { workspace = true }
 tracing = { workspace = true }
-# transparent gzip decompression for gzipped --geneMap annotations
-flate2 = { workspace = true }
+# transparent decompression of compressed --geneMap annotations
+niffler = { workspace = true }
 
 [lints]
 workspace = true
--- a/crates/salmon-core/Cargo.toml
+++ b/crates/salmon-core/Cargo.toml
@@ -12,9 +12,7 @@ description = "Shared core types for the Rust port of salmon (transcripts, libra
 thiserror = { workspace = true }
 serde = { workspace = true }
 tracing = { workspace = true }
-
-[dev-dependencies]
-# builds the gzipped gene-map fixture in the genemap tests
+# transparent gzip decompression for gzipped --geneMap annotations
 flate2 = { workspace = true }
 
 [lints]
--- a/crates/salmon-core/Cargo.toml
+++ b/crates/salmon-core/Cargo.toml
@@ -13,5 +13,9 @@ thiserror = { workspace = true }
 serde = { workspace = true }
 tracing = { workspace = true }
 
+[dev-dependencies]
+# builds the gzipped gene-map fixture in the genemap tests
+flate2 = { workspace = true }
+
 [lints]
 workspace = true
--- a/crates/salmon-core/src/genemap.rs
+++ b/crates/salmon-core/src/genemap.rs
@@ -5,6 +5,7 @@
 //! and EffectiveLength are the TPM-weighted means of the transcript (effective)
 //! lengths, falling back to the unweighted mean when the gene's total TPM is 0.
 
+use flate2::read::MultiGzDecoder;
 use std::collections::{BTreeMap, HashMap};
 use std::io::{self, BufRead, Write};
 use std::path::Path;
@@ -13,14 +14,13 @@ use std::path::Path;
 /// `transcript_id` and `gene_id` attributes (GTF `key "value"` and GFF3
 /// `key=value` syntaxes); anything else is read as a TSV whose first two
 /// whitespace-separated columns are `transcript` and `gene`.
+///
+/// The file may be gzip-compressed — as Ensembl and GENCODE ship it. Compression
+/// is detected from the content, not the name, so a compressed file that kept a
+/// bare `.gtf` name is read correctly too.
 pub fn read_transcript_gene_map(path: &Path) -> io::Result<HashMap<String, String>> {
-    let ext = path
-        .extension()
-        .and_then(|e| e.to_str())
-        .unwrap_or("")
-        .to_ascii_lowercase();
-    let is_gtf = matches!(ext.as_str(), "gtf" | "gff" | "gff3");
-    let reader = io::BufReader::new(std::fs::File::open(path)?);
+    let is_gtf = has_gtf_extension(path);
+    let reader = open_maybe_gzip(path)?;
     let mut map = HashMap::new();
 
     for line in reader.lines() {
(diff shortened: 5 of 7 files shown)


rob-p:
Hi @BenjaminDEMAILLE,

  Thanks for the issue and the fix. I'd like to request a simplifying change.  Specifically, salmon already pulls in [`niffler`](https://docs.rs/niffler/latest/niffler/) as a dependency.  So, rather than rolling our own bespoke gzip magic sniffer and dispatching on the appropriate reader, let's just use `niffler` and it's [`get_reader`](https://docs.rs/niffler/latest/niffler/basic/fn.get_reader.html) functionality here?

--Rob

BenjaminDEMAILLE:
Done in a591aec4 — you were right, and it bought more than the deletion.

`niffler` was already in `Cargo.lock` (via `paraseq` and `piscem-rs`) with its default features, so `bzip2`, `liblzma` and `zstd` were being compiled either way. `salmon-core` moves off `flate2` and onto `niffler`; the dependency tree is unchanged. The test fixtures use `niffler::get_writer`, so no dev-dependency was needed for them either.

`get_reader`'s gzip path is `flate2::read::MultiGzDecoder`, so the multi-member guarantee the bespoke version was written for is kept rather than reimplemented — and that is also what makes a BGZF annotation work, since BGZF is a multi-member gzip.

Three things fell out of the switch that were worth a test each:

- **bzip2, xz and zstd now work**, so the error hint no longer tells people to decompress those first. `reads_every_supported_codec` covers all four end to end.
- **`has_gtf_extension` has to strip their suffixes too.** Missing that would route a decompressed `annotation.gtf.bz2` to the TSV branch, whose first two whitespace-separated fields on a GTF line are the sequence name and the source — the same non-empty, matches-nothing `{"1": "havana"}` map the `.gz` case was guarded against. Same trap, four more doors.
- **`niffler` reads five bytes to sniff and returns `FileTooShort` below that**, where the old sniff just saw a short buffer. A gene map that small is plain text by construction, so that case re-opens the file and reads it as text; `reads_a_file_shorter_than_the_sniff_window` pins it with a four-byte `t\tg\n`.

`salmon-core` goes from 30 to 32 tests. `cargo fmt --all --check`, `cargo clippy -p salmon-core -p salmon-cli --all-targets` (0 errors) and `cargo test --workspace` are green at the tip.

The "other compression formats" item under *Deliberately not addressed* in the description is obsolete — they are addressed now, for free. I have left the earlier commits alone so this one reads as the review diff; say the word if you would rather I squash it into commit 2 before merge.

BenjaminDEMAILLE:
One process note before this is merged: CI has not run on the niffler commit. The only runs on this branch are against `648f6b70`, the pre-review tip — `a591aec4` never got one, so the fork workflow is presumably waiting on your approval. Worth a click before merge, since the switch is the kind of change CI exists to catch.

That said, it should be uneventful, and I checked rather than assumed. On `develop` as it stands, `niffler` is not merely present in `Cargo.lock` — it is built, with default features, via `paraseq` → `piscem-rs`:

```
$ cargo tree -i niffler          # on develop, unmodified
niffler v3.0.0
└── paraseq v0.4.14
    └── piscem-rs v0.6.4
        ├── salmon-cli
        ├── salmon-index
        └── ...

$ cargo tree -e features -i niffler | grep 'niffler feature'
niffler feature "bgz" / "bz2" / "gz" / "lzma" / "zstd"  (all via "default")
```

So `bzip2-sys`, `liblzma-sys` and `zstd-sys` already compile on every target the CI matrix builds today; `salmon-core` naming the crate adds no new build surface, on any platform. My local verification is macOS arm64 only, which is exactly why the matrix run is still worth having rather than something to skip.

Nothing else outstanding from my side on this one.

PR #1077, opened 2026-08-06
fix(genemap): report and remedy a gene map that matches no quantified transcript

### LINKED ISSUE #1075: Empty quant.genes.sf with Ensembl cDNA + GTF: version-mismatched IDs match nothing, and the summary line reports success

BenjaminDEMAILLE:
Fixes #1075.

Unmatched transcripts stay dropped rather than reverting to C++'s transcript-as-its-own-gene fallback, and both migration documents now say so instead of claiming parity. Rationale in "A correction to the code" below; say the word if you want the fallback restored and I will send it as its own change.

## The problem

An index built from the Ensembl cDNA FASTA quantifies `ENST00000622028.1`. The Ensembl GTF says `transcript_id "ENST00000622028"` and puts the version in a separate `transcript_version` attribute. `write_gene_quant` joins them with a literal string compare, so the match rate is not low, it is **zero** — and `quant.genes.sf` comes out with only its header, exit code 0.

## salmon was not silent — it was contradictory

I assumed "no warning" when I filed #1075 and that was wrong; the code disproved it. `main.rs:878` does warn. The problem is the line after it. Running unpatched `develop` against a 15-transcript index whose FASTA is versioned and whose GTF is not:

```
warning: 15 transcript(s) had no entry in the gene map and were omitted from quant.genes.sf
wrote gene-level estimates for 15 genes to quant.genes.sf
```

```
$ wc -l out/quant.genes.sf
       1
```

"wrote gene-level estimates for 15 genes" — to a file with **zero** gene rows. `n_genes` was counted from the gene map, not from what was written:

```rust
let n_genes = map.values().collect::<HashSet<_>>().len();
println!("wrote gene-level estimates for {n_genes} genes to quant.genes.sf");
```

So the run's last word is a success message, on stdout, contradicting a warning on stderr. That is the actual defect: not silence, but a reassuring summary that is wrong.

## After

Same input, same branch, no flag — behaviour unchanged, diagnosis added:

```
warning: 15 of 15 transcript(s) had no entry in the gene map and were omitted from quant.genes.sf
warning: the gene map matched only 0 of 15 quantified transcripts (0.0%); quant.genes.sf has 0 gene row(s)
warning: 15 would match with --ignoreTxVersion, which compares identifiers without their
         trailing version suffix (e.g. NM_001168316.1 vs NM_018953). The Ensembl cDNA FASTA
         carries transcript versions in the identifier while the GTF puts them in a separate
         transcript_version attribute, so an index built from the FASTA does not join to the
         GTF as-is.
wrote gene-level estimates for 0 genes to quant.genes.sf
```

The third warning only fires when ignoring versions would genuinely match more, so it diagnoses rather than guesses. With the flag:

```
$ salmon quant … -g annotation.gtf --ignoreTxVersion
wrote gene-level estimates for 15 genes to quant.genes.sf

$ wc -l out/quant.genes.sf
      16
```

Mass is conserved through the join:

| | TPM | NumReads |
|---|---|---|
| `quant.sf` | 1000000.000 | 10000.000 |
| `quant.genes.sf` | 1000000.000 | 10000.000 |

And the flag is inert when identifiers already agree — same index, a GTF whose `transcript_id` carries the version, with and without `--ignoreTxVersion`:

```
--- identical output with and without the flag? ---
IDENTICAL
```

## Summary of changes

| | |
|---|---|
| `write_gene_quant` return | `usize` → `GeneQuantSummary { matched, unmapped, genes_written }` |
| CLI summary line | counts gene-map genes → counts rows written |
| low match rate | undetected → warns at ≤ `LOW_MATCH_RATE` (0.5) |
| cause | unstated → names the version mismatch when that is demonstrably it |
| remedy | none → `--ignoreTxVersion`, off by default |

**On the threshold.** 0.5, justified in the doc comment where it is defined. The failure mode is total — a version mismatch matches nothing — and a correctly paired annotation matches essentially everything, so the two regimes are far apart and the exact cut is not delicate. Half means the majority of what was quantified is missing from the output, which is worth flagging in any workflow, while a deliberately partial annotation (single-chromosome GTF, spike-in-only map) is nowhere near being mistaken for correct. It gates a warning only; nothing fails, nothing changes.

**On the default.** `--ignoreTxVersion` is off deliberately, and I would push back on making it automatic. The complaint in #1075 is that a bad join produced an empty file without saying so; stripping versions implicitly would replace a silent wrong answer with a silent right one and leave the next mismatch just as invisible. Version-aware annotations are also a legitimate setup. Semantics match tximport's flag of the same name, so the name should already mean the right thing to users. `strip_tx_version` only strips when the final dot-separated part is all digits, so GENCODE's `ENST00000456328.2_PAR_Y` survives.

## A correction to the code

The doc comment on `write_gene_quant` said unmatched transcripts are skipped "(salmon's behavior)". That is not what C++ salmon did — `aggregateEstimatesToGeneLevel` emitted each unmatched transcript as its own single-transcript gene and warned once per transcript:

```cpp
auto gn = tgm.geneName(er.target, foundGene);   // returns transcriptName when not found
if (!foundGene) { logger->warn("… returning transcript as it's own gene", er.target); }
geneExps[gn].push_back(std::move(er));          // emitted regardless
```

So on this input C++ 1.x wrote 515,982 rows and half a million warnings — useless as gene-level output, impossible to miss. 2.x drops them and reported success.

This branch keeps the dropping and documents it, rather than restoring the C++ fallback. Emitting every unmatched transcript as its own gene turns a failed join into a `quant.genes.sf` full of transcript IDs posing as genes — 515,982 plausible-looking rows that a downstream `tximport` or `DESeq2` will happily consume — which is harder to notice than an empty file, not easier. What C++ had going for it was the half-million warnings, and the warnings added in commit 2 carry that signal without the junk output. The doc comment now records the reasoning.

`MIGRATION.md` and `migrating/from-cpp.mdx` listed `-g/--geneMap` under *Unchanged*; commit 5 moves it to *Behavior differences to be aware of* in both, and lists `--ignoreTxVersion` under *New in 2.0*.

## Commits

1. `test(genemap): cover a gene map that matches no quantified transcript` — characterisation test; asserts the *current* outcome, passes as-is, so the following commits move against a documented baseline.
2. `fix(genemap): warn when the gene map matches few or no transcripts` — no change to matching.
3. `feat(genemap): add --ignoreTxVersion to strip trailing version suffixes` — opt-in.
4. `docs: document --ignoreTxVersion and the Ensembl cDNA + GTF pitfall`
5. `docs: correct the C++ parity claim for -g/--geneMap`

Verified per commit, not just at the tip — `cargo fmt --all --check`, `cargo clippy -p salmon-core -p salmon-cli --all-targets`, `cargo test`:

```
14a56b7e test(genemap): cover a gene map that matches no q | fmt=OK clippy_err=0 test_fail=0
bf185c84 fix(genemap): warn when the gene map matches few  | fmt=OK clippy_err=0 test_fail=0
01074258 feat(genemap): add --ignoreTxVersion to strip tra | fmt=OK clippy_err=0 test_fail=0
2d7d6a4d docs: document --ignoreTxVersion and the Ensemb | fmt=OK clippy_err=0 test_fail=0
79fb7cd1 docs: correct the C++ parity claim for -g/--gen | fmt=OK clippy_err=0 test_fail=0
```

`cargo test --workspace` is green at the tip. salmon-core goes from 20 to 27 tests.

## Note on the API and the plumbing

`write_gene_quant` now returns a struct and takes an `ignore_tx_version` flag — a breaking change to a `pub fn` in salmon-core, with salmon-cli the only caller in-tree. The gene-map path and the flag travel together as a small `GeneMapOpts` rather than as a bare `bool` threaded through `run_deterministic` and the other two quantification entry points; that keeps their arity unchanged and keeps the option attached to the path it qualifies. Happy to do it the other way if you prefer.

## Deliberately not addressed

- **Other identifier mismatches** (GENCODE `|`-delimited FASTA headers, RefSeq/Ensembl crossovers). The low-match-rate warning will now flag them, but only the version case gets a named cause and a flag, because it is the one with a settled convention to copy from tximport.
- **The gzip issue (#1074)** is independent and is #1076; no overlap with this branch beyond both touching `genemap.rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2684,6 +2684,7 @@ name = "salmon-core"
 version = "2.4.1"
 dependencies = [
  "serde",
+ "serde_json",
  "thiserror 2.0.19",
  "tracing",
 ]
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -81,7 +81,8 @@ These still parse so existing scripts run; 2.0 logs a warning and ignores them
   level and `tximport` can apply its own transcript-to-gene policy to the whole
   file. What changed is the reporting: C++ warned once *per unmatched transcript*
   (half a million lines on a failed join), while 2.x warns once with the count
-  and writes the names to `aux_info/genemap_unmatched_txps.txt`, one per line.
+  and writes the names to `aux_info/genemap_unmatched_txps.json`, under an
+  `unmatched_transcripts` key.
   That file is removed when nothing is unmatched, so its presence always refers
   to the current run. 2.x additionally warns when the match rate is at or below
   50%, says how many transcripts would match without the version suffix, and
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -65,10 +65,24 @@ These still parse so existing scripts run; 2.0 logs a warning and ignores them
   alignment more closely.
 - `--allowDovetail` — now honored in `--sketch` as well (admits dovetailed
   short-insert fragments).
+- `--ignoreTxVersion` — compare `-g/--geneMap` identifiers without their trailing
+  `.N` version suffix, as tximport's option of the same name does. Off by
+  default. Needed for an Ensembl cDNA index against an Ensembl GTF, where the
+  FASTA carries the version in the identifier and the GTF puts it in a separate
+  `transcript_version` attribute.
 
 ## Behavior differences to be aware of
 
 - **Sketch orphan rule** defaults to the relaxed policy (see `--sketchStrictOrphans`).
+- **`-g/--geneMap`: unmatched transcripts are dropped, not emitted.** C++
+  salmon's `aggregateEstimatesToGeneLevel` gave a transcript with no gene-map
+  entry a gene of its own (named after the transcript) and warned once per
+  transcript. 2.x omits it from `quant.genes.sf` instead, and reports the total
+  once. If every identifier fails to match — the usual cause is the Ensembl
+  cDNA + GTF version mismatch above — C++ wrote one row per transcript where 2.x
+  writes a header and no rows. salmon now warns when the match rate is at or
+  below 50%, says how many transcripts would match without the version suffix,
+  and names `--ignoreTxVersion`.
 - **Selective-alignment chain pruning:** 2.0 currently defaults
   `--orphanChainSubThresh` and `--postMergeChainSubThresh` to `0.0` (off) — it
   aligns every candidate, which is marginally more sensitive than C++ (which uses
@@ -82,5 +96,5 @@ Index/quant basics, `quant.sf`, `cmd_info.json`, `lib_format_counts.json`,
 `aux_info/meta_info.json`, `--libType/-l`, `--threads/-p`, `-k/--kmerLen`,
 `-m/--minimizerLen`, `-n/--no-clip` (poly-A clipping, on by default),
 `--seqBias`, `--gcBias`, `--posBias`, `--numBootstraps`, `--numGibbsSamples`,
-`--useEM`, `--meta` (metagenomic preset), `--dumpEq`, `-g/--geneMap`, decoys, and
+`--useEM`, `--meta` (metagenomic preset), `--dumpEq`, decoys, and
 `salmon quantmerge`.
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -74,15 +74,20 @@ These still parse so existing scripts run; 2.0 logs a warning and ignores them
 ## Behavior differences to be aware of
 
 - **Sketch orphan rule** defaults to the relaxed policy (see `--sketchStrictOrphans`).
-- **`-g/--geneMap`: unmatched transcripts are dropped, not emitted.** C++
-  salmon's `aggregateEstimatesToGeneLevel` gave a transcript with no gene-map
-  entry a gene of its own (named after the transcript) and warned once per
-  transcript. 2.x omits it from `quant.genes.sf` instead, and reports the total
-  once. If every identifier fails to match — the usual cause is the Ensembl
-  cDNA + GTF version mismatch above — C++ wrote one row per transcript where 2.x
-  writes a header and no rows. salmon now warns when the match rate is at or
-  below 50%, says how many transcripts would match without the version suffix,
-  and names `--ignoreTxVersion`.
+- **`-g/--geneMap`: unmatched transcripts are reported differently, not
+  aggregated differently.** As in C++ salmon's `aggregateEstimatesToGeneLevel`, a
+  transcript with no gene-map entry is emitted as its own single-transcript gene,
+  named after the transcript, so nothing quantified is dropped on the way to gene
+  level and `tximport` can apply its own transcript-to-gene policy to the whole
+  file. What changed is the reporting: C++ warned once *per unmatched transcript*
+  (half a million lines on a failed join), while 2.x warns once with the count
+  and writes the names to `aux_info/genemap_unmatched_txps.txt`, one per line.
+  That file is removed when nothing is unmatched, so its presence always refers
+  to the current run. 2.x additionally warns when the match rate is at or below
+  50%, says how many transcripts would match without the version suffix, and
+  names `--ignoreTxVersion`. Note that a `quant.genes.sf` whose rows are mostly
+  stand-ins is transcript-level output under a gene-level file name — the run's
+  closing line says so explicitly when any row is a stand-in.
 - **Selective-alignment chain pruning:** 2.0 currently defaults
   `--orphanChainSubThresh` and `--postMergeChainSubThresh` to `0.0` (off) — it
   aligns every candidate, which is marginally more sensitive than C++ (which uses
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -864,7 +864,7 @@ fn write_gene_level(
 ) -> Result<()> {
     let map = salmon_core::genemap::read_transcript_gene_map(gene_map)
         .with_context(|| format!("reading gene map {}", gene_map.display()))?;
-    let unmapped = salmon_core::genemap::write_gene_quant(
+    let summary = salmon_core::genemap::write_gene_quant(
         &out_dir.join("quant.genes.sf"),
         names,
         lengths,
@@ -874,13 +874,50 @@ fn write_gene_level(
         &map,
     )
     .context("writing quant.genes.sf")?;
-    let n_genes = map.values().collect::<std::collections::HashSet<_>>().len();
-    if unmapped > 0 {
+
+    if summary.unmapped > 0 {
         eprintln!(
-            "warning: {unmapped} transcript(s) had no entry in the gene map and were omitted from quant.genes.sf"
+            "warning: {} of {} transcript(s) had no entry in the gene map and were omitted \
+             from quant.genes.sf",
+            summary.unmapped,
+            summary.total()
         );
     }
-    println!("wrote gene-level estimates for {n_genes} genes to quant.genes.sf");
+    // A near-total mismatch means the identifiers do not line up, not that the
+    // annotation is merely incomplete — say so, rather than leaving the user to
+    // infer it from an empty file.
+    if summary.match_rate_is_low() {
+        eprintln!(
+            "warning: the gene map matched only {} of {} quantified transcripts ({:.1}%); \
+             quant.genes.sf has {} gene row(s)",
+            summary.matched,
+            summary.total(),
+            summary.match_rate() * 100.0,
+            summary.genes_written
+        );
+        let would_match = salmon_core::genemap::matches_ignoring_tx_version(names, &map);
+        if would_match > summary.matched {
+            eprintln!(
+                "warning: {would_match} would match if the trailing version suffix were ignored \
+                 (e.g. {} vs {}). The Ensembl cDNA FASTA carries transcript versions in the \
+                 identifier while the GTF puts them in a separate transcript_version attribute, \
+                 so an index built from the FASTA does not join to the GTF as-is.",
+                names
+                    .iter()
+                    .find(|n| !map.contains_key(*n))
+                    .map(String::as_str)
+                    .unwrap_or("ENST00000456328.2"),
+                map.keys()
+                    .next()
+                    .map(String::as_str)
+                    .unwrap_or("ENST00000456328"),
+            );
+        }
+    }
+    println!(
+        "wrote gene-level estimates for {} genes to quant.genes.sf",
+        summary.genes_written
+    );
     Ok(())
 }
 
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -871,9 +871,9 @@ struct GeneMapOpts {
 }
 
 /// Name of the file under `aux_info/` listing transcripts the gene map did not
-/// cover, one per line. It sits beside `unmapped_names.txt` — which is about
-/// *fragments* that did not map — so the name says transcripts explicitly.
-const UNMATCHED_TXP_FILE: &str = "genemap_unmatched_txps.txt";
+/// cover. It sits beside `unmapped_names.txt` — which is about *fragments* that
+/// did not map — so the name says transcripts explicitly.
+const UNMATCHED_TXP_FILE: &str = "genemap_unmatched_txps.json";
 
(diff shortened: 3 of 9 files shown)


rob-p:
Thanks @BenjaminDEMAILLE, I have some work cooking on the new version which will include the rapidgzip stuff (but with a truly dynamic scheduler) and some other goodies. So it would be nice to include this and your other PR in the upcoming release.

I'm actually torn a bit on the new behavior though. I kind of prefer falling back to the C++ version's behavior with a _very loud_ error message (though not half a million separate error messages).  Downstream tools like `tximport` have their own policies and mechanisms for handling transcript-to-gene mapping (and, honestly, the gene level quant feature predates `tximport` which is now the recommended way to aggregate transcript level abundances to the gene level).  Thus, I think I'd prefer reverting to the prior behavior of treating an unmapped transcript as it's own gene, but then loudly complaining and reporting the number of unmatched transcripts, and possibly including such a list in a non-transient output file.

Let me know what you think about that design decision, and if you have any other questions about it.

BenjaminDEMAILLE:
You've convinced me, and the reason is the part of your comment I had not weighed: gene-level quant predates `tximport`, and `tximport` has its own transcript-to-gene policy. Dropping rows takes that decision away from the tool that should be making it, and drops abundance on the floor to do it. My "515,982 plausible-looking rows" argument was about a reader who never sees the warnings — but the fix for that is to make the warnings unmissable, which is exactly what you proposed, not to delete data.

Implemented in 6a5e82dd + b8ab1b0f, all three parts.

**1. Fallback restored.** An unmatched transcript is written as its own single-transcript gene, keyed by its own name, as `aggregateEstimatesToGeneLevel` did. Mass is now conserved through the join whether or not the join succeeds.

**2. Loud, once.** One warning with the count, plus the low-match-rate warning reworked: with the fallback in place the file looks full either way, so the row split is the only thing separating a gene-level result from a transcript list, and it is now stated rather than left to be inferred.

**3. The list, in a non-transient file.** `aux_info/genemap_unmatched_txps.txt`, one identifier per line in `quant.sf` order, no header — greppable, and it outlives the terminal. It is *removed* when nothing is unmatched, so a stale list from an earlier run into the same output directory can't be read as a current failure.

Measured, 15-transcript index whose FASTA is versioned and whose GTF is not:

```
warning: 15 of 15 transcript(s) had no entry in the gene map. Each was written to
         quant.genes.sf as its own single-transcript gene, so those rows are named by
         transcript, not by gene. Their names are listed in
         out/aux_info/genemap_unmatched_txps.txt
warning: the gene map matched only 0 of 15 quantified transcripts (0.0%), so of the 15
         rows in quant.genes.sf, 15 are unmatched transcripts standing in as their own
         gene. Treat this output as transcript-level until the join is fixed.
warning: 15 would match with --ignoreTxVersion, ...
wrote 15 rows to quant.genes.sf: 0 gene(s), plus 15 unmatched transcript(s) as their own gene
```

| | TPM | NumReads |
|---|---|---|
| `quant.sf` | 1000000.000 | 10000.000 |
| `quant.genes.sf` | 1000000.000 | 10000.000 |

With `--ignoreTxVersion`, or against a GTF whose `transcript_id` carries the version: 5 genes, no warnings, and the list file removed.

**One place I pushed past the brief, because it is the bug this PR is about.** The closing line used to read `wrote gene-level estimates for 15 genes` — the #1075 defect, a success summary contradicting the warnings above it. Restoring the fallback brings it straight back in a new form: 15 rows, none of them a gene. So the count is called "genes" only when every row is one, and otherwise the split is spelled out, as above. Happy to revert to the flat wording if you'd rather.

**Two properties worth knowing about.** The fallback keys on the transcript name, so an unmatched transcript named exactly like a real gene merges into that gene's row — C++ had this too; there's a test pinning it as known rather than surprising. And `GeneQuantSummary::unmapped` is renamed `unmatched`, since they're no longer omitted from anything and `unmapped` reads as being about fragments, which `aux_info/unmapped_names.txt` already is.

`salmon-core` goes from 27 to 29 tests. `cargo fmt --all --check`, `cargo clippy -p salmon-core -p salmon-cli --all-targets` (0 errors) and `cargo test --workspace` green at the tip. Both migration documents now say the aggregation matches C++ and that the difference is the reporting.

---

**A question your reasoning raises, separate from this PR.** If `tximport` is the recommended way to aggregate to gene level, and gene-level quant predates it, is `-g/--geneMap` still worth carrying? Everything it does downstream of `quant.sf` is a join and a group-by that `tximport` does better and with policy the user controls, and this PR plus #1076 is three rounds of work on the failure modes of a feature whose whole job is to be a worse version of something else. Not proposing to touch it here — plenty of pipelines use it, and it's your call, not mine. But if you'd want it marked as legacy in the docs, or eventually deprecated with a pointer to `tximport`, I'm happy to open an issue and do that work. If you'd rather keep it first-class, that's an equally clear answer and I'll stop asking.

Good to hear about the rapidgzip work — glad these fit the release.

rob-p:
Thanks! I have one more small change to request and then I think this is ready to merge. Can we make the `aux_info/genemap_unmatched_txps.txt` into a `json` file rather than a `txt` file? I'm trying, where possible, to move toward json for text-based, non-primary artifacts in these tools to ease downstream parsing.  This could be implemented as either just a list of identifiers or a single element map with a distinctive key name.

On the broader picture about dropping `--gene-map`, I think the answer may, in fact, be yet, but maybe not yet. Last time I tried to remove it, some users asked me not to. It is duplicative with `tximport` (and now `pytximport`), but some users use it and want it.

I say, let's get it fixed and in for a couple of releases, and then we can consider potentially deprecating and removing it if there's not much pushback from the community.

BenjaminDEMAILLE:
Done in fd84a955 — `aux_info/genemap_unmatched_txps.json`:

```json
{
  "unmatched_transcripts": [
    "ENST00000622028.1",
    "ENST00000456328.2"
  ]
}
```

I took the one-key object of the two shapes you offered. Both parse as easily, but an object can gain fields later — a count, the gene-map path, whether `--ignoreTxVersion` was on — without breaking a consumer written against it today, and a parsed object still says what it is once it has been separated from its filename. The key is named for the list rather than for the file it sits in, for the same reason. Say the word if you'd rather have the bare array.

`to_string_pretty` puts one identifier per line, so it stays as greppable as the text it replaces; `jq -r '.unmatched_transcripts[]'` is the intended reader. The round-trip is asserted through a JSON parse rather than on the raw text, so the test pins the document a consumer sees and not the formatting around it, plus that the object has exactly one key.

`salmon-core` picks up `serde_json`, already a workspace dependency and already used by `salmon-quant` for `aux_info/meta_info.json`, so the tree is unchanged.

Verified end to end on the same 15-transcript reproduction: 15 identifiers written, `python -m json.tool` and `jq -r '.unmatched_transcripts | length'` both read it, and a clean rerun into the same output directory still removes it. `cargo fmt --all --check`, `cargo clippy -p salmon-core -p salmon-cli --all-targets` (0 errors) and `cargo test --workspace` are green at the tip; `salmon-core` stays at 29 tests.

On `--geneMap`: that's a clear answer, thanks — fix it, ship it, revisit after a couple of releases with the community's reaction as the input rather than my inference from the code. I'll leave it alone and won't raise it again; if you do want the deprecation work later, ping me on this thread.

rob-p:
Brilliant; thanks! This will make it into the upcoming release.

Issue #1092, opened 2026-08-07
Tracking: performance, memory and correctness audit of the Rust workspace (v2.4.1)


BenjaminDEMAILLE:
Tracking issue for a performance, memory and correctness audit of the Rust workspace (`crates/salmon-*`) at `v2.4.1`.

The audit read the hot paths end to end: mapping (`salmon-map`), per-fragment recording (`salmon-quant/processor.rs`), equivalence-class construction (`salmon-eqclass`), inference (`salmon-infer`), the RAD read/write paths (`salmon-align/rad.rs`, `salmon-rad`), index loading (`salmon-index`) and the bias models (`salmon-model`). `cargo clippy --workspace --all-targets` is clean apart from two cosmetic lints, so everything below is structural rather than lint-level.

Findings are split into sub-issues, each self-contained with file and line references, an explanation of why the current shape costs what it costs, a proposed fix and acceptance criteria.

## Themes

**Per-fragment allocation.** The `*_into` buffer-reuse pattern is applied at the top of the mapping call chain and then stops. Below it, the per-fragment path still allocates 5 to 7 times in `record()`, a hash map per fragment in `finalize_mappings_counted_into`, and in the RAD path a `BTreeMap` plus two `Vec`s per transcript per fragment. At realistic fragment counts these are billions of malloc/free pairs.

**Shared-atomic contention.** Up to seven atomic read-modify-writes per fragment on counters every worker shares, plus two `RwLock` acquisitions and two `Arc` clones per fragment for a fragment-length snapshot that provably cannot change within a batch. These do not show up at low thread counts and get steadily worse as `-p` grows.

**Avoidable copies of the largest structures.** The equivalence-class builder deep-clones its entire map in a function that already owns it. The packed CSR inference layout is rebuilt from scratch two to four times per run when only one of its arrays changes between uses. The reference sequence is read into anonymous memory rather than mapped.

**Correctness.** Two genuine bugs: CSR offsets truncate silently past 2^32 incidences (wrong results, no diagnostic), and the auxiliary-model burn-in gate is both non-deterministic and inconsistent between two code paths. A third suspected bug (#1081) was investigated and withdrawn: `compat[0]` is safe because every surviving mapping of a fragment shares one status. Only its misleading comment is being fixed.

## Sub-issues

Ordered by recommended attack order: effort ascending within impact descending. The first four are a few lines each.

| # | Issue | Kind | Effort |
|---|---|---|---|
[...]

PR #1094, opened 2026-08-07
perf(eqclass,infer): stop copying the largest structures needlessly

### LINKED ISSUE #1079: perf(eqclass): EquivalenceClassBuilder::finish clones the whole map instead of moving it
Part of #1092.

## Problem

`EquivalenceClassBuilder::finish` consumes `self` by value, then **clones every key and value out of the map anyway**:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-eqclass/src/lib.rs#L431-L438

```rust
pub fn finish(self) -> CollapsedEqClasses {
    let mut classes: Vec<(TranscriptGroup, TGValue)> = Vec::with_capacity(self.map.len());
    self.map
        .iter()
        .for_each(|e| classes.push((e.key().clone(), e.value().clone())));
```

Every clone is deep:

- `TranscriptGroup` owns `txps: Vec<u32>` and `bins: Vec<u32>`
- `TGValue` owns `acc: Vec<u128>` (16 bytes per transcript per class, still live at this point)

So at the moment of `finish` the process holds **two full copies** of the largest structure in the run, and pays a heap allocation plus a memcpy per class per vector. On a human-scale run with millions of equivalence classes that is a multi-GB transient spike and tens of millions of allocations, for no reason: the original is dropped immediately afterwards.

## Fix

`dashmap::DashMap` implements `IntoIterator`. Since `self` is already owned, the entries can simply be moved:

```rust
let mut classes: Vec<(TranscriptGroup, TGValue)> = self.map.into_iter().collect();
```

This removes both the peak doubling and the copying. The subsequent `sort_by` / `materialize` / `total_count` logic is unchanged.

## Acceptance criteria

- [ ] `finish` moves entries out of the map rather than cloning them
- [ ] Peak RSS during the eq-class finalize step measurably drops (report before/after on the sample dataset)
- [ ] `crates/salmon-eqclass` unit tests pass unchanged
- [ ] `quant.sf` output is byte-identical before/after

### LINKED ISSUE #1080: perf(infer): PackedEqClasses::from_collapsed grows three large arrays without reserving
Part of #1092.

## Problem

`PackedEqClasses::from_collapsed` builds three large flat arrays starting from empty, with no capacity reserved:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-infer/src/packed.rs#L69-L91

```rust
let mut labels = Vec::new();
let mut starts = Vec::with_capacity(n + 1);
let mut combined = Vec::new();
let mut weights = Vec::new();
```

`labels`, `combined` and `weights` grow to the total number of `(class, transcript)` incidences, which on a human-scale run is tens of millions of elements. Growing from zero means:

- ~log2(N) reallocations, each copying the entire accumulated buffer
- a transient peak of roughly 1.5x the final size at the last doubling (old buffer plus new buffer both live during the copy)

`starts` already reserves correctly, which shows the intent; the other three were just missed.

## Fix

The exact size is cheap to compute from the input:

```rust
let total_labels: usize = eq
    .classes
    .iter()
    .filter(|(g, _)| g.valid)
    .map(|(g, _)| g.txps.len())
    .sum();
labels.reserve_exact(total_labels);
combined.reserve_exact(total_labels);
weights.reserve_exact(total_labels);
```

One extra pass over the class list (cheap, it only touches `Vec::len`) in exchange for zero reallocation and no transient peak.

## Acceptance criteria

- [ ] All four flat arrays are pre-sized before the fill loop
- [ ] No behavioural change: `PackedEqClasses` contents identical
- [ ] `crates/salmon-infer` tests pass unchanged

### LINKED ISSUE #1088: bug(infer): equivalence-class CSR offsets truncate silently past 2^32 incidences
Part of #1092.

**Latent correctness bug: silent integer truncation, no panic, no warning.**

## Problem

`PackedEqClasses` stores its CSR offsets as `u32`:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-infer/src/packed.rs#L52

```rust
pub starts: Vec<u32>,
```

and fills them with an unchecked cast:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-infer/src/packed.rs#L90

```rust
starts.push(labels.len() as u32);
```

Past 2^32 cumulative `(class, transcript)` incidences, `as u32` wraps silently. `class(i)` then computes:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-infer/src/packed.rs#L113-L118

```rust
let s = self.starts[i] as usize;
let e = self.starts[i + 1] as usize;
(&self.labels[s..e], &self.combined[s..e])
```

With wrapped offsets this either panics on a reversed range (best case) or silently returns the wrong transcripts and weights for a class, which the EM then happily optimises. Wrong abundances, no diagnostic.

## Reachability

4.29 billion CSR entries is about 17 GB of `labels` alone, so this is not a small-index concern. It becomes reachable on a heavily multi-mapping library against a large decoy-aware index, which is exactly the configuration where someone would already be running on a high-memory node and would have no reason to suspect arithmetic.

## Fix

Minimum: assert before the cast, so the failure is loud and immediate.

```rust
assert!(
    labels.len() <= u32::MAX as usize,
    "equivalence-class CSR exceeds u32 offsets ({} incidences); \
     rebuild with a smaller index or file an issue",
    labels.len()
);
```

Better: widen `starts` to `u64`. The array is `num_classes + 1` entries, so doubling it is negligible next to `labels`/`combined`/`weights`, and it removes the ceiling rather than reporting it.

## Acceptance criteria

- [ ] Overflow is impossible or loudly detected; it cannot silently produce wrong slices
- [ ] If widened to `u64`, memory impact confirmed negligible relative to the flat arrays
- [ ] A unit test covers the boundary behaviour (via a synthetic small-limit build or a direct check on the guard)

BenjaminDEMAILLE:
Part of #1092 (performance / memory audit). Two independent allocation fixes on the equivalence-class path, no behavioural change.

### `EquivalenceClassBuilder::finish` cloned a map it already owns

`finish` takes `self` by value, so the concurrent map is dropped either way. It nonetheless cloned every key and value out of it, and those clones are deep: `TranscriptGroup` owns two `Vec<u32>`, and `TGValue` owns the `Vec<u128>` fixed-point accumulator, which is still live at that point.

The effect was two full copies of the largest structure in the run held simultaneously, plus a heap allocation and a memcpy per class per vector. `DashMap` implements `IntoIterator`, so the entries can simply be moved. Fixes #1079.

### `PackedEqClasses::from_collapsed` did not reserve

`labels`, `combined` and `weights` grew from empty to tens of millions of elements: a reallocation and full copy at every doubling, plus a transient peak of roughly 1.5x the final size while both buffers are live. `starts` was already reserved correctly, which shows the intent; the other three were missed. The exact total is one cheap pass over the class list (it only reads `Vec::len`). Fixes #1080.

### CSR offsets could truncate silently

While in `from_collapsed`: the offsets are `u32` and `starts.push(labels.len() as u32)` wrapped silently past `u32::MAX` incidences. After that `class()` hands the optimizer slices belonging to *other* classes — wrong abundances, no panic, no warning, no diagnostic of any kind.

4.29 billion incidences is ~17 GB of `labels`, so this is not a small-index concern; it is reachable on a heavily multi-mapping library against a large decoy-aware index, which is exactly where nobody would think to suspect arithmetic. This adds an assertion so the limit fails loudly instead of producing plausible nonsense. Widening `starts` to `u64` would remove the ceiling rather than report it, and is a reasonable follow-up if you would prefer that. Fixes #1088.

### Verification

- `cargo test --workspace` passes; `cargo clippy --workspace --all-targets` adds no warnings; `cargo fmt` clean.
- Built against `develop` and compared to a `develop` baseline binary on `sample_data`: `quant.sf` byte-identical for plain, `--seqBias --gcBias --posBias`, and `--softclip` runs; `meta_info.json` (modulo timing) and `lib_format_counts.json` identical.

Caveat worth stating: `sample_data` has 15 transcripts, so the parity check confirms the code paths agree, not that the memory behaviour holds at transcriptome scale. The memory claims here are structural (fewer copies of the same data), not measured. I have not benchmarked peak RSS on a human-scale index.

### DIFF
--- a/crates/salmon-eqclass/src/lib.rs
+++ b/crates/salmon-eqclass/src/lib.rs
@@ -429,12 +429,12 @@ impl EquivalenceClassBuilder {
     /// class *indices* — which the EM uses everywhere — are identical on every
     /// run, no matter what order the threads happened to insert them in.
     pub fn finish(self) -> CollapsedEqClasses {
-        let mut classes: Vec<(TranscriptGroup, TGValue)> = Vec::with_capacity(self.map.len());
-        // Drain the concurrent map into a plain vector. Iteration order of a
-        // hash map is arbitrary, hence the sort on the next line.
-        self.map
-            .iter()
-            .for_each(|e| classes.push((e.key().clone(), e.value().clone())));
+        // Move the entries out of the concurrent map rather than cloning them.
+        // `self` is owned here, so the map is dropped either way; cloning would
+        // deep-copy every `txps`/`bins`/`acc` vector and hold two full copies of
+        // the largest structure in the run at once. Iteration order of a hash map
+        // is arbitrary, hence the sort on the next line.
+        let mut classes: Vec<(TranscriptGroup, TGValue)> = self.map.into_iter().collect();
         classes.sort_by(|a, b| a.0.txps.cmp(&b.0.txps));
         // Materialize f64 weights from each class's order-independent fixed-point
         // accumulator (the sort above already makes the class *order* stable).
--- a/crates/salmon-eqclass/src/lib.rs
+++ b/crates/salmon-eqclass/src/lib.rs
@@ -432,9 +432,20 @@ impl EquivalenceClassBuilder {
         // Move the entries out of the concurrent map rather than cloning them.
         // `self` is owned here, so the map is dropped either way; cloning would
         // deep-copy every `txps`/`bins`/`acc` vector and hold two full copies of
-        // the largest structure in the run at once. Iteration order of a hash map
-        // is arbitrary, hence the sort on the next line.
-        let mut classes: Vec<(TranscriptGroup, TGValue)> = self.map.into_iter().collect();
+        // the largest structure in the run at once.
+        //
+        // Reserve first: `dashmap`'s owning iterator implements neither
+        // `size_hint` nor `ExactSizeIterator`, so `collect` would report `(0,
+        // None)` and grow this vector from empty by doubling — a realloc and a
+        // memcpy of every element at each step, plus a transient ~1.5x peak at
+        // the last one. At 144 bytes per entry that is the same cost this change
+        // removes from `PackedEqClasses::from_collapsed`. `len` has to be read
+        // before `into_iter` consumes the map.
+        //
+        // Iteration order of a hash map is arbitrary, hence the sort below.
+        let num_classes = self.map.len();
+        let mut classes: Vec<(TranscriptGroup, TGValue)> = Vec::with_capacity(num_classes);
+        classes.extend(self.map);
         classes.sort_by(|a, b| a.0.txps.cmp(&b.0.txps));
         // Materialize f64 weights from each class's order-independent fixed-point
         // accumulator (the sort above already makes the class *order* stable).
--- a/crates/salmon-infer/src/packed.rs
+++ b/crates/salmon-infer/src/packed.rs
@@ -68,10 +68,30 @@ impl PackedEqClasses {
     /// ([`update_eff_lengths`](salmon_eqclass::CollapsedEqClasses::update_eff_lengths)).
     pub fn from_collapsed(eq: &CollapsedEqClasses, num_txps: usize) -> Self {
         let n = eq.classes.len();
-        let mut labels = Vec::new();
+        // Size the flat arrays up front. Growing them from empty costs a
+        // reallocation plus a full copy at every doubling, and leaves a transient
+        // peak of ~1.5x the final size while both buffers are live. The exact
+        // total is one cheap pass over the class list (it only reads `Vec::len`).
+        let total_labels: usize = eq
+            .classes
+            .iter()
+            .filter(|(g, _)| g.valid)
+            .map(|(g, _)| g.txps.len())
+            .sum();
+        // The CSR offsets below are `u32`. Past `u32::MAX` incidences the `as u32`
+        // cast would wrap silently and `class()` would hand the optimizer slices
+        // belonging to other classes — wrong abundances with no diagnostic. Fail
+        // loudly instead.
+        assert!(
+            total_labels <= u32::MAX as usize,
+            "equivalence-class CSR needs {total_labels} incidences, which exceeds \
+             the u32 offset limit ({}); please report this input",
+            u32::MAX
+        );
+        let mut labels = Vec::with_capacity(total_labels);
         let mut starts = Vec::with_capacity(n + 1);
-        let mut combined = Vec::new();
-        let mut weights = Vec::new();
+        let mut combined = Vec::with_capacity(total_labels);
+        let mut weights = Vec::with_capacity(total_labels);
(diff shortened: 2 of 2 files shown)


rob-p:
Reviewed and taking this — all three items match the assessment we made against #1092. I pushed one fix onto the branch rather than sending you round again (a4938bd).

## The fix: `finish` lost its reservation

The move to `into_iter` is the right change and removes the dominant cost, but it also dropped the `Vec::with_capacity(self.map.len())` that the cloning version had.

`dashmap`'s `OwningIter` implements neither `size_hint` nor `ExactSizeIterator`, so `collect` sees `(0, None)` and grows the vector from empty by doubling — a realloc plus a memcpy of everything accumulated so far at each step, and a transient ~1.5x peak while both buffers are live. I measured `(TranscriptGroup, TGValue)` at **144 bytes**, so at ten million classes that is ~1.4 GB regrown roughly two dozen times.

That is the same cost this PR removes from `from_collapsed` a few lines away, so it should not reappear here. Reserving first keeps both properties — entries are moved, destination sized once:

```rust
let num_classes = self.map.len();      // before into_iter consumes the map
let mut classes: Vec<(TranscriptGroup, TGValue)> = Vec::with_capacity(num_classes);
classes.extend(self.map);
```

## On the `assert!`

Taking it as-is. It is the right minimum: silent truncation producing plausible-but-wrong abundances is far worse than a loud stop, and bounding `total_labels` up front correctly bounds every intermediate `labels.len() as u32`, so one check covers the whole loop. `assert!` rather than `debug_assert!` is also right, since release is the only build where this is reachable.

We do want to revisit it more carefully before the next release, and your "widen to `u64`" suggestion is not quite what we would reach for either. The better shape is to choose the offset width when the packed representation is built — `u32` when the incidence count fits, `u64` only when it does not — so the common case keeps the smaller offsets and the large case simply works instead of failing. Filed separately so it does not hold this up.

## On the verification caveat

Appreciated, and agreed with the framing. #1092 asks for measured numbers on allocation work, and `sample_data`'s 15 transcripts cannot show memory behaviour — you said so yourself rather than letting it pass. Accepting on the basis that these claims are structural (strictly fewer copies of the same data, and sizing a buffer that was already going to be filled) rather than empirical. That reasoning does not extend to the contention items later in the audit, where the effect really does depend on scale and thread count.

Merging once CI is green.

BenjaminDEMAILLE:
Apologies — I force-pushed a rebase onto `develop` and clobbered your `a4938bd` in the process. It is restored (now `5864276`) and the branch is rebased on current `develop`; nothing else was lost.

The catch is right and I had missed it: `dashmap`'s `OwningIter` has no `size_hint`, so my `into_iter().collect()` reintroduced from-empty growth a few lines from where the same cost is being removed. Thanks.

I will use `--force-with-lease` against a freshly fetched ref from here on.

PR #1095, opened 2026-08-07
perf(quant): take the fragment-length snapshot once per batch

### LINKED ISSUE #1078: perf(quant): FLD snapshot taken per fragment instead of per batch, 6 shared atomics per fragment
Part of #1092.

## Problem

`record()` takes a fresh fragment-length-distribution snapshot **once per fragment**:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L603-L604

```rust
let fld_pmf = sh.fld.online_snapshot();
let fld_cmf = sh.fld.online_cmf_snapshot();
```

Each of those acquires an `RwLock` read guard and clones an `Arc`:

```rust
// crates/salmon-model/src/fld.rs
online_pmf: RwLock<Arc<Vec<f64>>>,
online_cmf: RwLock<Arc<Vec<f64>>>,
```

So the per-fragment cost is **6 read-modify-write atomics on 3 shared cache lines**: two `RwLock` acquire/release pairs, two `Arc` refcount increments, two decrements when the guards drop. Every worker thread hits the same words, so this is cache-line ping-pong that gets *worse* as `-p` grows.

The in-code comment calls these "a pair of cheap `Arc` clones". That is true single-threaded and false at 32-64 threads.

## Why it is avoidable

The snapshot only changes when `refresh_online()` runs, and that is called exactly once per mini-batch in `on_batch_complete`:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L1037

The snapshot is therefore **invariant across an entire batch**. Taking it per fragment re-reads a value that provably cannot have changed.

## Fix

Hoist both snapshots to the top of `process_record_batch` / `process_record_pair_batch`, and pass them into `record()` as `&[f64]` arguments.

This preserves the property the current code is careful about (every mapping of one fragment indexes the *same* immutable snapshot, so duplicate transcripts get identical fragment-length probabilities) and strengthens it: now every fragment in a batch shares one view instead of each fragment racing for its own.

Atomic traffic drops from `6 × num_fragments` to `6 × num_batches`.

## Acceptance criteria

- [ ] `record()` takes `pmf: &[f64], cmf: &[f64]` instead of calling `online_snapshot()` itself
- [ ] Both paired-end and single-end batch loops take the snapshot once, before the per-record loop
- [ ] `quant.sf` output is byte-identical on the sample dataset before/after
- [ ] Existing integration tests pass unchanged

BenjaminDEMAILLE:
Part of #1092. Fixes #1078.

**This PR is justified on determinism and code structure, not on throughput. Measured at scale, it makes no measurable throughput difference at any thread count.** The numbers are below.

### What it changes

`record()` runs once per fragment and began by taking its own fragment-length snapshot:

```rust
let fld_pmf = sh.fld.online_snapshot();
let fld_cmf = sh.fld.online_cmf_snapshot();
```

Both fields are `RwLock<Arc<Vec<f64>>>`, so that is two lock acquisitions and two `Arc` refcount round trips per fragment, on cache lines shared by every worker.

The snapshot is only replaced by `refresh_online()`, which runs once per mini-batch in `on_batch_complete`, so it is invariant across a batch. This takes it once at the top of each batch loop and passes it in.

### The semantic change, stated plainly

This is **not** a pure reordering of the same arithmetic. Previously a fragment late in a batch could pick up a snapshot that another thread had refreshed mid-batch; now every fragment in a batch uses the batch-start snapshot. **Fragments therefore use a uniformly staler FLD than before.** That changes what the fragment-length term means for a given fragment.

Consequently `quant.sf` is byte-identical to develop at `-p 1` **and nowhere else**. At higher thread counts there is a systematic shift, measured by the maintainers at 222–243 max TPM difference at `-p 32` against a 21–37 noise floor — small in context (r = 0.999986) and in the better direction, but real and not noise.

### Why it is still worth doing

It strengthens the invariant the per-fragment snapshot was introduced for in `4a25b3db` ("consistent FLD reads so duplicate transcripts stay symmetric"), rather than weakening it: previously two fragments in the same batch could see different snapshots, and now they cannot.

That shows up directly as reproducibility. Run-to-run spread of each binary against **itself**, 4 interleaved reps, max TPM difference over 207,249 transcripts:

| p | develop | this PR |
| --- | --- | --- |
| 1 | 0.000 | 0.000 |
| 8 | 9.656 | 7.284 |
| 16 | **42.296** | **7.094** |

**6.0x tighter at `-p 16`.** The maintainers independently measured 7x at `-p 32` on different hardware.

### Throughput: no change

Interleaved, 4 rounds, median wall clock against develop. 10M read pairs from ERR188044, GRCh38 cDNA release-110 (207,249 transcripts), 16-core Apple Silicon:

| p | develop | this PR | delta |
| --- | --- | --- | --- |
| 1 | 222.77s | 222.84s | +0.03% |
| 8 | 35.21s | 36.92s | +4.9% |
| 16 | 28.07s | 27.27s | −2.8% |

The p=8 and p=16 figures swing in both directions depending on which rounds are used (rounds 3–4, machine settled: +0.3% and −5.0%), which means they measure the machine, not the code. User CPU tracks wall throughout. Peak RSS flat.

**The original rationale for this PR — six atomic read-modify-writes per fragment collapsing to six per batch — does not materialise as throughput.** It was asserted from first principles and the measurement does not support it. Maintainer measurement at `-p 32` agrees.

A ~3.5% `-p 1` regression was observed by the maintainers on x86-64; it does not reproduce on arm64 (+0.03% wall, +0.07% user, interleaved). A plausible mechanism for the asymmetry: `record` gained two `&[f64]` parameters, and a slice is two registers where `&Arc<Vec<f64>>` is one, so this adds four register-equivalents to a per-fragment call — which x86-64 SysV (six register args) feels and AArch64 (eight) may absorb. Bundling the pair into one `&FldSnapshot` would restore a single register if that is worth closing off.

### Benchmark methodology

All timings interleave the binaries, rotating every round. A first blocked sweep (develop, then this PR, then #1099 back to back) produced an apparent +13%/+21% regression for whichever ran last, entirely from thermal drift; interleaving removed it completely.

Byte-comparison alone cannot settle parity here, because **develop is not run-to-run deterministic at fixed thread count** in the default path — see the noise-floor table above. Every cross-binary number is reported against that measured floor.

### DIFF
--- a/crates/salmon-quant/src/processor.rs
+++ b/crates/salmon-quant/src/processor.rs
@@ -514,10 +514,20 @@ fn record_discrete(sh: &Shared, maps: &[ScoredMapping], acc: &salmon_model::Disc
 ///
 /// Counting, library detection, and FLD training always run; `log_fm` is the
 /// current minibatch's forgetting mass (used only when `use_online`).
+///
+/// `fld_pmf`/`fld_cmf` are the batch's fragment-length snapshot pair, taken once
+/// by the caller (see the batch loops). They are passed in rather than read here
+/// because the snapshot only changes at `refresh_online`, which runs at the batch
+/// boundary — so re-reading it per fragment would re-acquire two `RwLock`s and
+/// bump two `Arc` refcounts, on cache lines every worker shares, to obtain a value
+/// that provably cannot have changed.
+#[allow(clippy::too_many_arguments)]
 fn record(
     sh: &Shared,
     maps: &[ScoredMapping],
     log_fm: f64,
+    fld_pmf: &[f64],
+    fld_cmf: &[f64],
     seqbias: Option<&mut (SBModel, SBModel)>,
     gcbias: Option<&mut GcFragModel>,
     posbias: Option<&mut (Vec<SimplePosBias>, Vec<SimplePosBias>)>,
@@ -578,9 +588,25 @@ fn record(
     // num_mapped`. C++ salmon likewise counts a fragment as mapped only once it
     // has a compatible mapping.
     sh.num_mapped.fetch_add(1, Ordering::Relaxed);
-    // Classify the fragment as orphan when its best compatible mapping has only
-    // one mate of a pair placed (PairedEndLeft / PairedEndRight). Single-end
-    // libraries never count as orphan here.
+    // Classify the fragment as orphan when only one mate of a pair was placed
+    // (PairedEndLeft / PairedEndRight). Single-end libraries never count as
+    // orphan here.
+    //
+    // Reading `compat[0]` is sound because every surviving mapping of a fragment
+    // carries the SAME status, so the first is as good as any:
+    //   * selective alignment: `map_read_pair_into` drops all orphans as soon as
+    //     one non-decoy concordant pair survives, and `finalize_mappings_counted_into`
+    //     never emits decoy mappings — so the set is either all-paired or
+    //     all-orphan;
+    //   * sketch: `map_read_pair_sketch_into` derives one `status` from the
+    //     fragment's `map_type` and stamps it on every accepted hit.
+    // Note `compat` is ordered by transcript id (finalize sorts by `tid`), not by
+    // score, so this must not be read as "the best mapping" — it is "the status
+    // this fragment mapped with", which is well defined.
+    debug_assert!(
+        compat.iter().all(|(m, _)| m.status == compat[0].0.status),
+        "a fragment's surviving mappings must share one status"
+    );
     if matches!(
         compat[0].0.status,
         MateStatus::PairedEndLeft | MateStatus::PairedEndRight
@@ -588,20 +614,18 @@ fn record(
         sh.num_orphan.fetch_add(1, Ordering::Relaxed);
     }
 
-    // Capture ONE immutable FLD snapshot pair (PMF + CMF) for this fragment — a
-    // pair of cheap `Arc` clones, no per-fragment allocation — and share it
-    // between the online posterior and the persistent eq-class weight. Indexing
-    // the same snapshot for every mapping guarantees that transcripts sharing a
-    // fragment length (in particular exact-duplicate transcripts) receive an
-    // identical fragment-length probability even if another thread refreshes the
-    // shared snapshot meanwhile. Reading the live FLD per mapping (racing atomic
-    // loads on a concurrently-updated distribution) returned slightly different
-    // values for the same length and let the VBEM α<1 prior break duplicate
-    // symmetry. Mirrors C++ salmon's per-worker `LogCMFCache`. Empty until the
-    // first online refresh (the pre-burn-in window, where the FLD term is not
-    // folded in anyway).
-    let fld_pmf = sh.fld.online_snapshot();
-    let fld_cmf = sh.fld.online_cmf_snapshot();
+    // `fld_pmf`/`fld_cmf` are ONE immutable FLD snapshot pair (PMF + CMF), taken
+    // once per batch by the caller and shared here between the online posterior
+    // and the persistent eq-class weight. Indexing the same snapshot for every
+    // mapping guarantees that transcripts sharing a fragment length (in
+    // particular exact-duplicate transcripts) receive an identical
+    // fragment-length probability even if another thread refreshes the shared
+    // snapshot meanwhile. Reading the live FLD per mapping (racing atomic loads
+    // on a concurrently-updated distribution) returned slightly different values
+    // for the same length and let the VBEM α<1 prior break duplicate symmetry.
+    // Mirrors C++ salmon's per-worker `LogCMFCache`. Empty until the first online
+    // refresh (the pre-burn-in window, where the FLD term is not folded in
+    // anyway).
 
     // Per-fragment bias-collection weights. With online (dual-phase) inference
     // these are abundance-aware posteriors `softmax(mass_t + log w_t)`, which
@@ -646,8 +670,8 @@ fn record(
                             sh.model_single_frag_prob,
                             sh.paired_lib,
                             sh.no_frag_length_dist,
-                            &fld_pmf,
-                            &fld_cmf,
+                            fld_pmf,
+                            fld_cmf,
                         );
                         let log_cov = if *w > 0.0 { w.ln() } else { f64::NEG_INFINITY };
                         (m.tid, log_cov + start_pos_prob + log_frag_prob)
@@ -762,8 +786,8 @@ fn record(
                 sh.model_single_frag_prob,
                 sh.paired_lib,
                 sh.no_frag_length_dist,
-                &fld_pmf,
-                &fld_cmf,
+                fld_pmf,
+                fld_cmf,
             )
         })
         .collect();
@@ -931,6 +955,15 @@ impl<'a, 'r> PairedParallelProcessor<RefRecord<'r>> for QuantProcessor<'a> {
         } else {
             0.0
         };
+        // One FLD snapshot pair for the whole batch. The shared snapshot is only
+        // replaced by `refresh_online`, which runs at the batch boundary, so a
+        // per-fragment re-read would take two `RwLock`s and bump two `Arc`
+        // refcounts — on cache lines shared by every worker — to fetch a value
+        // that cannot have changed. Taking it here also makes the per-fragment
+        // view *more* stable than before: previously a concurrent refresh could
+        // land between two fragments of the same batch.
+        let fld_pmf = sh.fld.online_snapshot();
+        let fld_cmf = sh.fld.online_cmf_snapshot();
         // Reused across the batch's reads (the `*_into` calls clear it), so the
         // per-fragment result Vec isn't reallocated per read.
         let mut maps: Vec<ScoredMapping> = Vec::new();
@@ -1021,6 +1054,8 @@ impl<'a, 'r> PairedParallelProcessor<RefRecord<'r>> for QuantProcessor<'a> {
                     &sh,
                     &maps,
                     log_fm,
+                    &fld_pmf,
+                    &fld_cmf,
                     seqbias.as_mut(),
                     gcbias.as_mut(),
                     posbias.as_mut(),
@@ -1085,6 +1120,9 @@ impl<'a, 'r> ParallelProcessor<RefRecord<'r>> for QuantProcessor<'a> {
         } else {
             0.0
         };
+        // One FLD snapshot pair for the whole batch (see the paired-end loop).
+        let fld_pmf = sh.fld.online_snapshot();
+        let fld_cmf = sh.fld.online_cmf_snapshot();
         // Reused across the batch's reads (the `*_into` calls clear it).
         let mut maps: Vec<ScoredMapping> = Vec::new();
         for rec in records {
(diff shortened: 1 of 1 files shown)


rob-p:
I ran this at real scale before merging — 10M paired reads from ERR188044 against a GRCh38 cDNA index (206,723 transcripts), `-p` 1 / 8 / 32 on a 256-core host, develop vs this branch. The results change what this PR is *for*, so it is worth laying them out.

Note first that `sample_data` cannot show any of this, and neither can byte-comparison on its own: **develop is not run-to-run deterministic at fixed thread count** in the default path. Two develop runs at `-p 8`, same binary, same input, differ. So every comparison below is against a measured noise floor rather than an equality check.

## Determinism: a clear, substantial win

Max TPM difference and Pearson r over all transcripts:

| comparison | max TPM diff | r |
| --- | --- | --- |
| develop vs **itself**, p=32 | 21.0 – 37.5 | 0.9999996 |
| this PR vs **itself**, p=32 | **3.0 – 4.7** | **0.99999999** |
| develop vs PR, p=1 | **0.000** | **1.0000000000** |

The PR's own run-to-run spread is roughly **7x tighter** than develop's. That is mechanistically exactly what the change should do: before, a fragment late in a batch could pick up an FLD snapshot refreshed by *another thread* mid-batch, so what it saw depended on timing. Now every fragment in a batch uses the batch-start snapshot. This strengthens the invariant the original per-fragment snapshot was introduced for (4a25b3db, "consistent FLD reads so duplicate transcripts stay symmetric") rather than weakening it, which was our main concern going in.

## Parity: the claim does not hold at scale

| comparison | max TPM diff | r |
| --- | --- | --- |
| develop vs PR, p=32 | **221.9 – 242.8** | 0.999986 |

That is 6-11x *outside* either binary's own noise floor, so it is a systematic shift, not reshuffling. The magnitude is small in context (r = 0.99999, 222 out of 1e6 total TPM) and it moves in the better direction, but "byte-identical" holds only at `-p 1` and on 15-transcript `sample_data`.

The mechanism is not subtle in hindsight: fragments now use a **staler** FLD than before, uniformly. That is a change in what the fragment-length term means for a given fragment, not a reordering of the same arithmetic. Worth stating plainly in the PR description.

## Performance: not measurably faster, and slower at `-p 1`

Wall clock, three runs each at p=32: develop 22.16 / 21.58 / 21.41 (mean 21.72s), PR 21.01 / 21.76 / 20.38 (mean 21.05s). Overlapping ranges — noise on a shared host.

Across the sweep:

| p | develop wall | PR wall | user CPU |
| --- | --- | --- | --- |
| 1 | 247.8s | 256.7s | **+3.5% (slower)** |
| 8 | 44.75s | 43.24s | −3.8% |
| 32 | 21.12s | 20.96s | −3.4% |

Peak RSS is flat (1429 vs 1339 MB at p=32).

So the headline rationale — 6 atomic RMWs per fragment collapsing to 6 per batch — **does not materialise as throughput** at any thread count we tested.

## What we would like

Two things, and I would rather you ran them than take our numbers on faith, since you know the change best and the harness should live with the PR:

1. **Benchmark this at scale yourself**, on a real index and read set, and report wall/CPU at 1, 8 and 32 threads — the bar #1092 sets. If the throughput win is genuinely absent, say so; the determinism result is a better justification anyway and we would take the PR on that basis.

2. **Look at whether the `-p 1` regression is avoidable.** ~3.5% single-threaded is small but it is the wrong direction, and single-threaded is the configuration where the change can only cost (there is no contention to remove). If it is inherent to passing slices instead of `Arc` clones, fine — but it is worth one look before we accept a slowdown in exchange for a win we cannot measure.

To be explicit about the standard, so it does not read as arbitrary: **we would rather not merge changes that should theoretically improve performance but practically do not**, especially when they carry a measurable regression elsewhere. Structural simplifications and determinism improvements are their own justification and we are happy to take them as such — but then the PR should say that, and not claim a performance benefit that the numbers do not support.

The same request applies to #1099, which reduces atomic traffic on the same reasoning. Given that the per-fragment to per-batch reduction here yielded ~0%, the six-counter conversion there may well be structurally sound and performance-neutral too. Please measure it the same way before we take it on performance grounds.

Happy to share the harness (index build, sweep script, the noise-floor controls) if useful — the controls are the fiddly part, since without them the byte-differences at p>1 are uninterpretable.

BenjaminDEMAILLE:
Benchmarked as asked. Headline: **your conclusion reproduces — no throughput win — and the determinism result reproduces too. Your `-p 1` regression does not reproduce here, but see the caveat.**

## Setup

10M read pairs from ERR188044 against a GRCh38 cDNA index (Ensembl release-110, **207,249** transcripts). Machine is a **16-core Apple Silicon (arm64)**, 128 GB, macOS — so `-p` 1 / 8 / 16 rather than your 1 / 8 / 32. Different architecture and an order of magnitude fewer cores than your host; that matters below.

**All timings are interleaved.** My first (blocked) sweep ran develop, then #1095, then #1099 back to back, and #1099 came out +13% at `-p 8` and +21% at `-p 16` — which was pure machine drift from running last after an hour of continuous load. Rotating the binaries every round made that entirely disappear. Flagging it because it is exactly the trap that produces a confident wrong number, and blocked sweeps are the obvious way to write this harness.

## Throughput: no win, confirming your result

Interleaved, 4 rounds, median wall vs develop:

| p | develop | #1095 | delta |
| --- | --- | --- | --- |
| 1 | 222.77s | 222.84s | **+0.03%** |
| 8 | 35.21s | 36.92s | +4.9% |
| 16 | 28.07s | 27.27s | −2.8% |

The p=8 and p=16 numbers swing in *both* directions depending on which rounds you take (rounds 3–4, once the machine settles: p=8 +0.3%, p=16 −5.0%), which is the signature of measuring the machine rather than the code. User CPU tracks wall throughout.

So: **no measurable throughput change at any thread count.** The "6 atomic RMWs per fragment becoming 6 per batch" rationale does not show up as throughput, exactly as you found.

## `-p 1`: your regression does not reproduce

| | develop | #1095 |
| --- | --- | --- |
| median wall | 222.77s | 222.84s (+0.03%) |
| median user | 221.82s | 221.97s (+0.07%) |

Four interleaved rounds each; runs 225.6 / 223.3 / 222.4 / 222.0 against 228.9 / 224.1 / 221.4 / 221.4. Flat.

**I would not treat this as contradicting you.** Two reasons to prefer your number:

1. **Architecture.** AArch64 passes eight arguments in registers; x86-64 SysV passes six. `record` gained two `&[f64]` parameters, and a slice is *two* registers (pointer + length) where an `&Arc<Vec<f64>>` is one — so the change adds four register-equivalents to a call made once per fragment. That is a plausible mechanism for a real x86-64 regression that arm64 absorbs, and it predicts exactly the asymmetry we are seeing.
2. If your sweep was blocked rather than interleaved, ordering could account for some of it — but I have no way to know, and the mechanism above stands on its own.

If it is worth closing off, the fix suggested by that mechanism is cheap: bundle the pair into one `&FldSnapshot { pmf, cmf }` so it is a single register again. Say the word and I will do it and re-measure; I did not want to change the PR on a hypothesis I cannot reproduce on my own hardware.

## Determinism: reproduces, ~6x

Run-to-run spread of each binary against *itself*, 4 interleaved reps, max TPM difference over all 207,249 transcripts:

| p | develop | #1095 |
| --- | --- | --- |
| 1 | 0.000 | 0.000 |
| 8 | 9.656 | 7.284 |
| 16 | **42.296** | **7.094** |

**6.0x tighter at p=16**, against your 7x at p=32. On completely different hardware, which is about as good as this kind of result gets.

## Parity: where I differ from you, and I think you are right

| p | develop vs #1095, max TPM diff | develop's own noise floor |
| --- | --- | --- |
| 1 | **0.000** | 0.000 |
| 8 | 6.061 | 9.656 |
| 16 | 28.579 | 42.296 |

At p=8 and p=16 the cross-binary difference sits *below* develop's own run-to-run spread, so on this machine I cannot distinguish it from noise. You measured 222–243 against a 21–37 floor at p=32 — clearly outside.

I do not think these conflict: 32 threads exercises the mid-batch-refresh race far harder than 16, and your numbers are the ones that should be believed. My machine is simply not big enough to resolve the effect. **Your framing is the correct one and I have written it into the PR description:** fragments now use a uniformly staler FLD, which is a change in what the fragment-length term means for a given fragment, not a reshuffle of the same arithmetic. Byte-identical holds at `-p 1` and nowhere else.

## PR description updated

Rewritten to drop the throughput claim entirely and justify the change on determinism, with the staleness semantics stated plainly and these numbers inline. Same treatment on #1099, measured the same way.

Agreed on the standard, and it caught a real error in my own reasoning: I had asserted a contention win from first principles and the machine says otherwise.

rob-p:
You were right and I was wrong. Re-measured properly, and the regression I reported does not exist.

## The `-p 1` regression does not replicate

Three-way interleaved on the same 10M library, warm-up round discarded by design, paired analysis (which is the correct test for interleaved runs — it cancels drift):

| | mean | sd |
| --- | --- | --- |
| develop | 248.69s | 2.15 |
| #1095 | 248.75s | 1.05 |

**+0.06s (+0.02%), paired t = 0.05 on 4 df.** That is as close to zero as this kind of measurement gets.

My original figure came from a **single blocked run** — my sweep looped `for bin in develop pr1095; do for p in 1 8 32`, so develop always ran first and each configuration was measured once. Your point 2 was exactly right, and it accounts for the whole thing. You could not reproduce the regression on your hardware because it is not there, not because your machine is smaller. I should not have reported 3.5% off n=1.

## I tried the `FldSnapshot` bundle anyway, and the assembly is interesting

Since the mechanism was plausible on its own terms, I implemented it and looked at the generated code. Two findings worth recording so nobody retries this.

**First, I initially got it wrong**: I wrote `fld: FldSnapshot<'_>` — *by value*. That is 32 bytes, byte-for-byte identical to the two slices it replaces, so it cannot reduce register pressure at all; and SysV classifies aggregates over 16 bytes as MEMORY, so it is passed on the stack, plausibly worse than two 16-byte slices in register pairs. Any future attempt at this must use `&FldSnapshot` (8 bytes), which is what you actually proposed.

**Second, the by-reference version does what you predicted — and still loses.** `record` is the only meaningfully changed function in the whole object:

| metric | two slices | `&FldSnapshot` | delta |
| --- | --- | --- | --- |
| instructions | 2012 | 2025 | **+13** |
| stack (`%rsp`) references | 378 | 370 | **-8** |
| loads via pointer register | 76 | 84 | **+8** |
| stack frame | 408 B | 392 B | **-16 B** |

**Your register-pressure argument is confirmed**: the frame shrinks by exactly one slice's worth and eight stack references disappear. It just does not pay. The pointer has to be dereferenced at every *use*, so those eight stack references become eight pointer indirections, plus 13 net instructions — and `record` touches `pmf`/`cmf` once per mapping, so the indirection is paid repeatedly while the register saving is banked once per call. Wall clock agreed: +0.54% against develop, numerically the worst of the three.

So the intuition "fewer arguments is cheaper" is right about the call and wrong about the body, for this access pattern. Not taking the bundle.

## Merging

Determinism win reproduced independently on both our machines (your ~6x at p=16, mine ~7x at p=32), parity is byte-identical at `-p 1`, throughput is flat, and the description already states the staleness semantics honestly. The `4a25b3db` invariant it strengthens — one immutable snapshot per fragment so exact-duplicate transcripts get identical fragment-length probabilities — is preserved and made stronger by the per-batch hoist.

Thanks for pushing back on the number rather than taking it. That is twice now that measurement has corrected me on this audit, and the standard is better for it.

Issue #1083, opened 2026-08-07
perf(map): alignment gap/flank cache allocates its key on every probe, including cache hits


BenjaminDEMAILLE:
Part of #1092.

## Problem

The per-thread gap/flank alignment memo cache allocates its lookup key on **every probe, including cache hits**:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-map/src/align.rs#L470-L504

```rust
fn cached_gap_score(qg: &[u8], tg: &[u8], cfg: &AlignConfig) -> i32 {
    ...
    GAP_CACHE.with(|c| {
        let key = (
            qg.to_vec().into_boxed_slice(),
            tg.to_vec().into_boxed_slice(),
        );
        if let Some(&s) = c.borrow().get(&key) {
            return s;          // <- two allocations and two frees, for nothing
        }
        ...
```

`cached_flank_score` does the same, plus an extra copy to append the orientation tag byte.

These are called once per inter-MEM gap and twice per candidate flank, i.e. several times per candidate placement per read. The cache exists to skip a DP, but it pays two heap allocations and a full byte-wise hash *before* it can tell whether it will skip anything.

The comment above `GAP_CACHE` concedes that "cross-read reuse is rare (flanks are read-specific)". So on the flank path the steady state is close to a 100% miss rate: allocate key, hash bytes, miss, run the DP, allocate again for the insert. Pure overhead on top of the work it was supposed to avoid.

## Secondary concern: key aliasing

Gap and flank results share one map with two different score semantics (`ksw2_gap_global` vs `ksw2_flank_extend`). They are only kept apart by `cached_flank_score` appending a `0`/`1` byte to the query key:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-map/src/align.rs#L494-L496

That holds only because no ACGT base equals `0x00` or `0x01`. It is an implicit invariant on read content with no assertion behind it, guarding a cache that may not be earning its keep.

## Fix

Options, in increasing order of change:

1. **Hash-keyed cache.** Key on a `u64` xxh3 digest of `(kind_tag, qlen, q, tlen, t)`; store `(qlen, tlen, score)` as the value so a digest collision can be detected and treated as a miss. `xxhash-rust` is already a workspace dependency. Zero allocation on both hit and miss.
2. **Borrowed lookup.** Use `hashbrown`'s raw entry API to probe with a borrowed key and only allocate on insert. Removes the hit-path cost, keeps the miss-path cost.
[...]

PR #1096, opened 2026-08-07
perf(rad,map): remove per-fragment and per-probe allocations from the hot paths

### LINKED ISSUE #1082: perf(rad): process_rad_fragment allocates a BTreeMap plus 2 Vecs per transcript, per fragment
Part of #1092.

## Problem

`process_rad_fragment` runs once per RAD record and allocates roughly `2n + 7` times, where `n` is the number of distinct transcripts the fragment maps to:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-align/src/rad.rs#L324-L336

```rust
let mut agg: std::collections::BTreeMap<u32, Vec<usize>> = std::collections::BTreeMap::new();
for (k, &t) in sp_tid.iter().enumerate() {
    agg.entry(t).or_default().push(k);
}
let tids: Vec<u32> = agg.keys().cloned().collect();
let eq_log: Vec<f64> = agg
    .values()
    .map(|ks| logsumexp(&ks.iter().map(|&k| sp_eq[k]).collect::<Vec<_>>()))
    .collect();
```

Per fragment that is:

- one `BTreeMap` (plus one node allocation per distinct tid, scattered across the heap)
- one `Vec<usize>` per distinct tid
- one **more** `Vec<f64>` per distinct tid, built only to be passed to `logsumexp` and dropped immediately
- plus `sp_tid`, `sp_eq`, `tids`, `eq_log`, `weights`, `bins`
- plus the `Vec<RadPlacement>` that `RadFragment::placements()` returns by value on every record (https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-align/src/rad.rs#L105)

At 500M fragments with n=5 that is on the order of 8 billion malloc/free pairs, and the `BTreeMap` nodes give the worst possible locality for what is fundamentally a group-by over a handful of `u32`.

The identical pattern is duplicated in the bias pass:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-align/src/rad.rs#L850-L857

## Fix

A group-by over a few `u32` keys does not need a tree.

1. Change `RadFragment::placements` to `placements_into(&self, out: &mut Vec<RadPlacement>)` and hand it a per-worker scratch buffer.
2. Accumulate `(tid, aux)` into a reusable `Vec<(u32, f64)>`, `sort_unstable_by_key(|x| x.0)`, then walk it once doing a running log-sum-exp per run of equal tids (track the running max and the running sum of `exp(x - max)`, rescaling when the max moves).

That drops the per-fragment allocation count to zero for the intermediates, and to exactly the two vectors (`tids`, `weights`) that are genuinely moved into the equivalence-class builder.

Both call sites (`process_rad_fragment` and the bias-pass equivalent) should share the helper rather than duplicating it a third time.

## Acceptance criteria

- [ ] No `BTreeMap` and no per-transcript temporary `Vec` on the per-fragment path
- [ ] `placements()` no longer allocates per record
- [ ] The grouping helper is shared between the eq-class pass and the bias pass
- [ ] RAD round-trip tests (`crates/salmon-rad/tests/roundtrip.rs`, `crates/salmon-align/tests/rad_input.rs`) pass unchanged
- [ ] `quant.sf` from a RAD requant is byte-identical before/after

BenjaminDEMAILLE:
Part of #1092. Fixes #1082 and #1083.

**Justified on correctness and structure, not on throughput.** Measured at scale by the maintainers, both changes are flat. Numbers below.

### The `GapKind` fix is the part that earns this PR

Gap and flank scores come from *different* functions (`ksw2_gap_global` vs `ksw2_flank_extend`) over the same kind of byte pair, and were kept apart only by appending a `0`/`1` byte to the query key. That is correct solely because no ACGT base equals `0x00` or `0x01` — an unasserted assumption about read content guarding a silently-wrong-score failure mode. It is now an explicit `GapKind` field.

### Allocation removal

**Alignment cache.** `cached_gap_score` / `cached_flank_score` built their lookup key by copying both substrings into `Box<[u8]>` *before* asking whether the entry existed, so every probe cost two heap allocations and two frees — cache hits included. `hashbrown::HashTable` can be probed with the borrowed `(kind, query, target)` triple, so only a miss allocates now, and once instead of twice.

Measured hit rates on 10M paired reads, GRCh38 cDNA:

| cache | probes | hits | hit rate |
| --- | --- | --- | --- |
| gap | 5,044,438 | 4,550,348 | 90.2% |
| flank | 113,786,498 | 57,614,418 | 50.6% |

So the change removes roughly two allocations on each of ~62M hits. Worth noting the flank cache avoids ~57.6M DP calls per run and is probed 22:1 against the gap cache — I had earlier suggested, on the strength of a wrong in-code comment, that it might be near-useless and worth deleting. It is not. The comment is corrected in this PR with the measured numbers.

These caches are on the default (anchored) scoring path only: `--fullLengthAlignment` takes the whole-read `ksw2_align_score` branch and never reaches them (measured: zero probes).

**RAD grouping.** `process_rad_fragment` and `collect_bias_fragment` each built a `BTreeMap<u32, Vec<usize>>` per record to group a handful of placements by transcript — a map allocation, a tree node and a `Vec` per distinct transcript, plus one further temporary `Vec` per transcript purely to hand `logsumexp` a slice. Roughly `2n + 7` allocations per fragment. `RadFragment::placements` likewise returned a fresh `Vec` per record; it becomes `placements_into`.

### Throughput: no measurable change

| path | develop | this PR |
| --- | --- | --- |
| mapping (cache change), 3 runs | 22.04s | 21.83s |
| `--rad` requant (grouping change), 3 runs | 12.16s | 12.15s |

Both within noise. The `2n+7` allocations per fragment and the two allocations on each of ~62M cache hits are **not** where the time goes. The PR does not claim a throughput benefit.

### Parity

- `--rad` requant of a 10M-fragment RAD at `-p 1`: byte-identical. This is the check that matters for the grouping's ordering argument, on real data rather than `sample_data`'s 15 transcripts.
- `--seqBias --gcBias`: byte-identical.
- `sample_data`: byte-identical across plain / bias / `--softclip` / RAD write / RAD requant / RAD requant with bias.

The ordering property is load-bearing: `logsumexp` is a floating-point sum, so groups must come out in ascending tid order and placements must keep arrival order within a group. Follow-up commit `9a35c7e` (maintainer) restructures the grouping to hold each placement's data in one record rather than four hand-aligned parallel arrays, which removes the same class of index hazard that #1099 hit, and measured that a stable sort subsumes the `(tid, index)` permutation key (100% of fragments arrive already sorted, mean 6.8 placements per fragment).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2738,6 +2738,7 @@ name = "salmon-map"
 version = "2.4.1"
 dependencies = [
  "ahash",
+ "hashbrown 0.15.5",
  "ksw2rs",
  "piscem-rs",
  "salmon-core",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,9 @@ scc = "3"
 dashmap = "6"
 # fast (non-DoS-resistant) hasher for hot per-read integer-keyed maps
 ahash = "0.8"
+# raw hash table, for caches that must probe with a borrowed key (no key
+# allocation on a lookup that hits)
+hashbrown = "0.15"
 # numerics
 statrs = "0.19"
 # rng (PCG, matching salmon's pcg usage)
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -419,6 +419,22 @@ fn pair_records(recs: &[FragRecord]) -> Vec<Placement> {
     placements
 }
 
+/// `log(Σ exp(xs))` over an iterator, numerically stable. `xs` is non-empty.
+///
+/// Same two passes and the same summation order as [`logsumexp`], so the two
+/// agree bit for bit; this form exists so a caller holding a run of records can
+/// avoid copying their weights into a scratch slice first.
+pub(crate) fn logsumexp_iter<I>(xs: I) -> f64
+where
+    I: Iterator<Item = f64> + Clone,
+{
+    let m = xs.clone().fold(f64::NEG_INFINITY, f64::max);
+    if m == f64::NEG_INFINITY {
+        return f64::NEG_INFINITY;
+    }
+    m + xs.map(|x| (x - m).exp()).sum::<f64>().ln()
+}
+
 /// `log(Σ exp(xs))`, numerically stable. `xs` is non-empty.
 fn logsumexp(xs: &[f64]) -> f64 {
     let m = xs.iter().cloned().fold(f64::NEG_INFINITY, f64::max);
@@ -2305,5 +2321,51 @@ mod tests {
         let opts = AlignQuantOptions::new("x.rad".into(), "out".into());
         assert_eq!(opts.fld_policy, FldPolicy::Baked);
         assert!(!opts.explicit_fld_args.any());
+    }
+}
+
+#[cfg(test)]
+mod logsumexp_tests {
+    use super::{logsumexp, logsumexp_iter};
+
+    /// `logsumexp_iter` exists so a caller holding a run of records can avoid
+    /// copying their weights into a scratch slice. It is only safe to use in
+    /// place of [`logsumexp`] if it agrees *bit for bit* — the eq-class weights
+    /// it produces feed range factorization, where a last-bit difference can
+    /// move a fragment across a bin boundary and change the class set.
+    #[test]
+    fn iter_form_agrees_bit_for_bit_with_the_slice_form() {
+        let cases: &[&[f64]] = &[
+            &[-1.0],
+            &[-1.0, -1.0],
+            &[0.0, -1.0, -2.0, -3.0],
+            &[-1e-16, -1.0, -1e-16],
+            &[-700.0, -700.5, -701.25],
+            &[f64::NEG_INFINITY, -2.0],
+            &[f64::NEG_INFINITY, f64::NEG_INFINITY],
+            &[-0.1, -0.2, -0.3, -0.4, -0.5, -0.6, -0.7],
+        ];
+        for xs in cases {
+            let a = logsumexp(xs);
+            let b = logsumexp_iter(xs.iter().copied());
+            assert_eq!(
+                a.to_bits(),
+                b.to_bits(),
+                "slice={a} iter={b} for {xs:?} — the two forms must not diverge"
+            );
+        }
+    }
+
+    /// Order matters to the result, which is why the grouping sorts stably; the
+    /// two forms must agree on that too, not merely on sorted input.
+    #[test]
+    fn iter_form_matches_on_reordered_input() {
+        let xs = [-1e-16, -1.0, -1e-16, -2.0];
+        let mut rev = xs;
+        rev.reverse();
+        assert_eq!(
+            logsumexp(&rev).to_bits(),
+            logsumexp_iter(rev.iter().copied()).to_bits()
+        );
     }
 }
--- a/crates/salmon-align/src/rad.rs
+++ b/crates/salmon-align/src/rad.rs
@@ -229,38 +229,160 @@ impl Grouped for BiasPlacement {
     }
 }
 
-/// Sort `placements` by transcript and write each transcript's combined log
-/// weight into `group_log`.
+/// What became of one fragment, decided inside the traversal that already walks
+/// its placements.
 ///
-/// Groups come out in ascending tid order and, within a group, placements keep
-/// their arrival order. Both are load-bearing: `logsumexp` is a floating-point
-/// sum, so a different order would perturb the last bits of every eq-class
-/// weight and could push a fragment across a range-factorization bin boundary,
-/// changing the class set rather than only its values. A *stable* sort by `tid`
-/// gives both directly.
+/// Mirrors the alignment path's outcome so both report the concordant/orphan
+/// split the same way, and so the tally does not have to re-walk the slice to
+/// work it out.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum FragmentOutcome {
+    /// no placement survived library-compatibility filtering
+    Unmapped,
+    /// assigned, with at least one placement pairing both mates
+    Mapped,
+    /// assigned in a paired library, but every surviving placement placed only
+    /// one mate. Matches the RAD fragment-level rule, where the most complete
+    /// status across hits wins: a proper pair on one transcript and an orphan on
+    /// another counts as a pair.
+    MappedOrphan,
+}
+
+/// Whether the caller has already ordered a fragment's placements by tid.
 ///
-/// Salmon's own RAD writer emits placements already ascending and
-/// duplicate-free (measured: 100% of fragments across stranded, unstranded and
-/// sample inputs), which the sort handles by itself — a stable sort detects the
-/// existing run and moves nothing. The format does not require it, so the
-/// general case still has to work, and does.
-fn group_by_tid<T: Grouped>(placements: &mut [T], group_log: &mut Vec<f64>) {
-    placements.sort_by_key(|p| p.tid());
-    group_log.clear();
-    group_log.extend(
-        placements
-            .chunk_by(|a, b| a.tid() == b.tid())
-            .map(|run| logsumexp_iter(run.iter().map(|p| p.log_w()))),
-    );
+/// The fused eq-class + bias pass orders them once for both of its consumers, so
+/// neither needs to work out again what the other already knows. The
+/// single-consumer passes leave them in arrival order and let the builder derive
+/// it, which costs the same comparisons a sort would have spent detecting the
+/// run — just folded into a traversal that was happening anyway.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum PlacementOrder {
+    /// arrival order; the builder derives whether it happens to be ordered
+    Arrival,
+    /// already ascending by tid
+    ByTid,
+}
+
+/// A fragment's placements, tracking whether they are already ordered by tid.
+///
+/// The flag is *derived on push*, never set by a caller. A boolean that records
+/// a property someone has to remember to maintain is the same
+/// convention-over-construction hazard this module removed from the parallel
+/// arrays it used to keep index-aligned by hand: push without updating it and
+/// the grouping is silently wrong. Maintained here, it cannot disagree with the
+/// contents.
+///
+/// It costs one comparison per push — the same comparisons a sort would spend
+/// detecting an existing run, moved into a loop where the element is already in
+/// a register. What it buys is that the fused eq-class + bias pass stops
+/// ordering the same fragment twice: it sorts the placements once, both
+/// builders then push in ascending order, and neither sorts again. That falls
+/// out of the invariant rather than being a special case either builder has to
+/// know about.
+struct TidOrdered<T> {
+    items: Vec<T>,
+    /// whether `items` is non-decreasing by tid
+    sorted: bool,
+}
+
+impl<T> Default for TidOrdered<T> {
+    fn default() -> Self {
+        Self {
+            items: Vec::new(),
+            // An empty run is trivially ordered.
+            sorted: true,
+        }
+    }
+}
+
+impl<T: Grouped> TidOrdered<T> {
+    fn clear(&mut self) {
+        self.items.clear();
+        self.sorted = true;
+    }
+
+    fn is_empty(&self) -> bool {
+        self.items.is_empty()
+    }
+
+    fn push(&mut self, item: T) {
+        if let Some(last) = self.items.last() {
+            self.sorted &= last.tid() <= item.tid();
+        }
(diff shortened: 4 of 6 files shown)


rob-p:
Measured this at real scale and pushed a follow-up restructure (9a35c7e). Numbers first, since they change what the PR is for.

## The cache hit rates justify the change better than either of us argued

10M paired reads, GRCh38 cDNA, default path:

| cache | probes | hits | hit rate |
| --- | --- | --- | --- |
| gap | 5,044,438 | 4,550,348 | **90.2%** |
| flank | 113,786,498 | 57,614,418 | **50.6%** |

Stable across re-runs. **I owe you a correction here**: I had earlier suggested the flank cache might be near-useless and worth deleting outright, on the strength of the in-code comment conceding that "cross-read reuse is rare (flanks are read-specific)". That comment is simply wrong — the cache avoids ~57.6M DP calls per run, and flank probes outnumber gap probes 22:1, so the flank path is exactly where the allocator traffic was. Worth fixing the comment while you are in there.

Also worth noting for the description: `--fullLengthAlignment` takes the whole-read `ksw2_align_score` branch and never reaches these caches (measured: 0 probes), so this optimisation is default-path only.

## Parity holds at scale

- `--rad` requant of a 10M-fragment RAD, `-p 1`: **byte-identical**. That is the check that matters for the `(tid, index)` stability argument — you are right that it reproduces a stable sort exactly, and it does so on real data, not just `sample_data`'s 15 transcripts.
- `--seqBias --gcBias`: **byte-identical**.

## Throughput: no measurable change

| path | develop | this PR |
| --- | --- | --- |
| mapping (cache change), 3 runs | 22.04s | 21.83s |
| `--rad` (grouping change), 3 runs | 12.16s | 12.15s |

Both within noise. The `2n+7` allocations per fragment and the two allocations on each of ~61M cache hits are simply not where the time goes. For context, this is the fourth and fifth structurally-motivated change from the audit to measure flat — see #1092, where I will write that up properly, because I think it should redirect where the remaining items get judged.

None of that is a reason to reject this PR. The `GapKind` change alone earns it: gap and flank scores were kept apart only by appending `0`/`1` to the query key, which is correct solely because no ACGT base equals `0x00` or `0x01` — an unasserted assumption about read content guarding a wrong-score failure mode. Good catch.

## What the follow-up commit changes, and why

The grouping rewrite is correct but organised as four structures that must stay index-aligned by hand: `pairs` and `geom` (both per placement), `order` (a permutation), `group_start` (run boundaries). That is why it reads as `pairs[order[group_start[g]]].0`, and why `geom[k]` must be kept in step with `pairs[k]` with nothing enforcing it — the same hazard we just made unrepresentable in #1099, here guarded by nothing. I proved it is live by breaking exactly that alignment in an earlier attempt and getting three bias-test panics.

Holding each placement's data in one record removes it. Groups become subslices, consumers `zip` runs against the per-group buffers, and every `[]` disappears from the grouping and the bias loop. Five buffers collapse to two.

Two further consequences:

- **The `(tid, index)` permutation key is no longer needed.** A stable sort by tid gives both properties directly. It was there to dodge the temp buffer a stable sort allocates, but at mean 6.8 placements per fragment the stable sort is in-place insertion sort and allocates nothing.
- **No already-sorted fast path is needed either.** I instrumented the input: **100% of fragments arrive ascending and duplicate-free**, across stranded, unstranded (`-l IU`) and sample inputs — so the grouping is a no-op in practice and a stable sort detects the run and moves nothing. I tried an explicit `is_sorted_by` fast path; it measured 0.86% (t=1.32), i.e. nothing, and added a second code path to maintain. The sort subsumes it.

**Two record types, which matters more than it looks.** A single record carrying the geometry measured 64 bytes, of which `process_rad_fragment` reads 12 — it needs only tid and weight, so it would sort and scan 4x the memory for fields it never touches. `CorePlacement` is 16 bytes for that path; `BiasPlacement` is 40 for the path that wants the geometry (`u32` positions, since `Option<usize>` is 16 bytes against `Option<u32>`'s 8). One generic `group_by_tid`, monomorphised. The single fat record measured nominally *slower* than your version (-1.17%, t=1.06); split, it is dead flat (-0.11%, t=-0.12).

Also adds the first tests for the grouping — including the duplicate-tid combining path, which nothing salmon currently writes will exercise — and `logsumexp_iter` with a test asserting bit-for-bit agreement with the slice form, since these weights feed range factorization where a last-bit difference can change the class set rather than only its values.

BenjaminDEMAILLE:
Both asks done, and thank you for the correction on the flank cache — I was the one who put that inference in the issue, so it is my wrong claim you disproved.

**Comment fixed** (`3f82a69`). It now carries your measured rates rather than the "cross-read reuse is rare (flanks are read-specific)" line, plus a note that the flank path is probed 22:1 against the gap path and avoids ~57.6M DP calls, so it is the opposite of the near-useless thing I suggested deleting. Also records that these caches are default-path only — `--fullLengthAlignment` never reaches them.

**Description rewritten**: no throughput claim, justified on the `GapKind` correctness fix and the structural work, with your hit rates, timing and parity numbers inline.

On `9a35c7e`: the four hand-aligned arrays were exactly the hazard from #1099 with nothing enforcing them, and `geom[k]` keeping step with `pairs[k]` by convention was the weakest part of what I wrote. Two details I would not have got to:

- **The `(tid, index)` key was unnecessary.** I added it specifically to avoid the stable sort's temp buffer, without checking that at mean 6.8 placements the stable sort is in-place insertion sort and allocates nothing. I optimised against a cost that was not there.
- **100% of fragments arrive already sorted**, so the grouping is a no-op in practice — and you measured the explicit fast path at 0.86% (t=1.32) and dropped it rather than keeping a second code path for nothing. The split into `CorePlacement` (16 B) and `BiasPlacement` (40 B) because the single 64-byte record sorts 4x the memory the eq-class path reads is the kind of thing I would have missed entirely.

On the wider point: five structurally-motivated changes from this audit now measure flat. I think your read is right, and I would rather say so plainly than keep attaching perf rationales to them. The ones that have held up are the correctness items — the `GapKind` aliasing here, the silent `u32` CSR truncation in #1094, the compat-index hazard in #1099 — and the determinism improvement in #1095. The allocation-count arguments have not.

I will fold that into #1092 rather than have you write it up, unless you would rather do it yourself since you have the measurements.

rob-p:
Pushed the conflict resolution and a test for the cache. `MERGEABLE`/`CLEAN` again.

## Conflicts

Two, both in the RAD per-fragment loops: this branch's buffer reuse (`placements_into` plus the `&mut scratch` grouping) against develop's orphan tally, which replaced the two shared `fetch_add`s with `RadTallies::tally`. Compatible — keep the reused buffer and the scratch argument, call `tally` where the atomics were. `tally` reads `placements.first()` and the buffer is cleared and refilled per fragment, so it sees what it saw before.

Re-verified byte-identical against **current** develop (which has since gained the orphan tallies, RAD provenance, libradicl 0.17 and the CSR widening): plain, `--seqBias --gcBias --posBias`, `--writeRad`, `--rad` requant, and the written RAD file itself.

## The unsorted case, since every measurement so far was the fast path

The 100%-already-sorted finding means none of the earlier numbers exercised the reordering at all. Grouping cost per fragment, ns, where `perm` is the previous design (sort an 8-byte `(tid, idx)` key once, gather both payloads through it):

| n | order | core 16B | bias 40B | fused A+B | perm 8B |
| --- | --- | --- | --- | --- | --- |
| 7 | **sorted** | 7.8 | 6.2 | 14.0 | 10.6 |
| 7 | shuffled | 12.2 | 17.1 | **29.3** | **15.7** |
| 7 | reversed | 22.0 | 26.0 | **47.9** | **18.2** |
| 20 | **sorted** | 7.7 | 10.0 | 17.6 | 27.8 |
| 20 | shuffled | 46.7 | 72.7 | **119.4** | **57.6** |
| 100 | shuffled | 445 | 584 | **1029** | **435** |

The trade goes both ways, and the records are the right side of it. On sorted input they win decisively for the plain path (7.7 vs 27.8 at n=20) because a stable sort detects the run and moves nothing while the permutation still pays for a gather. On unsorted input the permutation wins ~2–2.6x — but that is 13.6 ns/fragment at n=7, **0.14 s per 10M fragments** against a ~12 s RAD quant. A producer that emits unsorted placements costs a few percent, not a cliff.

Two things that fell out of measuring it:

- the pathological order is **shuffled, not reversed** — at n=100 reversed is 144 ns against shuffled's 1029, because the sort detects a descending run and reverses in O(n)
- in the fused bias pass the same fragment is sorted **twice**, once as `core` and once as `bias`. Sorting `pls` itself once up front would make both downstream sorts run-detections, but it is worth ~6 ns/fragment (60 ms per 10M) and it would put `RadTallies::tally`'s `placements.first()` on a reordered slice, which is safe only if "a fragment's surviving mappings share one status" holds — currently an unasserted assumption of exactly the shape the `GapKind` fix removes. Noting it rather than proposing it.

## Test for the cache

`GapKind` is the strongest thing in this PR and nothing exercised the cache, so the bug it fixes could return unnoticed. `HashTable` also requires a hand-written hash/eq pair kept in step by a comment — the same convention-not-construction hazard removed from the RAD grouping, with asymmetric failure modes: an omission in `hash` costs hits, an omission in `eq` returns the wrong DP score.

Four tests: all three kinds stay distinct for identical bytes; the query/target split is part of the key (`seqs` is a concatenation, so the same bytes divided differently are a different DP); entries survive the table growing; the cap clears wholesale.

The growth one is worth its keep — it pins the rehash closure to each entry's own bytes. Deriving it from the key being inserted would leave older entries under a hash no probe ever computes, silently turning the cache off while staying correct. I misread the closure as doing exactly that before checking the source, which is roughly how it would reach review.

Both hazards negative-controlled: dropping `kind` from `eq` fails the first test, rehashing from the inserted key fails the third.

One bookkeeping note: this closes #1082 and #1083, and merges into `develop` do not auto-close, so they will need closing by hand.

rob-p:
Pushed a third change: the fused pass now orders each fragment's placements **once** instead of once per consumer.

## The redundancy

`run_rad_eq_and_bias_pass` calls both `process_rad_fragment` and `collect_bias_fragment` on the same fragment. Each built its own vector from the same placements, in the same order, and sorted it — the same ordering decided twice. The payloads genuinely differ (eq-class weight against un-normalized posterior) so each needs its own pass over the runs, but neither needs its own sort.

The fused pass now orders once and says so through a `PlacementOrder` marker; `TidOrdered::push_sorted` skips the comparison `push` would make, since the sort already established the answer. A `debug_assert` holds the caller to that promise in tests and compiles out of release. The single-consumer passes keep deriving it during the build, which costs the comparisons a sort would have spent detecting the run, folded into a traversal that was happening anyway.

## Measured with criterion

| n | order | before | after |
| --- | --- | --- | --- |
| 7 | sorted | 12.40 ns | 12.40 ns |
| 7 | shuffled | 33.99 ns | **21.45 ns (−37%)** |
| 7 | reversed | 46.78 ns | **27.36 ns (−41%)** |
| 20 | sorted | 30.04 ns | **27.69 ns (−8%)** |
| 20 | shuffled | 159.2 ns | **70.67 ns (−56%)** |
| 20 | reversed | 216.8 ns | **96.11 ns (−56%)** |

Never slower; materially faster on input we do not control.

**End-to-end is null, as it should be**: +0.11%, t=0.26, 95% CI −0.254 .. +0.328 on 11 paired rounds of the fused path over 10M human fragments. salmon's own writer emits placements already ordered, so the improved case never arises there. It arises for a piscem RAD, or any future producer — which is exactly the "the format does not require it, so the general case still has to work" point already in this PR's comments.

## A correction on method, since I got this wrong twice before landing it

My first two attempts measured the *opposite* result and I reported them as findings. Two defects, both in my harness rather than the code:

1. It charged the sorting variants for a `copy_from_slice` the others never paid — sorting mutates, so those needed a fresh copy each iteration. In production there is no such copy: `placements_into` fills the buffer regardless and the sort operates on what was going to be filled anyway. I was measuring my own scaffolding.
2. I never checked its repeatability. Re-running the same binary moved one cell from −38.6% to +99.7%.

Criterion's intervals here are ±0.2%. The lesson is the one this PR's discussion already reached about structural performance claims, applied to the measurement apparatus itself: validate the harness before quoting it.

## Also in this push

`RadTallies::tally` takes the decision rather than the placements. `process_rad_fragment` already walks them and already knows which survived, so working it out again was a second pass — and reading it off `placements[0]` was correct only while they happened to arrive in a particular order, which the fused pass no longer guarantees. It now counts **surviving** placements, matching the alignment path.

Byte-identical across plain / bias / RAD write / RAD requant / RAD requant with bias, and the RAD bytes themselves. 24 tests in `rad.rs` including one asserting `push_sorted` rejects an out-of-order push; clippy `-D warnings` and fmt clean.

Issue #1085, opened 2026-08-07
perf(map,quant): finish the *_into scratch pattern through the mapping and record path


BenjaminDEMAILLE:
Part of #1092.

## Problem

The `*_into` buffer-reuse pattern is applied at the top of the mapping call chain (`map_single_read_into`, `map_read_pair_into`, `finalize_mappings_counted_into`) but stops there. Everything below still allocates per read or per fragment.

### `record()` in salmon-quant, 5 to 7 allocations per fragment

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L553 `compat`
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L626 `mm`
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L658 `bias_w`
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L754 `log_fps`
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L786 `pairs`
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L806-L807 `tids`, `weights`

Only `tids` and `weights` need to outlive the call (they move into the equivalence-class builder). The other five are pure scratch.

Additionally https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L660 does `post.clone()`, deep-copying the online posterior vector when a borrow would do.

### `finalize_mappings_counted_into`, a HashMap per fragment

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-map/src/score.rs#L170

```rust
let mut best_per_tid: AHashMap<u32, RawMapping> = AHashMap::new();
```

A fresh hash map, allocated and dropped per fragment, to deduplicate what is typically 1 to 20 entries. At that size a `Vec` with a linear scan (or sort plus dedup) is both faster and allocation-free.

### Mapper, 4 or more Vecs per fragment

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-map/src/mapper.rs#L237-L239 `left`, `right`, `joints`, `raw`
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-map/src/collect.rs#L145 `cfg.chain.clone()` per read
[...]

PR #1099, opened 2026-08-07
perf(quant): per-worker counters and reusable buffers in the fragment path

### LINKED ISSUE #1084: perf(quant): up to 7 shared-atomic RMWs per fragment limit thread scaling
Part of #1092.

## Problem

The per-fragment path performs up to seven read-modify-write operations on shared atomics, one per counter:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L525
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L580
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L588
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L828-L842

`num_processed`, `num_mapped`, `num_orphan`, `num_decoy`, `num_dovetail`, `num_frags_filtered_vm`, `num_below_threshold_vm` are all `&AtomicU64` shared by every worker. There is also a shared `load` on the same hot path:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L745

The RAD reader has the same shape, two per record:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-align/src/rad.rs#L553-L556

Relaxed ordering makes these cheap in isolation, but the cost is not the ordering, it is the cache line: every thread is writing the same handful of lines, so each increment serialises behind a coherence transfer. This is a scaling tax that grows with `-p` and is invisible at low thread counts.

## Fix

The merge infrastructure already exists. `QuantProcessor` accumulates per-thread bias models and unmapped names privately and merges them once in `on_thread_complete`:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/processor.rs#L1044-L1051

Apply the same pattern to the counters: plain `u64` fields on `QuantProcessor`, incremented with ordinary adds, folded into the shared atomics with one `fetch_add` each in `on_thread_complete`.

The one counter that needs care is `num_processed`, because it is also *read* on the hot path to gate `use_aux`. See the companion issue on `use_aux` determinism; the two should be resolved together, since a per-thread counter changes what that gate observes.

## Acceptance criteria

- [ ] Per-fragment counter updates are non-atomic thread-local adds
- [ ] Shared atomics are touched once per thread, in `on_thread_complete`
- [ ] Same treatment applied to the RAD worker loops
- [ ] Final counter values in `meta_info.json` and `lib_format_counts.json` unchanged on the sample dataset
- [ ] Scaling measured at 1 / 8 / 32 threads before and after

BenjaminDEMAILLE:
Part of #1092. Fixes #1084, and the `record()` half of #1085.

**Stacked on #1095.** Justified on allocation and structure, **not on throughput**: measured at scale it makes no measurable throughput difference at any thread count. Numbers below.

### Per-worker counters

Six of the seven shared counters (`num_mapped`, `num_orphan`, `num_decoy`, `num_dovetail`, `num_frags_filtered_vm`, `num_below_threshold_vm`) were `fetch_add` on `AtomicU64`s shared by every worker. They are now plain `u64` on the worker, folded into the shared atomics once in `on_thread_complete`, where the bias models already merge.

`num_processed` deliberately stays shared: it is also *read* on the hot path to gate `use_aux`, so thread-localising it would change which fragments see the auxiliary model, and therefore the output. That is the #1084/#1089 ordering constraint.

### Reusable buffers in `record()`

Five throwaway `Vec`s per fragment (compatible mappings, online-posterior inputs, bias weights, per-mapping fragment-length terms, tid/weight pairs) move into a worker-owned `RecordScratch`. Only `tids` and `weights` stay fresh, since they are moved into the equivalence-class builder.

`compat` holds **indices** into `placements` rather than references, so the buffer can outlive a fragment.

### Two index spaces, now type-checked

That renumbering introduced a real hazard: `bias_w` is indexed by position within `compat`, not by placement index, and the two coincide only when nothing is strand-filtered. I got it wrong first time (caught by an unused-variable warning); it would have mis-weighted GC and positional bias on **stranded libraries specifically**, silently.

`ae61096` (maintainer) makes that unrepresentable rather than merely commented, with `MapIdx` / `CompatIdx` / `CompatAligned<T>`. Reverting to the original mistake now fails to build. Converting also surfaced that `online_post` is compat-aligned *and* read under a different guard than `bias_w`, which review had not caught.

### Throughput: no change

Interleaved, 4 rounds, median wall against develop. 10M read pairs from ERR188044, GRCh38 cDNA release-110 (207,249 transcripts), 16-core Apple Silicon, built from `ae610967`:

| p | develop | this PR | delta |
| --- | --- | --- | --- |
| 1 | 222.77s | 220.01s | −1.24% |
| 8 | 35.21s | 34.85s | −1.0% |
| 16 | 28.07s | 27.39s | −2.4% |

User CPU tracks wall. Consistently a hair on the fast side but inside run-to-run variation, so **no throughput claim is made**. The atomic-traffic rationale does not materialise as throughput, matching the same finding on #1095.

### Determinism and parity

Run-to-run spread against itself, 4 interleaved reps, max TPM difference:

| p | develop | this PR |
| --- | --- | --- |
| 1 | 0.000 | 0.000 |
| 8 | 9.656 | 6.600 |
| 16 | 42.296 | 10.775 |

Inherited from #1095 rather than added here.

Against develop: byte-identical at `-p 1`. At `-p 8`/`-p 16` the difference (5.6 / 28.6) sits below develop's own noise floor on this machine, so it cannot be resolved here; #1095 carries the parity caveat, since **develop is not run-to-run deterministic at fixed thread count** and the staleness change originates there.

### Benchmark methodology

Timings interleave the binaries, rotating every round. A first blocked sweep made this PR look +13% at `-p 8` and +21% at `-p 16` purely because it ran last after an hour of load; interleaving removed all of it.

### DIFF
--- a/crates/salmon-quant/src/processor.rs
+++ b/crates/salmon-quant/src/processor.rs
@@ -468,14 +468,14 @@ fn collect_context(
 /// per-alignment `observedGCMass.inc(gcDesc(start, stop), logProb)`.
 fn collect_gc(
     sh: &Shared,
-    maps: &[ScoredMapping],
+    placements: &[ScoredMapping],
     compat: &[(usize, f64)],
     bias_w: &[f64],
     gc: &mut GcFragModel,
 ) {
     let Some(store) = sh.gc_store else { return };
     for (k, &(i, _)) in compat.iter().enumerate() {
-        let m = &maps[i];
+        let m = &placements[i];
         if m.status != MateStatus::PairedEndPaired || m.fragment_len <= 0 {
             continue;
         }
@@ -499,7 +499,7 @@ fn collect_gc(
 /// `observedPosBias{Fwd,RC}[lengthClass].addMass(pos, refLen, logProb)`.
 fn collect_pos(
     sh: &Shared,
-    maps: &[ScoredMapping],
+    placements: &[ScoredMapping],
     compat: &[(usize, f64)],
     bias_w: &[f64],
     pos: &mut (Vec<SimplePosBias>, Vec<SimplePosBias>),
@@ -508,7 +508,7 @@ fn collect_pos(
         return;
     };
     for (k, &(i, _)) in compat.iter().enumerate() {
-        let m = &maps[i];
+        let m = &placements[i];
         // `bias_w` is aligned to `compat`, indexed by `k`, not by `i`.
         if bias_w[k] <= 0.0 {
             continue;
@@ -535,12 +535,12 @@ fn collect_pos(
 /// accumulator at end of pass and baked into the RAD.
 fn record_discrete(
     sh: &Shared,
-    maps: &[ScoredMapping],
+    placements: &[ScoredMapping],
     acc: &salmon_model::DiscreteFld,
     counters: &mut Counters,
 ) {
     sh.num_processed.fetch_add(1, Ordering::Relaxed);
-    if maps.is_empty() {
+    if placements.is_empty() {
         return;
     }
     counters.mapped += 1;
@@ -548,8 +548,8 @@ fn record_discrete(
     // orientation; its observed format also feeds order-independent `-l A`
     // detection. Uses the raw mapping (like the RAD records written this pass, and
     // like the reader's `derive_fld`), so a baked FLD matches a re-derived one.
-    if maps.len() == 1 {
-        let m = &maps[0];
+    if placements.len() == 1 {
+        let m = &placements[0];
         if m.status == MateStatus::PairedEndPaired && m.fragment_len > 0 {
             // Orientation from the mate strands (sketch mode leaves `m.format`
             // `None`), exactly as `build_rad_record` derives the RAD hit.
@@ -563,7 +563,7 @@ fn record_discrete(
     // leaves `m.format` `None`, so derive the observed paired format from the mate
     // strands (as `build_rad_record` / `DiscreteFld` do).
     if let Some(nb) = sh.naive_eq {
-        let sig: Vec<NaivePlacement> = maps
+        let sig: Vec<NaivePlacement> = placements
             .iter()
             .map(|m| {
                 let fmt_id = if m.status == MateStatus::PairedEndPaired {
@@ -611,7 +611,7 @@ fn record_discrete(
 #[allow(clippy::too_many_arguments)]
 fn record(
     sh: &Shared,
-    maps: &[ScoredMapping],
+    placements: &[ScoredMapping],
     log_fm: f64,
     fld_pmf: &[f64],
     fld_cmf: &[f64],
@@ -624,7 +624,7 @@ fn record(
     // Still a shared atomic: unlike the other counters this one is *read* below,
     // to gate `use_aux`. See `Counters`.
     sh.num_processed.fetch_add(1, Ordering::Relaxed);
-    if maps.is_empty() {
+    if placements.is_empty() {
         return;
     }
     // Independent `&mut` borrows of the reusable buffers.
@@ -641,7 +641,7 @@ fn record(
     // below, so detection sees the true observed orientation/strand.
     if let Some(det) = sh.detector {
         if det.is_active() {
-            if let Some(m) = maps
+            if let Some(m) = placements
                 .iter()
                 .filter(|m| m.format.is_some())
                 .max_by(|a, b| a.weight.total_cmp(&b.weight))
@@ -659,21 +659,26 @@ fn record(
     let expected = sh
         .expected_format
         .or_else(|| sh.detector.and_then(|d| d.resolved_format()));
-    // Indices into `maps` rather than references, so the buffer can outlive the
+    // Indices into `placements` rather than references, so the buffer can outlive the
     // call and be reused by the next fragment.
     compat.clear();
-    compat.extend(maps.iter().enumerate().filter_map(|(i, m)| match expected {
-        Some(exp) => {
-            if is_compatible(exp, m.format, m.is_fw, m.status) {
-                Some((i, m.weight))
-            } else if sh.ignore_incompat {
-                None
-            } else {
-                Some((i, m.weight * sh.incompat_prior))
-            }
-        }
-        None => Some((i, m.weight)),
-    }));
+    compat.extend(
+        placements
+            .iter()
+            .enumerate()
+            .filter_map(|(i, m)| match expected {
+                Some(exp) => {
+                    if is_compatible(exp, m.format, m.is_fw, m.status) {
+                        Some((i, m.weight))
+                    } else if sh.ignore_incompat {
+                        None
+                    } else {
+                        Some((i, m.weight * sh.incompat_prior))
+                    }
+                }
+                None => Some((i, m.weight)),
+            }),
+    );
     if compat.is_empty() {
         return; // no compatible mapping -> fragment is unassigned (not mapped)
     }
@@ -705,11 +710,11 @@ fn record(
     debug_assert!(
         compat
             .iter()
-            .all(|&(i, _)| maps[i].status == maps[compat[0].0].status),
+            .all(|&(i, _)| placements[i].status == placements[compat[0].0].status),
         "a fragment's surviving mappings must share one status"
     );
     if matches!(
-        maps[compat[0].0].status,
+        placements[compat[0].0].status,
         MateStatus::PairedEndLeft | MateStatus::PairedEndRight
     ) {
         counters.orphan += 1;
@@ -750,7 +755,7 @@ fn record(
                 let use_aux = online.num_assigned() >= sh.pre_burnin;
                 online_in.clear();
                 online_in.extend(compat.iter().map(|&(i, w)| {
-                    let m = &maps[i];
+                    let m = &placements[i];
                     {
                         let ref_len = sh.salmon.ref_len(m.tid as usize);
                         let rl = ref_len.max(1) as f64;
@@ -800,14 +805,14 @@ fn record(
     if collect_now {
         if let Some(obs) = seqbias {
             for (k, &(i, _)) in compat.iter().enumerate() {
-                collect_context(sh.salmon, &maps[i], bias_w[k], obs);
+                collect_context(sh.salmon, &placements[i], bias_w[k], obs);
             }
         }
         if let Some(gc) = gcbias {
-            collect_gc(sh, maps, compat, bias_w, gc);
+            collect_gc(sh, placements, compat, bias_w, gc);
         }
         if let Some(pos) = posbias {
-            collect_pos(sh, maps, compat, bias_w, pos);
+            collect_pos(sh, placements, compat, bias_w, pos);
         }
     }
 
@@ -825,7 +830,7 @@ fn record(
         // best pair at full weight, which overdisperses it). Frozen after the
         // training window (`online_post` is `None`).
         for (k, &(i, _)) in compat.iter().enumerate() {
-            let m = &maps[i];
+            let m = &placements[i];
             let conc = m.status == MateStatus::PairedEndPaired
                 && m.fragment_len > 0
                 && expected.is_none_or(|exp| is_compatible(exp, m.format, m.is_fw, m.status));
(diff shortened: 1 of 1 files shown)


rob-p:
Audited this one in detail. The structure is sound; two requests, one of which is the same measurement ask as #1095.

## What is clearly correct

- **The counter conversion is mechanical and right.** `merge_counters` is wired into *both* `on_thread_complete` impls (single-end and paired), the `if v != 0` guard skips no-op merges, and only the two `num_processed` increments remain on the hot path.
- **Leaving `num_processed` shared is the right call**, for the reason you give: it is *read* to gate `use_aux`, so making it thread-local would change which fragments see the auxiliary model. That is the #1084/#1089 ordering constraint respected rather than tripped over.
- **The scratch buffers are safe.** All five have exactly one `clear()` immediately before their fill, in the same function. `tids`/`weights` correctly stay fresh, since they move into the eq-class builder.

## The `i` / `k` renumbering

You flagged this yourself and you were right to. I checked all four `bias_w[...]` sites (492, 513, 520, 803) — every one indexes by `k`, the position in `compat`. `bias_w` is rebuilt per fragment as one entry per `compat` element on both paths (`online_post` and the score-weight fallback), and `online_in` is likewise built from `compat`, so the alignment holds. `log_fps` and `pairs` use `.zip()` rather than indexing, which is structurally immune.

So it is correct as written. The residual concern is that the invariant "`bias_w` is indexed by compat position" is now enforced only by comments and reviewer attention, and getting it wrong is silent and **stranded-library-specific** — the one case where `compat.len() != maps.len()` reliably.

We would rather make that a compile-time property than a review property, and we are prototyping it here rather than asking you to: collapsing the parallel `compat` / `bias_w` / `log_fps` arrays into a single `Vec` of records, so there is no second index to get wrong and no length invariant to assert. If it comes out performance-neutral we will push it to this branch for you to look at.

Incidentally, `maps: &[ScoredMapping]` reads as associative to anyone who did not write it, which makes "index into `maps`" harder to follow than it should be. A rename to `placements` or `mappings` would help.

## The measurement request

Same as on #1095. We benchmarked that PR at real scale (10M paired reads, 206,723-transcript index, `-p` 1/8/32) and its atomic-traffic reduction produced **no measurable throughput change** at any thread count, plus a ~3.5% *regression* at `-p 1`.

This PR reduces atomic traffic on the same reasoning, so the same question applies: please measure wall and CPU at 1, 8 and 32 threads on a real index and read set, and report it here. If the win is not there, the buffer-reuse and structural work still stands on its own merits — but we would rather the PR say so than claim a performance benefit the numbers do not support.

Note that byte-comparison alone will not settle parity: develop is not run-to-run deterministic at fixed thread count in the default path, so differences at `-p > 1` need a measured noise floor (repeat runs of the same binary) to interpret. Happy to share our harness.

rob-p:
Thanks for the `placements` rename — pushed the typed-index change on top of it (ae61096). Here is the reasoning, since we went round the houses on the design before landing on this.

## What the audit found

Most of this PR is clear and correct:

- **The counter conversion is mechanical and right.** `merge_counters` is wired into *both* `on_thread_complete` impls, the `if v != 0` guard skips no-op merges, and only the two `num_processed` increments remain on the hot path.
- **Leaving `num_processed` shared is the right call**, for exactly the reason you give — it is read to gate `use_aux`, so thread-localising it would change which fragments see the auxiliary model. That is the #1084/#1089 ordering constraint respected rather than tripped over.
- **The scratch buffers are safe.** All five have exactly one `clear()` immediately before their fill, in the same function; `tids`/`weights` correctly stay fresh since they move into the eq-class builder.
- **The `i`/`k` renumbering is correct as written.** All four `bias_w[...]` sites index by compat position. `bias_w` is rebuilt per fragment as one entry per `compat` element on both paths, `online_in` is likewise built from `compat`, and `log_fps`/`pairs` use `.zip()` rather than indexing, which is structurally immune.

So the only thing we wanted to change was how the remaining invariant is *held*.

## Why types rather than a struct

Our first attempt was to fold the parallel arrays into one `Vec` of records, so misalignment would be unrepresentable rather than merely checked. We prototyped it, and rejected it, for two reasons:

**It did not generalise.** Collapsing `compat`/`bias_w`/`log_fps` into records left `online_post` still separate and still compat-aligned, guarded by a `debug_assert_eq!`. The first new aligned buffer immediately reintroduced the hazard the refactor was meant to remove — which is the tell that this is a recurring *category*, not three specific arrays.

**It was not faster.** Three buffers becoming one is a performance claim, so we measured it: 21.22s mean against 20.82s for this branch at `-p 32`, overlapping ranges. No win. Having just asked you to justify changes on measurements rather than theory, it would have been poor form to push a structural rewrite of your PR on the strength of an unmeasured tidiness argument.

Typed indices are the better answer because indices are *unavoidable* here — `placements[i]` has to happen — so the fix is to make the index carry its meaning rather than chase arrays into structs one at a time.

## What landed

```rust
#[repr(transparent)] struct MapIdx(usize);       // into `placements`
#[repr(transparent)] struct CompatIdx(usize);    // position within `compat`
#[repr(transparent)] struct CompatAligned<T>(Vec<T>);   // Index<CompatIdx> only
```

`MapIdx::get(placements)` replaces `placements[i]`, so a placement index cannot reach a compat-aligned buffer. Everything is `#[repr(transparent)]` with `#[inline]` accessors — same generated code.

**The guard is demonstrated, not asserted.** Reverting to the original mistake now fails to build:

```
error[E0308]: mismatched types
    gc.inc(ff, cf, bias_w[i]);
                          ^ expected `CompatIdx`, found `MapIdx`
```

That is the bug you caught yourself via the unused-`k` warning — the one that would have mis-weighted bias on stranded libraries only. It is now a build failure rather than a thing review has to notice.

## The bit that justifies the whole exercise

Converting surfaced **`online_post`**: it is compat-aligned too, and it is read at the FLD-training site under a *different* guard than `bias_w` (`online_post.is_some()` versus `collecting`). So the two are not interchangeable even though both hold per-compat weights — folding them together, as our struct prototype nearly did, would have been wrong.

We found it by hand-writing the prototype and nearly getting it wrong. The type system found it immediately, as the last remaining compile error. That is a better argument for this approach than anything we reasoned about in advance.

## Verification

- Byte-identical `quant.sf` at `-p 1` on 10M paired reads (ERR188044) against a 206,723-transcript GRCh38 cDNA index.
- Throughput unchanged: 20.56 / 21.66 / 21.26s over three runs at `-p 32`, against 20.32 / 20.42 / 21.71s before. Overlapping ranges, no regression.
- `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean (two unused `CompatAligned` accessors deleted rather than `#[allow]`ed), full suite green.

The benchmark request from the earlier comment still stands for the PR as a whole — this change is output- and throughput-neutral, so it does not affect what you measure.

BenjaminDEMAILLE:
Measured this one the same way, on the branch **including your `ae61096`** (rebuilt from `ae610967`, not from my `4a777f85`). Same setup as the #1095 report: 10M pairs of ERR188044, GRCh38 cDNA release-110 (207,249 transcripts), 16-core Apple Silicon, all timings interleaved across binaries.

## Throughput: no win here either

Interleaved, 4 rounds, median wall vs develop:

| p | develop | #1099 | delta |
| --- | --- | --- | --- |
| 1 | 222.77s | 220.01s | −1.24% |
| 8 | 35.21s | 34.85s | −1.0% |
| 16 | 28.07s | 27.39s | −2.4% |

User CPU tracks wall (`-p 1`: −1.27%). Consistently very slightly on the fast side, but well inside the run-to-run variation, so I am not claiming it. **No measurable throughput change** — same answer as #1095.

Worth recording how nearly I got this wrong. My first sweep ran the binaries in blocks, develop → #1095 → #1099, and #1099 came out **+13% at p=8 and +21% at p=16**. That was entirely machine drift from running last after an hour of sustained load; interleaving removed all of it. I would have reported a serious regression in your typed-index commit that does not exist.

## Determinism

Run-to-run spread of each binary against itself, 4 interleaved reps, max TPM difference across all transcripts:

| p | develop | #1099 |
| --- | --- | --- |
| 1 | 0.000 | 0.000 |
| 8 | 9.656 | 6.600 |
| 16 | 42.296 | 10.775 |

3.9x tighter at p=16 — inherited from #1095, which this stacks on, rather than added here.

Parity against develop: `0.000` at `-p 1` (byte-identical), and at p=8/16 the cross-binary difference (5.6 / 28.6) sits below develop's own noise floor on this machine, so I cannot resolve it. Your p=32 numbers on #1095 should be taken as the better measurement; the PR description now states the staleness semantics rather than claiming byte-parity.

## On `ae61096`

Took the typed indices and read them properly. Two things I want to acknowledge rather than just accept:

**You rejected your own better-looking refactor on a measurement.** The struct-of-records version is the one I would have reached for, and "it did not generalise, and we measured it and it was not faster" is a more useful answer than the tidier diff would have been. Point taken about not pushing structural rewrites on unmeasured tidiness arguments — that is the same standard you asked of me, applied to yourself first.

**The `online_post` find is the real result.** That it is compat-aligned *and* read under a different guard (`online_post.is_some()` vs `collecting`) is exactly the class of thing that survives review — I had stared at that function repeatedly and not seen it. Having the type system produce it as the last compile error is a much better argument for the approach than the invariant-preservation case.

The `MapIdx`/`CompatIdx` split also makes the bug I originally introduced unrepresentable rather than merely commented, which is the right place for it. My `debug_assert` was a note to the next reader; yours is a build failure.

## PR description

Rewritten: no throughput claim, justified on the buffer-reuse and structural grounds plus the inherited determinism improvement, with these numbers inline and the parity caveat stated.

BenjaminDEMAILLE:
Rebased onto `develop` now that #1095 has merged — the stack base moved and it went CONFLICTING. Your `ae61096` is preserved (now `fc602b7`); nothing of yours was dropped, and I used `--force-with-lease` against a freshly fetched ref this time.

Suite green, clippy clean under the workspace lints, fmt clean.

rob-p:
Rebased onto current `develop` and pushed. `MERGEABLE`/`CLEAN` again — but one conflict was **not** mechanical, and this branch can no longer do part of what it did.

## #1107 took the counter this branch was thread-localising

Three of the four conflicts were routine: the `maps` → `placements` rename against develop's RAD write-error propagation (#1106), twice, and the orphan tally added to `record_discrete` (#1108), which becomes a per-worker `counters.orphan`.

The fourth: #1107 changed the `use_aux` burn-in gate to read `num_mapped`, where it previously read `num_processed`.

```rust
let assigned_before_this = sh.num_mapped.fetch_add(1, Ordering::Relaxed);
...
let use_aux = assigned_before_this >= sh.pre_burnin;
```

This branch had moved `mapped` to a per-worker tally merged at thread completion. Combined, the shared counter would have sat at **zero for the entire run**, so every fragment would be gated as pre-burn-in and the auxiliary models would never engage. A silent behaviour change — it compiles, the counter totals still come out right at the end, and only the abundances would be wrong.

So `mapped` stays shared and is removed from `Counters`. This is precisely the reasoning your review already applied to `num_processed`:

> Leaving `num_processed` shared is the right call, for the reason you give: it is *read* to gate `use_aux`, so making it thread-local would change which fragments see the auxiliary model.

That constraint has simply moved to `num_mapped`. The other five counters are write-only until the end and stay per-worker — which is where the contention was anyway: **five of the seven per-fragment RMWs are gone**, and the two that remain are the two that are read.

## Verification

The per-worker merge is only meaningfully exercised when the counters are non-zero, and on `sample_data` all five are zero — a parity run there could not have caught a mistake in this. On 10M human fragments against develop:

| counter | 1099 p1 | dev p1 | 1099 p16 | dev p16 |
| num_processed | 10,000,000 | 10,000,000 | 10,000,000 | 10,000,000 |
| num_mapped | 9,221,625 | 9,221,625 | 9,221,625 | 9,221,625 |
| num_orphan | 623,154 | 623,154 | 623,154 | 623,154 |
| num_dovetail_fragments | 22,754 | 22,754 | 22,754 | 22,754 |
| num_fragments_filtered_vm | 94,181 | 94,181 | 94,181 | 94,181 |
| below_threshold_vm | 13,990,843 | 13,990,843 | 13,990,843 | 13,990,843 |
| percent_mapped | 92.21625 | 92.21625 | 92.21625 | 92.21625 |

Identical in all four combinations, so the merge is thread-count independent as well as correct. `sample_data` parity also holds byte-for-byte across plain / bias / RAD write / RAD requant / RAD requant with bias and the RAD bytes.

One process note, since it nearly cost me: my first parity run reported `MISSING` for every comparison and `DIFFERS` for the RAD bytes. That was a broken script — repeated `sed` edits had left the comparison looking for output directories the runs no longer wrote — not a real result. The tell was that the counter table printed fine from the same run directories.

Issue #1098, opened 2026-08-07
bug(align): alignment mode produces a different quant.sf on every run (thread-order-dependent FLD accumulation)


BenjaminDEMAILLE:
Found while parity-checking an unrelated change: on unmodified `develop` (`fcccaa4`), alignment mode produces a **different `quant.sf` on every run** with the same input and the same flags. Reads mode does not.

I do not know whether this is considered acceptable — `--deterministic` exists and does fix it — so this is a report rather than a claim that it is wrong.

## Reproduction

`sample_data` from the repo, index built from `transcripts.fasta`.

```
# alignment mode, default flags
for i in 1 2 3 4 5 6; do
  salmon quant -t transcripts.fasta -l A -a sample_alignments.bam -p 8 -o run-$i
done
```

| configuration | runs | distinct `quant.sf` |
| --- | --- | --- |
| `-a`, `-p 1` | 3 | **1** |
| `-a`, `-p 2` | 4 | **4** |
| `-a`, `-p 4` | 6 | **6** |
| `-a`, `-p 8` | 6 | **6** |
| `-a`, `-p 16` | 4 | **4** |
| `-a --deterministic`, `-p 8` | 5 | **1** |
| `-i` (reads), `-p 8` | 8 | **1** |
| `-i --deterministic`, `-p 8` | 4 | **1** |

Every multi-threaded alignment-mode run differs from every other. `-p 1` is stable, so it is thread interleaving rather than anything input-dependent.

## Where it diverges

`aux_info/meta_info.json`, run 1 vs run 2:

```
-  "frag_length_mean": 178.51084891510467,
-  "frag_length_sd": 28.664480640290147,
+  "frag_length_mean": 178.51084891509896,
+  "frag_length_sd": 28.664480640290144,
```

~1e-11 relative. `aux_info/fld.gz` is byte-identical across runs (it is quantised), and the eq-class dump is unaffected — so the eq-class machinery is doing its job. The divergence enters through the fragment-length distribution's summary statistics, then propagates: effective lengths → EM → `quant.sf`.

## Mechanism

`process_fragment` is documented as safe to run in parallel and calls, for every fragment:

https://github.com/COMBINE-lab/salmon/blob/develop/crates/salmon-align/src/lib.rs#L1159-L1162

```rust
if frag_len > 0 {
    ctx.fld.add_val(frag_len as usize, 0.0);
}
```

`add_val` folds each observation into two *single shared* accumulators:

[...]

PR #1100, opened 2026-08-07
perf(infer): build the packed CSR layout once per run

### LINKED ISSUE #1087: perf(infer): PackedEqClasses rebuilt 2 to 4 times per run; weights unused outside Gibbs
Part of #1092.

## Problem

`PackedEqClasses::from_collapsed` copies the entire CSR representation (`labels`, `starts`, `combined`, `weights`, `counts`). It is called two to four times per run, on the same data.

`optimize` and `optimize_with_init` each build one internally:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-infer/src/lib.rs#L169
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-infer/src/lib.rs#L187

so a bias-corrected quant run builds it at least twice, and a third time for bootstrap/Gibbs:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/lib.rs#L785 (pre-EM / warm-up)
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/lib.rs#L969 (warm-started final EM)
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-quant/src/lib.rs#L1021 (bootstrap / Gibbs)

The RAD path is worse, up to four:

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-align/src/rad.rs#L1497
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-align/src/rad.rs#L1499
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-align/src/rad.rs#L1565
https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-align/src/rad.rs#L1599

Between the pre-EM and the final EM the only thing that changed is `combined_weights`, recomputed by `update_eff_lengths` after bias correction. `labels`, `starts` and `counts` are bit-identical across all of those rebuilds.

For 5M classes averaging 4 members that is roughly 400 MB copied each time.

## Related waste: `weights` is dead outside Gibbs

https://github.com/COMBINE-lab/salmon/blob/master/crates/salmon-infer/src/packed.rs#L55-L56

`weights` (the raw conditional weights, as opposed to `combined`) is read only by the Gibbs sampler. On any run without `--numGibbsSamples` it is 8 bytes per CSR entry of pure dead weight, roughly 160 MB on a human-scale run. It should be populated conditionally.

## Fix

1. Build the packed structure once in the caller and use `optimize_packed_with_init`, which already exists and takes `&PackedEqClasses`.
2. Add `PackedEqClasses::refresh_combined(&mut self, eq: &CollapsedEqClasses)` that rewrites only the `combined` array in place after `update_eff_lengths`, leaving the rest untouched.
3. Gate population of `weights` on whether Gibbs will run (a flag on `from_collapsed`, or a separate constructor).

## Acceptance criteria

- [ ] A bias-corrected quant run constructs `PackedEqClasses` exactly once
- [ ] Same for the RAD requant path
- [ ] `weights` is not populated when Gibbs is not requested
- [ ] `quant.sf` and bootstrap output byte-identical before/after
- [ ] Peak RSS and inference wall-clock measured before/after

BenjaminDEMAILLE:
Part of #1092. Fixes #1087.

**Measured: peak RSS −1.39% (plain), throughput null.** The larger half of this PR's original rationale was wrong; corrected below.

### What it changes

`optimize` and `optimize_with_init` each construct a `PackedEqClasses` internally, so a bias-corrected run built the flat CSR layout twice and a run with bootstrap or Gibbs three times. The three drivers (reads, alignment, RAD) now build it once and call `optimize_packed_with_init` directly; `refresh_combined` patches the combined weights in place after bias correction, since that is the only array `update_eff_lengths` changes. It asserts the class set has not changed, which holds because nothing ever clears `TranscriptGroup::valid`.

Also gates the raw `weights` array, which only the Gibbs sampler reads: `from_collapsed_with(.., keep_weights)` lets the drivers skip populating it.

### Correction to the original rationale

I claimed avoiding the rebuilds would cut peak memory. **It cannot.** `optimize()` builds its layout, uses it, and drops it on return, so two copies never coexist. The repeated construction costs allocation and copying, not peak RSS. That was wrong and I only found it by measuring.

What the measurement does show is the `keep_weights` gate working, and nothing else:

| configuration | develop | this PR | delta |
| --- | --- | --- | --- |
| plain, `-p 8` | 1079 MB | 1064 MB | **−1.39%** |
| `--numGibbsSamples 4`, `-p 8` | 1072 MB | 1082 MB | +0.89% (noise) |

~15 MB, only when Gibbs is off — which is exactly right, since `weights` is 8 bytes per CSR incidence and Gibbs is the only consumer. With Gibbs requested it is populated in both builds and there is nothing to save.

Wall clock over the same runs: 34.33s vs 34.59s. **No throughput change**, consistent with every other measured item in #1092.

10M read pairs from ERR188044, GRCh38 cDNA release-110 (207,249 transcripts), 16-core arm64, interleaved with the warm-up round discarded.

### Why take it anyway

- Building the structure once instead of three times is simpler to follow than three call sites that each rebuild it implicitly, whatever it costs.
- `refresh_combined` states what actually changes after bias correction, where the previous code re-derived everything and left the reader to work out that only `combined` differs.
- The `keep_weights` gate is a real if modest saving, and it makes the Gibbs-only lifetime of that array explicit rather than implicit.

If the memory number is too small to be worth the diff, that is a fair call and I will not argue it.

### Parity

Byte-identical to develop on `sample_data`: reads mode plain / all three bias models / `--numBootstraps 8` / `--numGibbsSamples 8` (including `bootstraps.gz`, which is what exercises the `keep_weights` gate) / bias+bootstrap; RAD requant plain / bias / Gibbs; alignment mode plain and bias+Gibbs at `-p 1`.

Alignment mode is pinned to `-p 1` because it does not reproduce run to run at `-p > 1` on develop either (#1098). Per the methodology note on #1092, byte comparison above `-p 1` cannot establish parity here.

### DIFF
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -1782,9 +1782,16 @@ pub fn quantify_alignments(opts: &AlignQuantOptions) -> Result<AlignQuantResult>
     // The packed CSR layout is built once and reused by the EM, the post-bias
     // re-run and bootstrap / Gibbs (see the reads-mode driver for the rationale).
     // `weights` is Gibbs-only, so it is populated only when Gibbs will run.
-    let want_gibbs = !opts.skip_quant && opts.num_bootstraps == 0 && opts.num_gibbs_samples > 0;
+    // One resolution of which posterior method (if any) this run will use; the
+    // packed layout and the dispatch below both read it, so the precedence is
+    // stated once rather than restated at each site.
+    let posterior = salmon_infer::PosteriorMethod::resolve(
+        opts.skip_quant,
+        opts.num_bootstraps,
+        opts.num_gibbs_samples,
+    );
     let mut packed =
-        salmon_infer::PackedEqClasses::from_collapsed_with(&collapsed, num_refs, want_gibbs);
+        salmon_infer::PackedEqClasses::from_collapsed_for(&collapsed, num_refs, posterior);
     // `--initUniform` forces the plain uniform EM start; otherwise warm-start
     // from the online-estimate-blended init.
     let mut em = if opts.skip_quant {
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -63,7 +63,7 @@ use salmon_core::{
     ReadType,
 };
 use salmon_eqclass::{range_factorize_bins, EquivalenceClassBuilder, TranscriptGroup};
-use salmon_infer::{optimize, optimize_with_init, EmOptions};
+use salmon_infer::EmOptions;
 use salmon_model::FragmentLengthDistribution;
 
 const SALMON_VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -1779,6 +1779,12 @@ pub fn quantify_alignments(opts: &AlignQuantOptions) -> Result<AlignQuantResult>
             .map(|&p| p * frac_observed + uniform_prior * (1.0 - frac_observed))
             .collect()
     });
+    // The packed CSR layout is built once and reused by the EM, the post-bias
+    // re-run and bootstrap / Gibbs (see the reads-mode driver for the rationale).
+    // `weights` is Gibbs-only, so it is populated only when Gibbs will run.
+    let want_gibbs = !opts.skip_quant && opts.num_bootstraps == 0 && opts.num_gibbs_samples > 0;
+    let mut packed =
+        salmon_infer::PackedEqClasses::from_collapsed_with(&collapsed, num_refs, want_gibbs);
     // `--initUniform` forces the plain uniform EM start; otherwise warm-start
     // from the online-estimate-blended init.
     let mut em = if opts.skip_quant {
@@ -1791,12 +1797,12 @@ pub fn quantify_alignments(opts: &AlignQuantOptions) -> Result<AlignQuantResult>
             dropped_mass: 0.0,
         }
     } else if opts.init_uniform {
-        optimize(&collapsed, num_refs, &opts.em, Some(&eff_lengths))
+        salmon_infer::optimize_packed_with_init(&packed, &opts.em, true, None, Some(&eff_lengths))
     } else {
-        optimize_with_init(
-            &collapsed,
-            num_refs,
+        salmon_infer::optimize_packed_with_init(
+            &packed,
             &opts.em,
+            true,
             init_alphas.as_deref(),
             Some(&eff_lengths),
         )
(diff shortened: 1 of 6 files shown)


rob-p:
Reviewed. The change is sound — I verified the three claims it rests on rather than taking them, and pushed two follow-ups (4094a35, fb1a269).

## What I checked

**`TranscriptGroup::valid` is never cleared**, so `refresh_combined`'s assertion always holds. Only three writes exist and all are `valid: true` at construction. Your reasoning is right.

Worth knowing as a side effect: since it is never false, the `if !group.valid { continue; }` filter in `from_collapsed` never skips anything either. The whole `valid` concept looks vestigial — a branch per class and an assertion that cannot fire. Not this PR's business, but someone should probably delete it.

**`weights` is Gibbs-only.** The single read is `uncertainty.rs:222`, inside `gibbs_round`; `bootstrap` never touches it. So skipping the array on non-Gibbs runs is safe.

**The `num_bootstraps == 0` term in the gate is correct**, and this is the one that could have been a bug. It assumes bootstrap and Gibbs are mutually exclusive, which nothing in the CLI enforced — so I traced the dispatch: `lib.rs` is an `if/else if` chain where `num_bootstraps > 0` wins, meaning Gibbs never runs when both are set. Correct, but correct *at a distance*: the gate restates the dispatch chain's precedence, in three separate drivers.

## Follow-up 1: name the method rather than pass a bool (4094a35)

`keep_weights: bool` asks each caller to assert a conclusion that is only recoverable by already knowing which sampler reads which array, and the precedence it encodes was written out four times — three drivers plus the dispatch chain. If those ever drifted, the symptom would be Gibbs reading a `weights` array that was skipped at construction.

```rust
pub enum PosteriorMethod { None, Bootstrap { replicates: u32 }, Gibbs { samples: u32 } }
```

Resolved once; `from_collapsed_for(.., method)` replaces the bool; the dispatch `match`es on the same value the layout was built for, so choosing a sampler the layout was not built for is no longer expressible. `needs_raw_weights` states "only Gibbs reads them" in one place, next to the enum. `resolve` keeps the bootstrap-wins branch documented as unreachable-from-CLI, so the library does not depend on the CLI for its own consistency.

## Follow-up 2: make the exclusivity real (fb1a269)

`--numGibbsSamples`'s help text has always claimed "mutually exclusive with bootstraps". It was not enforced — passing both ran the bootstrap and discarded the Gibbs request silently. `conflicts_with` now rejects it.

**Behaviour change worth a release note:** a command line passing both previously ran and now fails. It was already not doing what it asked for, so failing is better, but it is not a silent improvement for anyone scripting it.

## Verification

Byte-identical to your version at `-p 1` on plain, `--numGibbsSamples 8` and `--numBootstraps 8`, including `bootstraps.gz`. I checked first that the three modes genuinely differ from each other (gibbs 975B, bootstrap 668B, no file for plain) so the parity result is not three copies of one run — that check has caught vacuous comparisons twice on this audit.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` and the full suite are clean.

## On #1098

You cite it as the reason alignment-mode parity is pinned to `-p 1`. That issue has since been closed as a non-issue wanting at most a doc change, so that part of the description is stale — though the `-p 1` choice itself is right, and for a more general reason: byte comparison cannot establish parity above one thread on any path, since develop is not run-to-run deterministic there either. See the measurement summary on #1092.

Happy to drop either follow-up if you would rather keep your version.

PR #1101, opened 2026-08-07
feat(align): optional in-memory intermediate for deterministic alignment mode

BenjaminDEMAILLE:
Follow-up to #1098. **Draft in the sense that I would like a read on whether the direction is wanted before polishing** — the mechanism is complete and verified, but it adds a flag, and you may prefer a different shape (or none).

### What it does

`-a --deterministic` currently writes an intermediate RAD, reads it back, then deletes it. The read side opens that file up to **five separate times** — the prelude probe, plus the FLD-derivation, bias, eq-class and fused passes, depending on what has to be derived. So the round trip is one full write and several full reads of a file that exists only to be consumed and unlinked.

`--radInMemory` keeps the image in memory instead. The bytes are byte-for-byte what would have gone to disk, so the reader cannot tell the difference and the output is unchanged.

### Why it is small

Both ends of `libradicl` were already generic: `RadFileWriter<W: Write + Seek>` (it needs `Seek` for the header backpatching at finalize) and `ParallelRadReader<R, T: BufRead + Seek>`. Only salmon pinned them to `File`. So this is mostly removing that pin.

Two deliberate design choices:

- **The writer gains an internal `Sink` enum, not a type parameter.** A generic `RadOutputWriter<W>` would leak through every worker that holds a `&RadOutputWriter` (`WriteWorker`, `ScoreWriteWorker`, `write_record`, and `genome_project.rs`), which is a lot of signature churn for a choice made once at construction. I tried the generic first; the enum is much less invasive.
- **The read side gains a `RadSource` (path or bytes), not a reader.** It has to *re-open* the image per pass rather than stream it once, so a single `impl BufRead` would not do.

### Scope

Off by default. Ignored when the RAD is itself a deliverable — `--writeRad`, `--keepRad`, `--skipQuant` all force the on-disk path — so the flag can never silently withhold something the user asked for.

Reads-mode `--deterministic` is untouched. There the second pass exists to avoid **re-mapping**, and the intermediate is what makes that affordable; the trade is completely different and I did not want to conflate them.

### Verification

On `sample_data`:

- `quant.sf` byte-identical between on-disk and in-memory paths
- byte-identical to unmodified `develop`'s `-a --deterministic`
- no file left behind in the in-memory case
- each of the three deliverable guards verified to force the disk path
- reads-mode `--deterministic` output unchanged
- four consecutive in-memory runs identical

Full workspace suite green, clippy clean, fmt clean.

### What I have not done

**No benchmark.** `sample_data` produces a 128 KB intermediate for 10k fragments, far too small to time meaningfully. The argument is structural — one write and up to five reads of an N-byte file traded for N bytes resident — and it scales with the intermediate, which is exactly why this is opt-in rather than a default. If you want numbers on a real BAM before considering it, that is fair and I will produce them.

**This does not address #1098.** Default alignment mode is still nondeterministic; this only makes the existing deterministic path cheaper. I mention it because the two are easy to conflate.

Happy to close this if the flag is not wanted — the underlying `Sink`/`RadSource` split might still be useful on its own, or not.

### DIFF
--- a/crates/salmon-align/src/bam_rad.rs
+++ b/crates/salmon-align/src/bam_rad.rs
@@ -76,11 +76,38 @@ pub struct AlignRadSummary {
 
 /// Project a name-grouped transcriptome BAM into a fully-baked salmon RAD at
 /// `rad_path`. Reference ids/names/lengths come from the BAM `@SQ` header.
+/// Where [`write_alignment_rad_to`] should put the RAD it produces.
+pub enum RadDest<'a> {
+    /// Write to this path.
+    Path(&'a Path),
+    /// Build the image in memory and hand it back.
+    Memory,
+}
+
+/// The outcome of a BAM-to-RAD pass: the counts, plus the image itself when it
+/// was built in memory.
+pub struct AlignRadOutput {
+    pub summary: AlignRadSummary,
+    /// `Some` only for [`RadDest::Memory`].
+    pub bytes: Option<Vec<u8>>,
+}
+
+/// Write the alignment RAD to a file. Thin wrapper over
+/// [`write_alignment_rad_to`] for the common case.
 pub fn write_alignment_rad(
     opts: &AlignQuantOptions,
     rad_path: &Path,
     codec: ChunkCodec,
 ) -> Result<AlignRadSummary> {
+    Ok(write_alignment_rad_to(opts, &RadDest::Path(rad_path), codec)?.summary)
+}
+
+/// As [`write_alignment_rad`], but the destination is chosen by the caller.
+pub fn write_alignment_rad_to(
+    opts: &AlignQuantOptions,
+    dest: &RadDest,
+    codec: ChunkCodec,
+) -> Result<AlignRadOutput> {
     let header = read_alignment_header(&opts.bam)?;
     if coordinate_sorted_unusable(&header) {
         bail!(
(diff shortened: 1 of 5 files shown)


rob-p:
I don't think it's the case that there are many *full* reads of the rad file. Prelude probe and baked FLD just do small reads. The real question with this feature would be if any perf gains justify the complexity or near duplicate code cost
 Otherwise, rad inputs the second pass in read and alignment based modes.so I think there'd have to be a pretty big perf win to justify this.

In alignment mode with deterministic, there should actually be only one full read of the RAD file.

BenjaminDEMAILLE:
You are right, and I checked rather than argued: **there is exactly one full pass**, and my "up to five full reads" was wrong.

I instrumented `open_prelude` and each of the four pass functions and counted:

| configuration | `open_prelude` calls | full record passes |
| --- | --- | --- |
| `-a --deterministic` | 2 (one is the header probe) | **1** (`run_rad_pass`) |
| `-a --deterministic --seqBias --gcBias` | 2 | **1** (`run_rad_eq_and_bias_pass`, fused) |
| plain `--rad` requant | 2 | **1** |

I had counted the five `open_prelude` *call sites* in `rad.rs` and reported them as five reads. They are mutually exclusive branches — the FLD-derivation pass only runs when the FLD is not baked, and `-a --deterministic` bakes it; the bias pass and the fused pass are alternatives, not a sequence. Never more than one is taken. That was a careless reading and it is the entire load-bearing claim in the PR description.

## What the saving actually is

One write plus one read of the intermediate. And it is smaller than I assumed: `sample_data` produces 128,224 bytes for 10,000 fragments, i.e. **~12.8 bytes per fragment**, so

- 10M fragments → ~128 MB written, ~128 MB read back
- 100M fragments → ~1.3 GB each way

The file is also written and then immediately read, so the read is served from page cache and costs little beyond a memcpy — the real saving is closer to one write than to a round trip.

Against that: three new public types (`Sink`, `RadSource`, `RadDest`, plus `AlignRadOutput`), a new flag, a `finalize_into_bytes` that can fail at runtime on a file-backed writer, and a second construction path through the writer. That is a poor trade for saving ~128 MB of buffered write on a run that takes minutes, and your bar — "a pretty big perf win to justify this" — is not met. I have no benchmark showing one, and after this measurement I do not expect to find one.

**Closing.** No part of it is worth keeping on its own: the `Sink`/`RadSource` split exists only to serve this feature.

Thanks for pushing back on the premise rather than the diff. I had built the whole thing on a number I never verified, and it took you one sentence to spot it.

Issue #1102, opened 2026-08-08
Compression support is inconsistent across input paths: index -t accepts zstd, quant -1/-2 does not


BenjaminDEMAILLE:
`salmon index -t transcripts.fa.zst` works. `salmon quant -1 reads_1.fq.zst` does not. Four input paths in the codebase decide compression four different ways, and two of them reject input the other two accept.

## Measured behaviour

Tested on `develop` (`ab6870d`) with `sample_data`, each input in four forms: uncompressed, `.gz`, gzip content deliberately named `.gzip` (to separate name-sniffing from content-sniffing), and zstd.

| input | plain | `.gz` | gzip named `.gzip` | `.zst` |
| --- | --- | --- | --- | --- |
| `salmon index -t` | ✅ | ✅ | ✅ | ✅ |
| `--geneMap` (GTF/GFF) | ✅ | ✅ | ✅ | ✅ |
| `quant -a -t` (targets FASTA) | ✅ | ✅ | ✅ | ❌ |
| `quant -1/-2` (FASTQ reads) | ✅ | ✅ | ❌ | ❌ |

## Why they differ

Four independent implementations:

- **`salmon index -t`** — `needletail::parse_fastx_file`, which sniffs content. Everything works.
- **`--geneMap`** — `niffler::get_reader`, which sniffs magic bytes and covers gzip/BGZF/bzip2/xz/zstd. Everything works.
- **`quant -a -t`** — hand-rolled two-byte magic check for gzip only:
  ```rust
  f.read_exact(&mut magic).is_ok() && magic == [0x1f, 0x8b]
  ```
  Content-sniffed, so the `.gzip` name is fine, but gzip is the only format it knows.
- **`quant -1/-2`** — extension match, content never inspected:
  ```rust
  if path.extension().and_then(|e| e.to_str()) == Some("gz") { ... }
  ```
  (`open_reader`, `crates/salmon-quant/src/lib.rs`). `open_sam_reader` in `salmon-align` uses the same `.ends_with(".gz")` approach.

So the *most* important input path — the reads — has the *weakest* detection of the four.

## The failure mode

It errors rather than silently mis-quantifying, which is the important thing. But the message does not say what is wrong:

```
$ salmon quant ... -1 reads_1.fastq.zst -2 reads_2.fastq.zst
Error: quantification failed
Caused by:
    opening reads_1.fastq.zst: Invalid start character ((), expected either '>' or '@'
```

```
[...]

PR #1103, opened 2026-08-08
feat(io): detect input compression from content, shared across input paths

BenjaminDEMAILLE:
Fixes the non-FASTQ half of #1102.

### The problem

salmon decided "is this compressed, and how" independently at four entry points, with four different answers:

| path | mechanism | accepted |
| --- | --- | --- |
| `salmon index -t` | `needletail`, sniffs content | gzip, bzip2, xz, zstd |
| `--geneMap` | `niffler`, sniffs magic bytes | gzip, bzip2, xz, zstd |
| `quant -a -t` | hand-rolled 2-byte gzip magic check | gzip only |
| `quant -1/-2` | file-extension match on `.gz` | `.gz`-named gzip only |

So `salmon index -t transcripts.fa.zst` worked while `quant -a -t transcripts.fa.zst` did not, and a perfectly ordinary gzip file that was not named `.gz` was rejected with

```
Invalid start character (), expected either '>' or '@'
```

which describes the decompressed format that never happened, rather than the compression that was not detected.

### The change

Adds `salmon_core::compress`, a single helper over `niffler` — already a workspace dependency and already used by `--geneMap` — and routes the transcriptome FASTA (`load_ref_bytes`) and the SAM reader through it. Detection is by magic bytes, so a correctly compressed file is read whatever it is called, and gzip goes through `MultiGzDecoder` so multi-member and BGZF files are read in full rather than truncated at the first member.

`genemap` keeps its path-annotated error messages but now delegates the decoding, so there is one implementation rather than two that happen to agree.

`is_sam_path` still decides SAM-vs-BAM by extension. That is a different question from how the bytes are compressed, and it stays as it was.

### The FASTQ path is deliberately excluded

I had it implemented and dropped it: per your answer on #1102, `open_reader` will be replaced by piscem 0.9's decoder selection with auto-balanced mapping/decompression threads. No point landing something that is about to be replaced.

One thing worth checking when 0.9 lands, phrased as a question rather than a request: is its decoder selection format-general, or gzip-centric? If the latter, `-1/-2` would still reject zstd/bz2/xz while every other salmon input accepts them, and the helper added here is the piece that would compose with it. Entirely your call — I am not proposing anything for that path now.

### Verification

`quant -a -t`, all four forms of the same transcriptome, at `-p 1`:

| form | result |
| --- | --- |
| `t.fa` | ✅ |
| `t.fa.gz` | ✅ byte-identical `quant.sf` |
| gzip named `t.fa.gzip` | ✅ byte-identical |
| `t.fa.zst` | ✅ byte-identical |

and byte-identical to develop's own plain `-p 1` run, so the decoding change alters nothing else.

`--geneMap` still accepts plain / gz / zstd after being rewired onto the shared helper. The reads path is confirmed unchanged (zstd still rejected there, as intended).

Comparisons are at `-p 1` deliberately: alignment mode is not run-to-run reproducible at `-p > 1` (#1098), so a multi-threaded comparison here would measure that rather than this change.

Three unit tests on the helper cover content-sniffing regardless of file name, multi-member gzip being read to the end, and a file too short to sniff opening as plain text. Full suite green, clippy clean, fmt clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2684,11 +2684,14 @@ dependencies = [
 name = "salmon-core"
 version = "2.4.1"
 dependencies = [
+ "flate2",
  "niffler",
  "serde",
  "serde_json",
+ "tempfile",
  "thiserror 2.0.19",
  "tracing",
+ "zstd",
 ]
 
 [[package]]
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -517,25 +517,17 @@ fn base_2bit(b: u8) -> u8 {
     }
 }
 
-/// Load a (optionally gzip'd) transcriptome FASTA and return the (ASCII) base
-/// sequences aligned to the BAM's reference order (`names`); a name absent from
-/// the FASTA yields an empty sequence (its model contributions are skipped). The
-/// same bytes feed both the error model (2-bit on the fly) and the bias models.
+/// Load a (optionally compressed) transcriptome FASTA and return the (ASCII)
+/// base sequences aligned to the BAM's reference order (`names`); a name absent
+/// from the FASTA yields an empty sequence (its model contributions are
+/// skipped). The same bytes feed both the error model (2-bit on the fly) and the
+/// bias models.
+///
+/// Compression is detected from content, so gzip/BGZF, bzip2, xz and zstd all
+/// work regardless of how the file is named.
 fn load_ref_bytes(path: &Path, names: &[String]) -> Result<Vec<Vec<u8>>> {
-    let file = std::fs::File::open(path).with_context(|| format!("opening {}", path.display()))?;
-    let mut magic = [0u8; 2];
-    let is_gz = {
-        use std::io::Read;
-        let mut f = std::fs::File::open(path)?;
-        f.read_exact(&mut magic).is_ok() && magic == [0x1f, 0x8b]
-    };
-    let reader: Box<dyn BufRead> = if is_gz {
-        Box::new(std::io::BufReader::new(flate2::read::MultiGzDecoder::new(
-            file,
-        )))
-    } else {
-        Box::new(std::io::BufReader::new(file))
-    };
+    let reader: Box<dyn BufRead> = salmon_core::compress::open_maybe_compressed(path)
+        .with_context(|| format!("opening {}", path.display()))?;
 
     let mut by_name: HashMap<String, Vec<u8>> = HashMap::new();
     let mut cur_name: Option<String> = None;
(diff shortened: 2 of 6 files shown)


rob-p:
Reviewed, verified independently, and pushed two fixes to the branch. Recommending merge.

## Verified against the description

| claim | result |
| --- | --- |
| `quant -a -t` takes plain / `.gz` / gzip-named-`.gzip` / `.zst` | ✅ all four exit 0 |
| all four give identical output | ✅ same md5 |
| decoding changes nothing else | ✅ **byte-identical to develop's** plain run |
| `--geneMap` still works after being rewired | ✅ plain / gz / zst |
| FASTQ path deliberately untouched | ✅ zstd still rejected there |
| multi-member gzip read in full | ✅ (would fail under plain `GzDecoder`) |

Also confirmed niffler 3's default features really do include zstd/bz2/lzma, so the format support is not conditional on something nobody enables.

## Fixed: `.sam.zst` went to the BAM reader

The doc said a compressed SAM "is read whatever it is called", but dispatch happens before decoding, and `is_sam_path` matched only `.sam` / `.sam.gz`:

```
a.sam.zst  → is_sam_path == false → BAM reader → "invalid BGZF header"
b.sam      (identical zstd bytes) → SAM reader → works
```

So the decoding was right and the routing was not — and the resulting message is exactly the genre this PR set out to remove: a complaint about a format the file never had.

`is_sam_path` now strips one compression suffix before asking. SAM-vs-BAM is a question about the *name*; how the bytes are compressed is a question about the *content*; keeping them separate is what makes both answerable. The helper sits in `salmon_core::compress` next to the decoder, documented so it is clear this is not name-based compression detection sneaking back in — it removes compression noise from a name so a different question can be asked of what remains.

All five forms now agree, and BAM is unaffected:

| input | result |
| --- | --- |
| `a.sam` | ✅ |
| `a.sam.gz` | ✅ byte-identical |
| `a.sam.zst` | ✅ byte-identical |
| `a.sam.bz2` | ✅ byte-identical |
| `a.sam.xz` | ✅ byte-identical |
| `sample_alignments.bam` | ✅ unchanged |

## Dropped `IntoBoxedBufRead`

Its doc said it existed so `open_maybe_compressed` could name its return type without a turbofish, but `Box::new(BufReader::new(..))` coerces fine against the declared return type. Checked by making the change and running the tests before deleting the trait — ten lines of indirection for nothing.

## Noted, deliberately not fixed here

Two pre-existing warts this PR does not introduce, so they are not blockers. Both go in a follow-up so this can land cleanly:

- a corrupt compressed file reports `unexpected end of file` **without naming the path** — I confirmed develop behaves identically
- an **empty `-t` FASTA exits 0** and quantifies with the error and bias models silently disabled

Issue #1105, opened 2026-08-08
bug(rad): a failed --writeRad leaves an unmarked partial file; readers cannot tell it is incomplete (RELEASE BLOCKER)


rob-p:
**Release blocker**, and the producer-side half of #1104.

When a `--writeRad` run fails partway through — full disk, killed job, crash — salmon leaves the partial `.rad` on disk with nothing to distinguish it from a finished one at a glance. The next `--rad` run then reads it and hangs forever (#1104).

Original incident: a `--writeRad` run hit `ENOSPC`. It produced no `meta_info.json`, so the run clearly failed, but a 470 MB `.rad` was left behind (a complete one is ~564 MB). Feeding that to `--rad` deadlocked with no diagnostic.

## The signal already exists, and is not used

`num_chunks` is backpatched at `finalize`, so the header does distinguish the states:

| file state | header `num_chunks` | chunks actually present |
| --- | --- | --- |
| finalized | true count (e.g. 611) | 611 |
| finalized, then truncated | **611** | fewer than 611 |
| never finalized | placeholder | some |

Both failure modes are therefore detectable from data already in the file. The reader just never checks: with a real count it does not notice hitting EOF early, and with the placeholder it falls back to read-until-EOF and cannot distinguish "incomplete" from "streaming, count unknown".

So the reader fix in #1104 (release the consumers so the error can propagate) should be paired with an explicit up-front check, which turns a mid-parse failure into a clear message before any work starts.

## Suggested fixes

> **Note:** the original item 1 below (pre-parse chunk-count validation) is
> **withdrawn** — `num_chunks == 0` is a valid RAD signal for a dynamic/unknown
> chunk count, used by streaming producers that cannot backpatch. See the
> correction comment. The corrected plan is:
>
> - **Reader (#1104):** release the consumers on the producer's error path so a
>   parse error propagates. Correct for truncated input from *any* producer. No
>   count validation.
> - **Writer:** report the failed write clearly, and write to a temporary name,
>   renaming on successful finalize so a partial file never appears under the
>   user's chosen name. The error should name the partial file explicitly.
> - **Completeness marker:** set a file-level tag at finalize, as a
>   *confirmatory* fast path for salmon-produced files only. Its absence proves
>   nothing, so it must never be required.

**2. Do not leave a partial file behind (writer side).**

[...]

PR #1108, opened 2026-08-08
feat(rad): record what the mapping pass observed, so a requant reproduces it

rob-p:
Follow-up to the three-way counter run in #1105. **Stacked on #1106** — it touches the same writer, so review that one first; the diff here will shrink once it merges.

## The problem

A RAD holds only the fragments that *mapped*, so nothing in the records can say how many were *seen*. Measured on 10M human fragments:

| | num_processed | num_mapped | percent_mapped |
|---|---|---|---|
| online (reads) | 10,000,000 | 9,221,625 | 92.22% |
| `--deterministic --writeRad` | 9,221,625 | 9,221,625 | **100.00%** |
| `--rad` input | 9,221,625 | 9,221,625 | **100.00%** |

`--deterministic` is the worse of the two, because it *knew*:

```
deterministic: mapped 9221625 / 10000000 fragments (92.22%); quantifying from intermediate RAD
...
9221625 fragments from intermediate RAD, 9221625 quantified (100.00%)
```

It computes the true rate, logs it, then quantifies through the intermediate RAD and lets that pass's counters overwrite it.

The structural cause is a second, cut-down `MetaInfo` in `salmon-align`, written for `-a` BAM mode where "no index hashes since there is no index" is true — then reused by the `--rad` path, where it isn't. Hence empty index hashes and absent `keep_duplicates` / `num_orphan`.

## What is recorded

Two groups, by when they become known. Index identity and `mapping_type` are known before the first record, so they are written outright; the five counters are only final at end of pass, so their slots are reserved and backpatched at finalize, exactly like the FLD.

| tag | why it cannot be derived |
| --- | --- |
| `num_processed` | the root cause; every mapping-rate figure depends on it |
| `num_dovetail_fragments` | a property of the pass, not of its output |
| `num_fragments_filtered_vm` | filtered fragments were never written |
| `num_alignments_below_threshold_vm` | same |
| `num_decoy_fragments` | decoys are filtered before write |
| `index_*_hash`, `keep_duplicates` | the `--rad` path loads no index at all |
| `mapping_type` | not inferable from the profile (below) |

Two `baked_flags` bits (`BAKED_MAP_COUNTERS`, `BAKED_INDEX_PROV`) say the values are real — a reserved slot and a genuine count of zero are otherwise indistinguishable. A test caught that I had defined the second bit but never set it; since the reader gates on it, the index tags would have been written and then silently ignored.

`mapping_type` is recorded rather than inferred from the record profile because the two disagree: a BAM-derived RAD is written in the selective-alignment profile although its fragments came from an aligner.

`num_orphan` is deliberately **not** baked. Fragment type is in the records (`LEFT_ORPHAN`/`RIGHT_ORPHAN` is exactly the mapping pass's `PairedEndLeft|PairedEndRight`), so it is counted on read and works for a piscem RAD too. That also fixes `lib_format_counts.json`, which the RAD path never wrote at all and which `--deterministic` filled with every mapped fragment counted as concordant — `record_discrete` never tallied orphans.

## When the RAD does not carry it

The run proceeds — refusing to quantify a piscem RAD over reporting metadata would be a poor trade. It warns, naming the affected fields, and `meta_info.json` records the same thing durably:

```json
"meta_info_complete": false,
"incomplete_meta_info_fields": ["num_processed", "percent_mapped", ...]
```

so a result read long after the run still says which of its numbers are placeholders. The reads path always reports `true`.

## Verification

On 10M human fragments, against a RAD written by this build:

- `--rad` requant vs the online mapping run: **identical on all 40 comparable fields**
- `--rad` vs `--deterministic`: **identical on all 45** (run identity excluded)
- `percent_mapped` 92.21625 (was 100.0); `num_processed` 10,000,000 (was 9,221,625); index hashes, `mapping_type: mapping`, `num_orphan: 623154` all recovered
- `lib_format_counts.json` now agrees across all three: concordant 8,598,471 / orphan 623,154 / format IU

The five remaining online-vs-requant differences are genuine online/deterministic inference differences (FLD mean/sd/source, eq-class count, EM iterations), not lost data.

Control: quantifying a RAD written *before* this change warns as designed, sets `meta_info_complete: false`, lists all 13 fields — and still recovers `num_orphan` and writes `lib_format_counts.json`, since those come from the records.

`-a` BAM output is unchanged: the recovered fields are `skip_serializing_if` and stay absent rather than being guessed.

`cargo fmt`, `clippy --workspace --all-targets -D warnings`, and the full suite are clean; four new round-trip tests in `salmon-rad`.

## Noted, not fixed here

`keep_duplicates` is hardcoded `false` on **every** path, including reads mode, because `info.json` does not persist it. This carries through faithfully whatever the mapping run reported; making it truthful is an index-format change and wants its own issue.

### DIFF
--- a/crates/salmon-align/src/bam_rad.rs
+++ b/crates/salmon-align/src/bam_rad.rs
@@ -146,6 +146,12 @@ pub fn write_alignment_rad(
         RadProfile::SelectiveAlignment,
         opts.fld_max + 1,
         codec,
+        // Fragments came from an external aligner, and there is no salmon index
+        // behind them; a reader reads the absent index as "unknown".
+        &salmon_rad::WriterProvenance {
+            mapping_type: salmon_rad::MappingType::Alignment,
+            index: None,
+        },
     )?;
 
     let fld = DiscreteFld::new(opts.fld_max);
--- a/crates/salmon-align/src/bam_rad.rs
+++ b/crates/salmon-align/src/bam_rad.rs
@@ -151,6 +151,9 @@ pub fn write_alignment_rad(
         &salmon_rad::WriterProvenance {
             mapping_type: salmon_rad::MappingType::Alignment,
             index: None,
+            // Carry the aligner's own account of how these alignments were made,
+            // so a requant reports the command line and not merely "a BAM".
+            source_programs: crate::source_program_lines(&header),
         },
     )?;
 
@@ -214,6 +217,13 @@ pub fn write_alignment_rad(
         writer.set_initial_abundances(&em.alphas);
     }
 
+    // The BAM's own fragment total, which the records cannot show once unmapped
+    // ones are dropped. The `_vm` counters and the decoy count have no meaning
+    // for an aligner's output and stay zero.
+    writer.set_map_counters(salmon_rad::MapCounters {
+        num_processed: summary.num_processed,
+        ..Default::default()
+    });
     writer.finalize()?;
     Ok(summary)
 }
--- a/crates/salmon-align/src/genome_project.rs
+++ b/crates/salmon-align/src/genome_project.rs
@@ -160,6 +160,9 @@ pub fn project_genome_bam_to_rad(
         &salmon_rad::WriterProvenance {
             mapping_type: salmon_rad::MappingType::Alignment,
             index: None,
+            // Carry the aligner's own account of how these alignments were made,
+            // so a requant reports the command line and not merely "a BAM".
+            source_programs: crate::source_program_lines(&header),
         },
     )?;
 
--- a/crates/salmon-align/src/genome_project.rs
+++ b/crates/salmon-align/src/genome_project.rs
@@ -155,6 +155,12 @@ pub fn project_genome_bam_to_rad(
         RadProfile::SelectiveAlignment,
         opts.fld_max + 1,
         opts.rad_codec,
+        // Fragments came from an external aligner, and there is no salmon index
+        // behind them; a reader reads the absent index as "unknown".
+        &salmon_rad::WriterProvenance {
+            mapping_type: salmon_rad::MappingType::Alignment,
+            index: None,
+        },
     )?;
 
     // Short-read projection only. Long-read genome quantification is out of scope
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -271,6 +271,15 @@ impl AlignQuantOptions {
 }
 
 /// Quantification results.
+/// What a quantification run read its fragments from.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum FragmentSource {
+    /// an alignment file given with `-a`
+    Bam,
+    /// a RAD file given with `--rad`, or written as a `--deterministic` intermediate
+    Rad,
+}
+
 #[derive(Debug, Clone)]
 pub struct AlignQuantResult {
     pub names: Vec<String>,
@@ -290,6 +299,13 @@ pub struct AlignQuantResult {
     /// what the RAD says about the mapping pass that produced it; empty for a
     /// BAM input, which has no RAD provenance to read
     pub provenance: crate::rad::RadProvenance,
+    /// verbatim `@PG` lines describing how the fragments were produced: read
+    /// from a BAM's header directly, or carried through a BAM-derived RAD
+    pub source_programs: Vec<String>,
+    /// what this run actually read, which decides which metadata is even
+    /// applicable — an index hash is missing from a RAD requant but simply does
+    /// not apply to a BAM, and blaming a RAD that was never read would be wrong
+    pub source: FragmentSource,
     /// fragment mass dropped in the final min-alpha redistribution (every member
     /// of an equivalence class truncated); normally 0.
     pub inference_truncated_mass: f64,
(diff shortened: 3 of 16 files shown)


rob-p:
Pushed `e5a1ae0b`, covering three follow-ups: propagating `keep_duplicates`, recording *known* gaps rather than only silently omitting them, and carrying the BAM's `@PG` chain.

### `keep_duplicates`

It was never knowable by anyone. `--keepDuplicates` is a build-time option that `info.json` did not persist, so **every** path — including read-based runs — reported a hardcoded `false` no matter how the index was built. The index now records it, and it flows index → reads run → RAD → requant.

Stored as `Option<bool>` rather than a `#[serde(default)] bool`, because an index built before this must read as *unknown*: a defaulted `false` is indistinguishable from an observed one, and emitting it would state as fact something nobody measured. Same on the RAD side — the tag is written only when known, so an absent tag means unknown, not "duplicates were collapsed".

| | `keep_duplicates` | `meta_info_complete` |
|---|---|---|
| reads, index built with `--keepDuplicates` | `true` | `true` |
| requant of that run's RAD | `true` | `true` |
| reads, pre-2.5 index | *absent* | `false`, gap named |

### Known gaps are recorded

`meta_info.json` always carries `meta_info_complete`; when false, `incomplete_meta_info_fields` names each field, the upstream that could not supply it, and why:

```json
{"field": "keep_duplicates", "source": "index",
 "reason": "this index predates recording how it was built; rebuild the index to record it"}
```

`source` is `index`, `rad` or `bam`. This deliberately applies to read-based runs too — a fresh mapping run against an old index is now honest about the one thing it cannot know instead of quietly emitting a guess.

Fields that are **not applicable** are excluded rather than listed. Alignment mode has no salmon index, so its absent index hashes are by design, not a gap. Getting that right needed an explicit `FragmentSource`, because a direct `-a` run was otherwise blaming a RAD it had never read for thirteen fields — visible in testing as `complete: false` with `source: rad` on a run with no RAD anywhere in it.

### Alignment provenance (`@PG`)

`-a` now reads the BAM's `@PG` chain and reports it as `alignment_provenance`, so a result says *how* its alignments were made:

```json
"alignment_provenance": [
  {"id": "bowtie2", "program_name": "bowtie2", "version": "2.3.5.1",
   "command_line": "\"...bowtie2-align-s --wrapper basic-0 -k15 -X1000 -x bt2_sample_idx --no-discordant --no-mixed -1 ../sample_data/reads_1.fastq -2 ../sample_data/reads_2.fastq\""}
]
```

A BAM-derived RAD carries the `@PG` lines **verbatim** (stored as raw SAM text, not a parsed structure, so nothing the aligner recorded is dropped in transit) and a requant reports the same chain. The BAM→RAD writer also now bakes its own fragment total, which the records cannot show once unmapped fragments are dropped — so `BAM → RAD → requant` reports `meta_info_complete: true` with the full chain intact.

A BAM whose header carries no `@PG` records is itself reported, with `source: bam` — the gap is upstream, and saying so beats an empty field the reader has to interpret.

### Verification

10M human fragments, index rebuilt so it records the flag: online, `--deterministic` and `--rad` **all** report `meta_info_complete: true`, and the requant still matches the mapping run on all 40 comparable fields and `--deterministic` on all 45.

`cargo fmt`, `clippy --workspace --all-targets -D warnings` and the full suite are clean. New tests cover the three-state `keep_duplicates` (`Some(true)` / `Some(false)` / absent — asserting the unknown case writes *no* tag) and the `@PG` round trip, including a command line containing spaces and its own colons.

### Still not fixed

`num_orphan` is absent on the direct `-a` path — that path doesn't tally orphans. I left it out of `incomplete_meta_info_fields` because the mechanism names *upstream* gaps and this one is ours, not the BAM's. Easy to add if you'd rather it were tallied.

rob-p:
Added in `f104191d`: the direct `-a` path now tallies orphans, closing the last metadata gap in this PR that was ours rather than an upstream's. It reported no `num_orphan` and wrote no `lib_format_counts.json` simply because it never counted them.

**Where the count comes from.** `proper` is already computed per surviving placement for the fragment-geometry logic, so the added work is a bool OR inside a loop that was running anyway, plus one branch per mapped fragment. Tallies are plain per-worker `u64`s summed once at join — nothing atomic touches the hot path. `process_fragment` now returns a `FragmentOutcome` instead of a bool, since the caller needs three cases and a pair of bools would make "unmapped but orphan" expressible.

**Definition.** It matches the RAD fragment-level rule, which takes the *most complete* status across hits: a fragment that is a proper pair on one transcript and an orphan on another counts as a pair. So it is `paired_lib && no surviving placement paired both mates`. The `paired_lib` gate is what stops single-end libraries — where there is no mate to be missing — from being counted as entirely orphaned.

**Verified against constructed ground truth**, not a self-consistent run. Stripping the read2 record from every fifth query of a 10,000-query SAM:

| | mapped | num_orphan | concordant | orphan |
| BAM direct | 10,000 | **2,000** | 8,000 | 2,000 |
| BAM → RAD → requant | 10,000 | **2,000** | 8,000 | 2,000 |

Exactly the 2,000 injected, `concordant + orphan == num_mapped`, and the direct path agrees with the RAD-mediated one.

**Overhead measured, not asserted.** Interleaved A/B against the immediately preceding commit, `-a` on a 239 MB BAM at `-p 8`:

| n | delta | t |
| --- | --- | --- |
| 5 | +3.71% | 1.54 |
| 15 | **+1.08%** | **0.85** |

95% CI −0.057 … +0.142 s on a ~3.9 s run, so it spans zero. The n=5 reading shrinking toward null as n grew is the same pattern that produced phantom results earlier in this project, which is why I did not stop at the first number. No detectable cost; I would not claim it is provably zero, only that it is below what this benchmark can resolve.

PR #1112, opened 2026-08-08
fix(infer): widen CSR offsets to u64 and drop the incidence ceiling

### LINKED ISSUE #1097: infer: choose the packed CSR offset width from the incidence count, before the next release
Follow-up to #1088, which #1094 resolved with an assertion. **This should be settled before the next release.**

## Where things stand

`PackedEqClasses` stores CSR offsets as `u32` and filled them with an unchecked `labels.len() as u32`. Past `u32::MAX` cumulative `(class, transcript)` incidences that wrapped silently, and `class()` then handed the optimizer slices belonging to other classes — wrong abundances, no panic, no warning.

#1094 added an assertion so the limit fails loudly rather than producing plausible nonsense. That is the correct minimum and it removes the silent-corruption failure mode, but it converts a wrong answer into a refusal to run: an input above the ceiling still cannot be quantified.

## What we want instead

Not a blanket widening to `u64`. That doubles the offset array for every run in order to serve the rare one, and `starts` is one entry per class — real memory on a large index, spent almost always for nothing.

The incidence count is known before the arrays are filled — #1094 already computes `total_labels` in a cheap pre-pass to size them. So the width can be *chosen* at that point:

- `u32` offsets when `total_labels <= u32::MAX`, which is every normal run
- `u64` offsets only when it does not

Small runs keep the smaller offsets and better cache behaviour; large runs simply work. No ceiling and no cost to the common case.

## Design questions to settle

- **Where the branch lives.** An enum over the two offset vectors, a generic parameter on `PackedEqClasses`, or two concrete types behind a trait. The EM/VBEM inner loops index `starts` on the hot path, so whatever is chosen must not put a match inside them — most likely monomorphisation over an offset type rather than a runtime enum, but that should be measured, not assumed.
- **Blast radius.** `PackedEqClasses` is consumed by `optimize`, `optimize_with_init`, the bootstrap/Gibbs path and the RAD path. A generic parameter propagates; an enum does not but costs a branch.
- **Interaction with #1087.** That issue proposes building the packed layout once per run instead of two to four times. Both change `from_collapsed`'s construction, so whichever lands first should be written with the other in mind.
- **Testability.** The `u64` path is unreachable on any test-sized input, so it needs a unit test that constructs the CSR directly at a synthetic offset scale rather than going through a real quant.

## Acceptance criteria

- [ ] Offset width is selected from the actual incidence count, not fixed
- [ ] Inputs above `u32::MAX` incidences quantify correctly rather than asserting
- [ ] No measurable regression in EM iteration time for the `u32` case, reported rather than asserted
- [ ] A test exercises the `u64` path directly
- [ ] The assertion added in #1094 is removed, having become unreachable

Filed from the review of #1094; see that discussion for the reasoning behind accepting the assertion as an interim step.

rob-p:
Closes #1097.

## What is at risk, and what is not

`starts` holds cumulative *incidence* counts, so it grows with the whole dataset. Everything else in `PackedEqClasses` is safe: `labels` holds transcript ids (bounded by `num_txps`), `combined`/`weights` are already `f64`, `counts` is already `u64`, and the remaining `u32`s in the crate are user-specified iteration and sample counts. I checked the rest of the workspace for the same cumulative-cast shape — `starts.push(labels.len() as u32)` was the only instance.

So the fix is one field, one entry per class.

## Why widen rather than select the width

#1097 planned to choose `u32` or `u64` from the incidence count, rejecting a blanket widening as "real memory on a large index". The numbers do not support that:

| eq classes | `starts` u32 | u64 | delta | delta as % of packed |
| 411,419 (our human run) | 1.6 MB | 3.3 MB | 1.6 MB | **3.45%** |
| 1,000,000 | 4.0 MB | 8.0 MB | 4.0 MB | 3.45% |
| 10,000,000 | 40 MB | 80 MB | 40 MB | 3.45% |
| 100,000,000 | 400 MB | 800 MB | 400 MB | 3.45% |

A flat ~3.5% of the packed structure at every scale — 0.155% of peak RSS on the human run. And the regime the narrow width protects is unreachable: `u32` overflows past 4.29 B incidences, ~859 M classes at k=5, whose `starts` array is **already 3.4 GB as `u32`**, with ~86 GB of labels and weights beside it.

Against that, the selection machinery is an enum or a generic parameter — the latter propagating through `optimize`, `optimize_with_init`, bootstrap/Gibbs and the RAD path, as #1097 itself notes. A one-word type change does the same job.

## The cost, measured

Interleaved A/B of the EM phase on the 10M-fragment human RAD, warm-up discarded. The point estimate drifts down as rounds accrue and the sign is barely better than a coin flip:

| n | delta | t |
| --- | --- | --- |
| 11 | +1.62% | 1.80 |
| 20 | +1.40% | 2.07 |
| 27 | +1.10% | 2.06 |
| **29 (all rounds)** | **+0.95%** | **1.86** |

At the full sample: `u32` 11.782 s (sd 0.341), `u64` 11.894 s (sd 0.289), 95% CI `-0.009 .. +0.232 s` — spanning zero. `u64` is slower in **16 of 29 rounds**, i.e. 55%.

So: a point estimate near +1% that this benchmark cannot separate from noise. Not a demonstrated regression, and not a demonstrated absence of one either — if there is a real effect it is below what ~30 rounds at this variance can resolve. The plausible mechanism (doubling an array read once per class per iteration) argues for taking the ~1% as an upper bound rather than assuming zero.

`quant.sf` is byte-identical between the two builds.

## Acceptance criteria

- [x] Inputs above `u32::MAX` incidences quantify rather than asserting
- [x] The #1094 assertion is removed
- [x] EM-time effect reported rather than asserted (above)
- [x] `quant.sf` byte-identical between the two builds
- [n/a] "Offset width selected from the incidence count" and "a test exercises the `u64` path" — both moot once there is only one path; the `u64` path is now the only one, so every existing test exercises it

### DIFF
--- a/crates/salmon-infer/src/packed.rs
+++ b/crates/salmon-infer/src/packed.rs
@@ -49,7 +49,14 @@ pub struct PackedEqClasses {
     ///
     /// The extra trailing entry is what lets class `i`'s span be written without
     /// a special case for the last class.
-    pub starts: Vec<u32>,
+    ///
+    /// `u64` rather than `u32` because these are cumulative *incidence* counts,
+    /// not transcript ids: they grow with the whole dataset, and past
+    /// `u32::MAX` a narrower offset wrapped silently and handed the optimizer
+    /// slices belonging to other classes. One entry per class, so the width
+    /// costs ~3.5% of the packed structure at any scale — a fixed and small
+    /// price for having no ceiling at all (#1097).
+    pub starts: Vec<u64>,
     /// flat `combined_weights` (`weight/effLen`), aligned to `labels`; used by EM
     pub combined: Vec<f64>,
     /// flat raw conditional weights, aligned to `labels`; used by Gibbs
@@ -146,16 +153,6 @@ impl PackedEqClasses {
             .filter(|(g, _)| g.valid)
             .map(|(g, _)| g.txps.len())
             .sum();
-        // The CSR offsets below are `u32`. Past `u32::MAX` incidences the `as u32`
-        // cast would wrap silently and `class()` would hand the optimizer slices
-        // belonging to other classes — wrong abundances with no diagnostic. Fail
-        // loudly instead.
-        assert!(
-            total_labels <= u32::MAX as usize,
-            "equivalence-class CSR needs {total_labels} incidences, which exceeds \
-             the u32 offset limit ({}); please report this input",
-            u32::MAX
-        );
         let mut labels = Vec::with_capacity(total_labels);
         let mut starts = Vec::with_capacity(n + 1);
         let mut combined = Vec::with_capacity(total_labels);
@@ -164,7 +161,7 @@ impl PackedEqClasses {
         let mut counts = Vec::with_capacity(n);
         // The first class starts at offset 0; every later entry is appended after
         // that class's data has been copied in.
-        starts.push(0u32);
+        starts.push(0u64);
         let mut total = 0u64;
         for (group, value) in &eq.classes {
             // Classes the optimizer marked degenerate are simply not carried over.
@@ -178,7 +175,7 @@ impl PackedEqClasses {
             }
             counts.push(value.count);
             total += value.count;
-            starts.push(labels.len() as u32);
+            starts.push(labels.len() as u64);
         }
         Self {
             labels,
@@ -230,6 +227,10 @@ impl PackedEqClasses {
     ///
     /// Returning borrowed slices means no allocation and no copying: this is
     /// called once per class per iteration, millions of times per run.
+    ///
+    /// The `as usize` narrowing cannot lose anything: an offset only exceeds
+    /// `usize` on a 32-bit target, where `labels` would have had to exceed a
+    /// 16 GB allocation to get there.
     #[inline]
     pub fn class(&self, i: usize) -> (&[u32], &[f64]) {
         let s = self.starts[i] as usize;


rob-p:
Correcting the performance claim in the description above; the code is unchanged.

I originally reported +1.10% at t=2.06 and said the effect "firmed up as n grew rather than collapsing". That was wrong, and wrong in the flattering direction. Two more rounds finished after I ran the analysis, and with the full sample the picture is:

| n | delta | t |
| --- | --- | --- |
| 11 | +1.62% | 1.80 |
| 20 | +1.40% | 2.07 |
| 27 | +1.10% | 2.06 |
| **29 (all)** | **+0.95%** | **1.86** |

95% CI at n=29 is `-0.009 .. +0.232 s`, spanning zero, and `u64` is slower in 16 of 29 rounds — 55%, near a coin flip.

The mistake was analysing at a point I happened to stop at rather than at the end of the run, and n=27 sat near the peak of t. The honest reading is a point estimate around +1% that this benchmark cannot separate from noise — neither a demonstrated regression nor a demonstrated absence of one.

That does not change the recommendation, and arguably strengthens it: the measured upper bound on the cost is ~1% of one phase, against removing a hard ceiling and deleting an assertion. But "no measurable regression" would be overclaiming, so the description now says what the data supports.

PR #1113, opened 2026-08-09
perf(map): drop the per-fragment and per-call maps from the hot paths

rob-p:
Part of #1085 — its map-side half. The quant-side half landed with #1099.

**This is the first change out of this audit with an unambiguous end-to-end win.** 10M human fragments, reads path, `-p 16`, interleaved, warm-up discarded:

| | mean | sd |
| --- | --- | --- |
| develop | 28.845 s | 0.317 |
| this PR | **28.156 s** | 0.218 |

**−0.689 s (−2.39%), t = −5.28**, 95% CI −0.950 … −0.428, faster in **9 of 9 rounds**. Worth stating against the run of nulls in #1092: every other structural item measured |t| ≲ 2 with a CI straddling zero. This one does not.

It is also the size the microbenchmark predicts — dedup runs once per fragment, and at typical n the scan is 7–8× faster than allocating a map (8.2 ns vs 61.0 ns at n=4); ~50 ns × 10M ≈ 0.5 s.

## What changed

**Deduplicating raw mappings by transcript** allocated an `AHashMap` per fragment to handle what is typically 1 to 20 entries. Now a hybrid, split at the measured crossover:

| n / distinct | AHashMap | prefix scan | reused table |
| --- | --- | --- | --- |
| 4 / 4 | 61.0 ns | **8.2** | 38.5 |
| 20 / 12 | 231.5 | **29.1** | 104.7 |
| 60 / 30 | 513.9 | **173.6** | 270.4 |
| 100 / 50 | 851.9 | **399.7** | 429.3 |
| 128 / 64 | 1325 | 752.3 | **586.7** |
| 160 / 80 | 1439 | 1004.5 | **591.5** |

`DEDUP_HASH_THRESHOLD = 112`, the midpoint of the crossing interval, with a `const` assertion that it stays in a range where both paths are reachable in a real run.

**`best_per_target`** allocated an `AHashSet` per call; it now scans the kept prefix in place, preserving both the survivor and the highest-coverage-first ordering.

## The part that took two attempts

The tie rule is load-bearing. The map replaced an entry only on a **strictly** greater score, so among equal scores on one transcript the first in arrival order survived — and those records carry different positions, so a different survivor shifts the fragment's bias and fragment-length weights.

My first attempt sorted by `(tid, score, ref_pos)`. Every `sample_data` check passed, and it **silently changed 55 equivalence classes** on 10M human fragments (407,855 vs 407,800) — small enough to miss, large enough to matter. Only the human-scale byte comparison caught it.

Both paths now apply the original rule, with a test asserting they agree on cases loaded with ties, negative-controlled by flipping the table path to `>=` (fails on a 2-mapping case).

Conversely the survivors' *order* is **not** part of the contract, which is worth recording: the map being replaced was seeded per process (`ahash` `runtime-rng`), and develop is byte-identical across separate runs regardless, so the order it yielded cannot be load-bearing.

## Scratch, not thread-local

`DedupScratch` is threaded through 4 signatures rather than held in a `thread_local`. On 200 mappings:

| | |
| --- | --- |
| threaded parameter | **768 ns** |
| `thread_local` | 839 ns |
| fresh allocation per fragment | 1468 ns |

Reuse is what matters most — allocating on demand is 1.9× worse — and the parameter recovers the remaining ~10% that the lazy-init check and `RefCell` borrow cost (a `HashTable` has a `Drop`, so a bare const-init TLS is not available). `QuantProcessor` already owns a `RecordScratch` from #1099, so it has a natural home; the two convenience wrappers that allocate their own `out` allocate a scratch too, which costs them nothing relative to what they return.

Hashing is `ahash` with fixed seeds. `hashbrown::HashTable` takes the hash from the caller and has no default hasher, so `std`'s SipHash is not reachable here.

## Verification

- `quant.sf` **byte-identical** to develop on 10M human fragments (407,800 eq-classes, 9,221,625 mapped)
- byte-identical on `sample_data` across plain / `--seqBias --gcBias --posBias` / `--softclip` / `--sketch`
- full suite, `clippy --workspace --all-targets -D warnings`, fmt clean

## Not touched

`cfg.chain.clone()`, which #1085 lists among "4 or more Vecs per fragment". `ChainConfig` is `{i32, i32, f32, usize}` — a ~24-byte POD copy, not heap traffic. Same false positive as the `PiscemStreamingQuery` item already withdrawn in that issue.

### DIFF
--- a/crates/salmon-align/src/rad.rs
+++ b/crates/salmon-align/src/rad.rs
@@ -2197,7 +2197,12 @@ mod group_tests {
     /// `push_sorted` skips the comparison, so it is only correct when something
     /// upstream ordered the input. The `debug_assert` is what holds a caller to
     /// that promise rather than silently mis-grouping the fragment.
+    // `debug_assert!` compiles out under `--release`, which is how CI runs the
+    // suite, so this can only ever pass where debug assertions are live. Gating
+    // it is the point of the assert: it is a development-time guard, not a
+    // runtime check the release build pays for.
     #[test]
+    #[cfg(debug_assertions)]
     #[should_panic(expected = "out-of-order placement")]
     fn push_sorted_rejects_an_out_of_order_placement() {
         let mut o: TidOrdered<CorePlacement> = TidOrdered::default();
--- a/crates/salmon-map/src/collect.rs
+++ b/crates/salmon-map/src/collect.rs
@@ -32,7 +32,7 @@
 //! anchors increase colinearly with the reference coordinate. The resulting
 //! chains are stamped with `is_fw = false` so a later stage can map them back.
 
-use ahash::{AHashMap, AHashSet};
+use ahash::AHashMap;
 
 use piscem_rs::index::contig_table::EntryEncoding;
 use piscem_rs::index::reference_index::ReferenceIndex;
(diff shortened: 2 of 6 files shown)


rob-p:
CI is green now — the failure was `push_sorted_rejects_an_out_of_order_placement`, which is `#[should_panic]` on a `debug_assert!` while CI runs `cargo test --workspace --release`. Gated on `#[cfg(debug_assertions)]` in 43dd17f.

That test came in with #1096, so `develop` has been red for every PR since it merged; this fixes it for all of them.

Follow-up in #1114 closes out the remaining #1085 items (stacked on this branch).

PR #1114, opened 2026-08-09
perf(map): reuse the per-fragment raw mapping and revcomp buffers

### LINKED ISSUE #1085: perf(map,quant): finish the *_into scratch pattern through the mapping and record path

rob-p:
Closes #1085.

Stacked on #1113 — please merge that first; the base here is
`perf/1085-map-side-allocs`, so this diff shows only the new work.

#1085 listed four hot-path allocations. #1113 handled the two that measured
(the per-fragment `AHashMap` in `finalize_mappings_counted_into` and the
per-call `AHashSet` in `best_per_target`, together −2.39%, t=−5.28, faster in
9/9 rounds). This PR closes out the rest, so nothing in the issue is left
undecided.

### What changed

`MapScratch` gathers the mapper's per-fragment buffers into one struct that a
worker allocates once and then refills for the rest of the run. The mapping
entry points take one scratch parameter instead of three.

| buffer | before | now |
| --- | --- | --- |
| `raw` | fresh `Vec` per fragment | reused; `finalize_mappings_counted_into` drains it |
| `revcomp` | one `Vec` per orphan candidate examined | reused via `revcomp_into` |
| `dedup` | already threaded through from #1113 | folded into `MapScratch` |

`finalize_mappings_counted_into` now takes `raw: &mut Vec<RawMapping>` and
leaves it empty on every return path — including the two decoy early-returns,
which previously just returned and now clear explicitly. A new test drives all
five paths (empty input, decoy-only, decoy-dominated, ordinary, hashed-dedup)
and asserts the buffer comes back empty; removing either `clear()` makes it
fail with the path named.

### Correctness

Byte-identical against the #1113 build: sample data plain / bias / sketch, and
10M human fragments at `-p 1` — `quant.sf`, the eq-class file, and the
`meta_info.json` counters.

### Performance — null

10M human fragments at `-p 16`, interleaved A/B against the #1113 build, ten
rounds with the warm-up discarded:

```
  #1113 base   28.025s  sd=0.202
  + this PR    27.945s  sd=0.142
  delta        -0.080s  (-0.29%)
  paired       t=-0.92   95% CI  -0.255 .. +0.094 s
  faster in 4/9 rounds
```

That is not a win. The allocations are real, but at this scale they are
invisible against the alignment work around them — the same result the other
structural allocation changes in this series produced, and the reason #1113
was worth separating out. I am keeping this only for the invariant it
establishes (a worker allocates its mapping buffers once), not on a
performance claim. Happy to drop it if you would rather not take a change that
does not measure.

### Deliberately not changed

**The per-read `candidates` `Vec`** (#1085 item 2).
`join_reads_and_filter(left: Vec<MappingCandidate>, right: Vec<MappingCandidate>, ..)`
takes both by value and *moves* candidates into the `JointMapping`s it returns.
Each `MappingCandidate` owns a `MemChain`, so retaining the caller's buffers
would require cloning those chains — more work than the allocation saves. This
is a design change to the pairing interface, not a buffer swap, so it is out of
scope here.

**`cfg.chain.clone()` per read.** A false positive: `ChainConfig` is plain
`Copy` scalars, so the clone is a stack copy of a few words with no allocation.

### DIFF
--- a/crates/salmon-map/src/align.rs
+++ b/crates/salmon-map/src/align.rs
@@ -111,7 +111,7 @@ pub(crate) fn revcomp(s: &[u8]) -> Vec<u8> {
 
 /// Reverse-complement `s` into `out` (cleared first), reusing its allocation.
 #[inline]
-fn revcomp_into(s: &[u8], out: &mut Vec<u8>) {
+pub(crate) fn revcomp_into(s: &[u8], out: &mut Vec<u8>) {
     out.clear();
     out.extend(s.iter().rev().map(|&b| comp(b)));
 }
--- a/crates/salmon-map/src/lib.rs
+++ b/crates/salmon-map/src/lib.rs
@@ -73,7 +73,8 @@ pub use mem::Mem;
 pub use pair::{join_reads_and_filter, JointMapping, PairingConfig};
 pub use salmon_core::RefProvider;
 pub use score::{
-    filter_sketch_decoys, finalize_mappings, DedupScratch, RawMapping, ScoreConfig, ScoredMapping,
+    filter_sketch_decoys, finalize_mappings, DedupScratch, MapScratch, RawMapping, ScoreConfig,
+    ScoredMapping,
 };
 pub use sketch::{
     map_read_pair_sketch, map_read_pair_sketch_into, map_single_read_sketch,
--- a/crates/salmon-map/src/mapper.rs
+++ b/crates/salmon-map/src/mapper.rs
@@ -26,7 +26,7 @@ use piscem_rs::index::reference_index::ReferenceIndex;
 use piscem_rs::mapping::hit_searcher::HitSearcher;
 use salmon_core::{LibraryFormat, MateStatus, ReadOrientation, ReadStrandedness, ReadType};
 
-use crate::align::{align_chain, align_in_window, revcomp, AlignConfig};
+use crate::align::{align_chain, align_in_window, AlignConfig};
 use crate::collect::{
     best_per_target, collect_read_mems, consensus_filter, MappingCandidate, MemCollectorConfig,
 };
@@ -49,7 +49,7 @@ pub enum SeedMode {
 }
 use crate::pair::{join_reads_and_filter, PairingConfig};
 use crate::score::{
-    finalize_mappings_counted_into, DedupScratch, RawMapping, ScoreConfig, ScoredMapping,
+    finalize_mappings_counted_into, MapScratch, RawMapping, ScoreConfig, ScoredMapping,
 };
 use salmon_core::RefProvider;
 
@@ -152,7 +152,7 @@ pub fn map_single_read<'idx, R: RefProvider>(
         refs,
         read,
         cfg,
-        &mut DedupScratch::default(),
+        &mut MapScratch::default(),
     );
     out
 }
@@ -166,7 +166,7 @@ pub fn map_single_read_into<'idx, R: RefProvider>(
     refs: &R,
     read: &[u8],
     cfg: &MapConfig,
-    scratch: &mut DedupScratch,
+    scratch: &mut MapScratch,
 ) {
     let cands = consensus_filter(
         best_per_target(collect_candidates(
@@ -181,7 +181,11 @@ pub fn map_single_read_into<'idx, R: RefProvider>(
     );
     let had_candidates = !cands.is_empty();
     let mut below = 0u32;
-    let mut raw = Vec::with_capacity(cands.len());
+    // Reused across fragments; `finalize_mappings_counted_into` consumes it by
+    // draining, so it comes back empty for the next one.
+    let raw = &mut scratch.raw;
+    raw.clear();
+    raw.reserve(cands.len());
     for c in cands {
         if let Some(aln) = align_chain(read, refs.ref_seq(c.tid), &c.chain, &cfg.align) {
             if aln.valid {
@@ -228,7 +232,7 @@ pub fn map_single_read_into<'idx, R: RefProvider>(
         }
     }
     let (decoy_dominated, below_final) =
-        finalize_mappings_counted_into(out, raw, &cfg.score, scratch);
+        finalize_mappings_counted_into(out, raw, &cfg.score, &mut scratch.dedup);
     set_last_map_stats(MapStats {
         had_candidates,
         decoy_dominated,
@@ -255,7 +259,7 @@ pub fn map_read_pair<'idx, R: RefProvider>(
         r1,
         r2,
         cfg,
-        &mut DedupScratch::default(),
+        &mut MapScratch::default(),
     );
     out
 }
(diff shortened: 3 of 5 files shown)


rob-p:
Pushed 9342e75, dropping the redundant producer-side `raw.clear()` calls.

`scratch.raw` had two invariants asserting the same thing — the consumer drains on every return path, and both producers cleared again on entry — with only the consumer side tested. Neither producer can return between filling `raw` and the drain, and `MapScratch::raw` is `pub(crate)`, so nothing external can leave anything in it. The entry clears were hiding a violation rather than reporting one; they are now `debug_assert!`s that compile out of release builds.

Verified a pure no-op against the pre-removal build of this branch: sample data plain / bias / sketch / single-end, and 10M human fragments at `-p 1` — `quant.sf`, `eq_classes.txt.gz`, `ambig_info.tsv`, `fld.gz`, the bias grids and the `meta_info.json` counters all byte-identical (meta_info differs only in timestamps and peak RSS).

Negative control on the assertion: reverting the consumer's `drain(..)` to a non-consuming iteration aborts a debug run at `mapper.rs:314` with the expected message. Worth noting the sample-data index has no decoys, so a debug run never reaches the two decoy early-returns — those remain covered only by `raw_buffer_is_emptied_on_every_path`.

rob-p:
Rebased onto develop (post-#1113) and retargeted; base is now `develop`.

### Verified on a decoy-bearing index

Rebuilt the GRCh38 gentrome index (cDNA+ncRNA, primary assembly as decoy — 194 decoys, `first_decoy_index=532552`, index format v1) and reran, so the two decoy early-returns in `finalize_mappings_counted_into` are exercised for real rather than only by the unit test. Release parity against develop, all byte-identical:

| run | result |
| --- | --- |
| 1M PE, `-p 1`, `--dumpEq` | quant.sf IDENTICAL; eq_classes IDENTICAL (702,048 lines); 51,532 decoy fragments |
| 300k PE `--allowDecoyOrphans` | quant.sf + counters IDENTICAL; 12,221 decoy fragments |
| 300k PE `--sketch` | quant.sf + counters IDENTICAL; 14,914 decoy fragments |
| 300k SE | quant.sf + counters IDENTICAL; 14,336 decoy fragments |

Debug builds (so the new `debug_assert!`s are live) ran clean on `--sketch` and single-end. The new `scratch.raw was not drained` assertion never fired in any configuration.

### Pre-existing failure found, unrelated to this PR

Debug builds abort on this index in selective-alignment paired mode (plain, `--seqBias --gcBias`, `--allowDecoyOrphans`) at the *existing* `debug_assert` "a fragment's surviving mappings must share one status". I confirmed this is not mine: a debug build of `develop` at 174d2d24 fails identically, same assertion, same 16 threads — only the line number moved (806 -> 807) because this PR adds a doc comment above it. Filing separately; it does not block here.

PR #1119, opened 2026-08-09
Adopt the adaptive thread broker (piscem-rs 0.9.0, paraseq-temp)

rob-p:
## What this is

Adopts piscem-rs's adaptive thread broker in salmon: `-p` becomes a single
execution-slot budget shared between gzip decoding and mapping, solved from
live measurement rather than split by a fixed guess. Includes `--decoder`
(auto/serial/parallel/parallel=N) and `--threadPolicy <FILE>`.

**Ready for review.** Every stage below is verified; nothing is provisional.

## Dependency changes

- **piscem-rs 0.6.4 → 0.9.1** (`rapidgzip` feature on). 0.7.x/0.8.0 were yanked
  upstream; 0.9.1 additionally requires the pool fix below.
- **paraseq 0.4 → paraseq-temp 0.5.0-pre.1** — required, not incidental:
  `ThreadPool` is the resize mechanism the broker acts through. Lib target is
  still `paraseq`; no import changes. `paraseq` remains the official crate;
  move back at 0.5.0.
- **thread-broker 0.1.0**, **rapidgzip-core ≥ 0.3.1** (transitively pinned).

## A deadlock found, rooted, and fixed upstream

Enabling the broker on paired input at `-p ≥ 48` hung salmon. Bisect: the split
was not the variable (pinned `parallel=4` plans the identical 56/8 split and
completes — pinning disables the broker, which is the trigger); paired input
and thread count were. Backtraces showed a permit-starvation picture in
rapidgzip's shared `DecoderPool`.

Root cause — **not in salmon, not in paraseq**: a lost wakeup in
`DecoderPool::release`, which notified waiters without holding the scheduler
lock. The final release could land between a waiter's admission check and its
park, leaving free permits, parked decoders, and no future notify. piscem never
reproduced it because its decode share keeps admission off the contended path;
salmon's heavier consumer drives the broker to shrink the pool, which is what
makes the window reachable. Fixed in **rapidgzip-core 0.3.1** with a regression
test that deadlocks the unfixed pool in <2 s; **piscem-rs 0.9.1** requires it.
Verified here: all six previously-hanging configurations complete (37–62 s)
with mapped counts identical to their serial-decoder runs. Full narrative:
`notes/pool-lost-wakeup-deadlock.md`.

## Engagement thresholds: measured, all four cells

`min_threads_per_stream = 1 + C/P`, read from the broker's own solved
`producer_cost_share` at `-p 64` on 26.1 M paired human RNA-seq fragments:

| cell | share | threshold |
|---|---|---|
| sketch | 0.094–0.100 (×3) | **10** |
| sketch + `--deterministic` | 0.110 (×2) | **9** |
| selective alignment | ~0.02 | **50** |
| SA + `--deterministic` | ~0.01 (not resolvable from SA) | **50** |

- Sketch confirms the `--deterministic` prediction: lighter mapping → higher
  producer share → earlier engagement. Thresholds round down because the
  measured cost asymmetry favours engaging (missing parallel cost up to 4.2×;
  engaging needlessly cost 8–28%).
- SA's figure carries an honest error bar (a CPU-ratio cross-check suggests
  ~24) but its content survives it: with paired input, even the low estimate
  engages only at `-p ≥ 48`. Serial decode is simply right for SA at nearly
  every real budget — which the theory predicted and the wall clocks confirm
  (SA parallel ≈ SA serial at `-p 64`).
- SA's `--deterministic` delta is not resolvable: alignment scoring dominates
  its cost, and the deterministic pass strips bookkeeping that is a large
  fraction of sketch's cost and a small one of SA's. Constants set equal; the
  test asserts `≤`, not an invented `<`.

## Safety under a resizable pool

- `quant.sf` **byte-identical** to pre-change `develop` at `-p 1` (238,443
  rows), serial decoder both sides.
- Thread-local flush contract pinned by test: counters merged at thread end,
  output buffers flushed per batch and at retirement, `Clone` rebuilds fresh
  state. The test churns the pool through ~134 real retire/respawn cycles and
  asserts exactly-once delivery **by id**, with a non-vacuity guard.
- The decoder decision is **per input file** (metadata-only classification —
  a FIFO is never read), pinned by tests: a plain file among gzip inputs does
  not cost them the parallel decoder, and no slots are reserved for decoders
  that don't exist.

## Verification

277 workspace tests, clippy clean, zero non-registry dependencies in the
lockfile.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2268,9 +2268,9 @@ checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
 
 [[package]]
 name = "piscem-rs"
-version = "0.9.0"
+version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c20e73778bc07814ffd1576873608f7a160c12b68e690c66a324a12567244cc5"
+checksum = "91e9d61329e506636f779f403afda39d2f05b4a0f5d3d98aed6e7d357ad0c9ea"
 dependencies = [
  "ahash",
  "anyhow",
@@ -2493,9 +2493,9 @@ dependencies = [
 
 [[package]]
 name = "rapidgzip-core"
-version = "0.3.0"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "73479522ef5ac15e4015ff42faab6c3f6372f88b95ffd5fa4577f93dd76e907e"
+checksum = "da5be74bd353811db3efff881e87200a18150d72fb0b1d39ea04e7972a5b4ca1"
 dependencies = [
  "bon",
  "crossbeam-deque",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2824,6 +2824,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tempfile",
+ "thread-broker",
  "tracing",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -619,6 +619,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a6ef517f0926dd24a1582492c791b6a4818a4d94e789a334894aa15b0d12f55c"
 
 [[package]]
+name = "cpu-time"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e9e393a7668fe1fad3075085b86c781883000b4ede868f43627b34a87c8b7ded"
+dependencies = [
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "cpufeatures"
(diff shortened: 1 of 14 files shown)


rob-p:
The SA engagement threshold is now empirically bracketed rather than estimated.

Ground truth (serial vs parallel wall, full 26.1M-pair input): at `-p 64` serial wins by 12%; at `-p 128` parallel wins by 7% (counterbalanced, both passes agree). Crossover ∈ (32, 64] per stream — the shipped **50** makes the correct decision at both measured budgets, while the model-derived alternatives (24, 29) would have engaged at `-p` 48–58 and taken the 12% regression.

Also settled by the new integrated busy-time telemetry (`P` constant at ~26 s across all cells): `--deterministic` removes **58%** of sketch's per-fragment cost but only **2%** of SA's — so sketch±det keep distinct thresholds (10/9) and SA±det stay equal, as the source now documents with the full evidence table.

PR #1123, opened 2026-08-12
feat(quant): --spliced, mapping output in genome coordinates (stacked on #1141, #1142)

BenjaminDEMAILLE:
**Rebased and narrowed, now into a three-PR stack.** Following the scope objection below, this PR was split:

1. **#1141** the statements the mapping output makes that are *false*: a CIGAR synthesized from the read length, `AS` describing a different alignment from the CIGAR beside it, `MAPQ` pinned to 1. 2.4% of records were self-contradicting.
2. **#1142** the fields SAM allows to be absent: `QUAL`, `@SQ M5`/`UR`, `MC`/`MQ`, `ZW`, `@RG`/`--rgLine`, unmapped records. Omissions, not falsehoods, so a separate decision.
3. **This PR**, stacked on both: the genome projection alone. `splice.rs`, `--spliced`, `XS`, the projection accounting. Seven commits, ~1.4k lines.

GitHub cannot point a base branch at a fork, so until #1141 and #1142 merge the diff shown here still includes them.

## What `--spliced` does

A transcript has no introns, so a read spanning an exon junction is contiguous in transcriptome coordinates and its CIGAR never contains an `N`. `--spliced` re-expresses each record in genome coordinates: the alignment is cut at the exon boundaries the annotation declares, the gaps reappear as `N`, `@SQ` names chromosomes, and records gain `XS:A` for the transcript's genomic strand.

It needs `--annotation <gtf|gff>` for the exon structures and `--genome <fasta>` for the chromosome lengths `@SQ` must declare (a `.fai` is used when present). A transcript on the minus strand runs against the genomic axis, so its alignments are turned around whole: CIGAR, sequence, qualities, strand bits, `TLEN` and `MD` together. `M5` is omitted, because the references are then chromosomes salmon never loaded and a guessed checksum is worse than none.

Placements on a reference the annotation does not describe (a decoy, or a transcript absent from the GTF) are refused rather than half-projected, counted, and reported as a warning at the end of the run. With `--sampleUnaligned` a fragment whose every placement was refused is written as unaligned instead of vanishing.

## Why the genome coordinates are the point

The correctness half of this work stands on its own, but it does not remove a second alignment: a transcriptome-coordinate BAM cannot be fed to the tools that need genome coordinates. Two concrete workflows in our pipeline:

**RNA-seq QC.** Picard `CollectRnaSeqMetrics`, RSeQC (`read_distribution.py`, `geneBody_coverage.py`, `inner_distance.py`) and Qualimap `rnaseq` all take a genome-coordinate BAM plus an annotation. A salmon-quantified sample therefore gets aligned twice today: once by salmon for the quantification we actually want, and once by STAR purely to produce a file for QC. The second alignment is 100% of the QC cost and none of the analysis. With `--spliced` the QC reads the same alignments the quantification used, which is also the more honest QC: it describes the file the numbers came from rather than a different aligner's view of the same library.

**Allele-specific expression.** GATK `ASEReadCounter` and phASER need genome coordinates and real CIGARs with `NM`/`MD`, the same second alignment for the same reason.

I am one lab with one pipeline, not a queue of requests, so this is evidence rather than demand. If it is not enough, a decision criterion would help more than a merge: what would count? Two or three independent reports? A user issue naming a tool by name? Happy to keep this parked until then, and #1141 stands on its own either way.

## Verification (unchanged from the original review)

Against samtools 1.24, on 500 genes over a 5 Mb simulated chromosome (three to eight exons, both strands) and on human chr22 (hg38) + GENCODE v44, 300k fragments, 1.85M records of which 685k span a junction:

- `NM`/`MD` match `samtools calmd` recomputed against the genome, on every projected record.
- `samtools quickcheck` passes, SAM round-trips through `samtools view -b`.
- Reverse-strand transcripts: `SEQ`/`QUAL` orientation, strand bits, `TLEN` sign and `MD` all agree after the flip.
- Full workspace test suite green, including the projector unit tests in `splice.rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2852,6 +2852,7 @@ name = "salmon-quant"
 version = "2.6.0"
 dependencies = [
  "anyhow",
+ "bramble-rs",
  "crossbeam-channel",
  "flate2",
  "jiff",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1767,6 +1767,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "59dbb09ed53f8e4f314e353dc6c1853ae5b4c480a668a422657804a544ea9f65"
 
 [[package]]
+name = "md-5"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "69b6441f590336821bb897fb28fc622898ccceb1d6cea3fde5ea86b090c4de98"
+dependencies = [
+ "cfg-if",
+ "digest",
+]
+
+[[package]]
 name = "mem_dbg"
 version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2846,6 +2856,7 @@ dependencies = [
  "flate2",
  "jiff",
  "libradicl",
+ "md-5",
  "noodles-bam",
  "noodles-bgzf",
  "noodles-sam",
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -532,8 +532,8 @@ struct QuantArgs {
     /// Write the names of unmapped fragments to aux_info/unmapped_names.txt.
     #[arg(long = "writeUnmappedNames")]
     write_unmapped_names: bool,
-    /// Write per-mapping SAM records to this file. Records carry a base-level
-    /// CIGAR with NM and MD, describing the alignment the record reports.
+    /// Write per-mapping SAM records to this file. Records carry base qualities,
+    /// a base-level CIGAR with NM and MD, and the mate and placement tags.
     /// `--writeSam` is an accepted alias, naming the format explicitly to pair
     /// with `--writeBam`.
     #[arg(
@@ -915,10 +915,17 @@ struct QuantArgs {
     /// sampled BAM (requires --sampleOut).
     #[arg(short = 'u', long = "sampleUnaligned")]
     sample_unaligned: bool,
-    /// [accepted; not yet implemented] write qualities into the sampled BAM
-    /// (only meaningful with --sampleOut / --writeMappings).
+    /// [accepted for salmon compatibility; has no effect] base qualities are not
+    /// optional output. They are written into every record whenever the input
+    /// carries them, with or without this flag.
     #[arg(long = "writeQualities")]
     write_qualities: bool,
+    /// Read group to declare in the mapping output's header, as tab-separated
+    /// KEY:value fields with a mandatory ID (for example
+    /// `ID:sample1\tSM:sample1\tPL:ILLUMINA`). A literal \t is accepted in place
+    /// of a tab. Every record is then tagged RG:Z:<ID>.
+    #[arg(long = "rgLine", value_name = "LINE")]
+    rg_line: Option<String>,
     /// [accepted; not yet implemented] hit-filtering policy (BEFORE/AFTER/BOTH/
     /// NONE) for selective alignment; the Rust port filters after chaining.
     #[arg(long = "hitFilterPolicy")]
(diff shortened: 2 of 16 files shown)


BenjaminDEMAILLE:
@rob-p little reminder. this correct a silent error where bam is not complete

rob-p:
Marking this **postponed** rather than closed, and converting to draft.

The features are genuinely nice — qualities, real CIGARs, NM/MD, read groups, unmapped records, spliced genome-coordinate output would make `--writeMappings` a "complete" SAM/BAM producer. But that's also the argument against merging it now: it's ~4.5k lines expanding salmon into territory it has never occupied. `--writeMappings` has always been a diagnostic view of salmon's *mappings*, not a general-purpose aligner output, and C++ salmon never promised more. Every field this adds is surface we then own — NM/MD correctness against every reference edge case, read-group plumbing, the spliced projection — and it all has to be maintained whether or not anyone uses it.

What would change the calculus is demonstrated need: users showing up with concrete workflows that require salmon's mapping output to be a full SAM/BAM (rather than using a dedicated aligner for that job, which remains the natural tool). If those requests accumulate, this PR is the right starting point and the work here won't be wasted — that's why it's parked (labeled `postponed`, retargeted at `develop`, which is the integration branch) instead of closed.

BenjaminDEMAILLE:
Thanks, and the scope objection is fair, so I acted on it rather than arguing it: this PR is split.

**#1141** is the half that is not new territory. `--writeMappings` has always been a diagnostic view of salmon's mappings, and I agree that is all it should promise, but the view is currently wrong in ways that are not a matter of how much it promises:

- The CIGAR is built from the read length, so a placement with an indel is written as if it had none and every base past the gap is off by the gap length.
- Auditing `AS` against a recomputed `NM` over a whole output found 2.4% of records self-contradicting: a `15M1I1D15M1I1D...` sawtooth at `NM:i:80` on a 100 base read, beside an `AS` claiming three mismatches. Three separate causes, each of which alone produces a false record (anchor search bounded by `indel_margin`, the search abandoned when the read did not fit at its anchor, and `AS` describing a different alignment from the CIGAR).
- An orphan's mapped mate carries `MATE_UNMAPPED`, but that mate is never written, so the file describes a read it does not contain.
- `MAPQ` is hardcoded to `1`, so `samtools view -q 255` discards the entire file.
- `--sampleUnaligned` and `--writeQualities` are accepted, warn "not yet implemented", and do nothing.

None of that is salmon promising to be an aligner. It is the diagnostic view misdescribing salmon's own mappings, which is worse than a minimal view. #1141 fixes exactly that and adds no new output mode. Its cost is zero when no mapping output is requested (a run that writes no records never enters the realization module), and about 2x on `--writeBam` runs, measured.

**This PR** keeps the part that genuinely is new surface: the genome projection, `--spliced`, `XS`. It is now stacked on #1141 and I have narrowed it to those seven commits (the diff here still shows #1141 until that merges, since a base branch cannot point at a fork).

On demonstrated need, I have added the concrete workflow to the description rather than claiming a queue of users: Picard `CollectRnaSeqMetrics`, RSeQC and Qualimap `rnaseq` all require a genome-coordinate BAM plus an annotation, so a salmon-quantified library is aligned a second time with STAR purely to produce a QC input, and the QC then describes a different aligner's view of the library than the one the numbers came from. ASE calling (`ASEReadCounter`, phASER) needs the same file for the same reason. That is one lab, not a groundswell, so if it is not enough I would rather have your criterion than a merge: what would count as demonstrated? Parked is fine in the meantime, and #1141 does not depend on the answer.

BenjaminDEMAILLE:
Follow-up on the split: it is now three PRs rather than two, because two of the things I had bundled as "correctness" are not the same kind of claim.

- **#1141** is what the output states *falsely* about salmon's own mappings: the CIGAR is synthesized from the read length so a placement with an indel is written as if it had none, `AS` came from the mapper while the CIGAR came from re-aligning (2.4% of records self-contradicting on an `AS`-vs-`NM` audit), and `MAPQ` is the constant 1 so `-q 255` discards the whole file. Three commits.
- **#1142** is what the output *omits*, which SAM permits: `QUAL` may be `*`, and `M5`, `UR`, `MC`, `MQ`, `ZW` may be missing. Useful, but a weaker claim, so it should be judged separately rather than smuggled in beside the fixes. Two commits, stacked on #1141.
- **This PR** is the genome projection alone, stacked on both.

The distinction matters for your scope argument: a diagnostic view that omits fields is defensible, and I am not arguing #1142 is obligatory. A diagnostic view that describes an alignment salmon did not make is not, whatever its scope, which is the whole of #1141.

PR #1126, opened 2026-08-14
perf(deps): hash with foldhash instead of ahash

BenjaminDEMAILLE:
Every hashed key on the mapping hot path is small and fixed-size: `(u32, bool)` in `collect::group_by_tid`, `u32` in the paired-tid scratch set and in the dedup table, a `(kind, &[u8], &[u8])` triple in the gap cache. foldhash's fixed-size path is a few multiply/rotate steps with no per-byte streaming loop, which is exactly that shape.

### Measured

A/B of the three map shapes in isolation, identical key stream per hasher, divan medians (200 samples; 100 for the cache):

| shape | ahash | foldhash | rapidhash |
|---|---|---|---|
| `best_per_tid` (`u32` -> mapping) | 387.7 µs | **354.6 µs** (-8.5%) | 378.2 µs |
| `group_by_tid` (`(u32,bool)` -> mems) | 1.150 ms | **1.094 ms** (-4.9%) | 1.121 ms |
| gap-cache key (byte-slice pair) | 123.9 µs | 114.1 µs (-7.9%) | **112.8 µs** |

`examples/bench_align`, which drives the real gap cache, best of three runs per case, ns per alignment on the default anchored path:

| case | before | after |
|---|---|---|
| 51 bp, 0 mm | 21.7 | 20.8 |
| 51 bp, 2 mm | 21.2 | 20.7 |
| 100 bp, 0 mm | 38.5 | 37.3 |
| 100 bp, 2 mm | 37.8 | 35.6 |
| 150 bp, 0 mm | 51.0 | 46.9 |
| 150 bp, 2 mm | 46.3 | 41.6 |

All twelve alignment checksums are unchanged.

### Notes

- The seeded hashers stay seeded: `ahash::RandomState::with_seeds` becomes `foldhash::fast::FixedState::with_seed`, so the gap cache and the dedup table keep the reproducible bucket layout their comments promise.
- rapidhash was measured alongside and only edges foldhash out on the byte-key cache while losing on both integer shapes, so it is not worth a second hashing crate in the tree.
- gxhash is faster still but requires AES-NI, which the portable `target-cpu` floor in `.cargo/config.toml` rules out.
- These maps are a slice of total quant time, not all of it. This is a few percent on the paths that use them, not an end-to-end figure; I have no real dataset here to quote a wall-clock delta on.

`cargo test --workspace`, `cargo clippy --workspace --all-targets` and `cargo fmt --check` are clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2744,10 +2744,10 @@ dependencies = [
 name = "salmon-index"
 version = "2.5.1"
 dependencies = [
- "ahash",
  "anyhow",
  "cf1-rs",
  "ensure_simd",
+ "foldhash 0.2.0",
  "needletail",
  "piscem-rs",
  "salmon-core",
@@ -2776,7 +2776,7 @@ dependencies = [
 name = "salmon-map"
 version = "2.5.1"
 dependencies = [
- "ahash",
+ "foldhash 0.2.0",
  "hashbrown 0.15.5",
  "ksw2rs",
  "piscem-rs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,8 +47,12 @@ crossbeam-channel = "0.5"
 # concurrent map for equivalence classes
 scc = "3"
 dashmap = "6"
-# fast (non-DoS-resistant) hasher for hot per-read integer-keyed maps
-ahash = "0.8"
+# fast (non-DoS-resistant) hasher for hot per-read integer-keyed maps.
+# foldhash's fixed-size integer path is a few multiply/rotate steps with no
+# per-byte loop, which is the shape of the (u32, bool) and u32 keys the mapper
+# builds per read. `gxhash` is faster still but needs AES-NI, which the portable
+# `target-cpu` floor in .cargo/config.toml rules out.
+foldhash = "0.2"
 # raw hash table, for caches that must probe with a borrowed key (no key
 # allocation on a lookup that hits)
 hashbrown = "0.15"
--- a/crates/salmon-index/Cargo.toml
+++ b/crates/salmon-index/Cargo.toml
@@ -18,7 +18,7 @@ serde_json = { workspace = true }
 tracing = { workspace = true }
 needletail = "0.7"
 sha2 = { workspace = true }
-ahash = { workspace = true }
+foldhash = { workspace = true }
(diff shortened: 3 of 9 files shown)


rob-p:
Thanks Benjamin — this was carefully done and honestly reported (the "I have no real dataset here to quote a wall-clock delta on" caveat turned out to be the load-bearing sentence), so we ran that missing measurement, and it decides against the change.

**Correctness first: the swap is provably output-neutral.** We verified on real data (200k pairs, GENCODE v49) that `quant.sf` is byte-identical pre/post in every mode we can compare bytes on — `--deterministic -p 16`, selective alignment `-p 1`, sketch `-p 1` — with identical mapped counts. That's worth stating because it wasn't obvious: `project_raw_hits`' consumers iterate the map in hash order, so a hasher change reorders the candidate stream, and only the downstream sorting/scoring makes that unobservable. No concerns on that axis.

**But the end-to-end effect isn't there.** Interleaved A/B on 10M real read pairs (GENCODE v49, selective alignment `-l IU -p 16`, 5 reps per binary, same machine, warm cache):

| binary | wall median (range) | CPU median (range) |
|---|---|---|
| develop (`ahash`) | 22.69 s (21.42–22.90) | 296.2 s (283.5–299.6) |
| this PR (`foldhash`) | 22.91 s (22.55–23.26) | 298.9 s (295.4–300.1) |

Both deltas are ~1%, inside the between-binary code-layout noise floor we've measured on this codebase, and the medians lean *against* the PR — CPU time included, which a pure hashing change should improve if it improves anything. The isolated 5–8% wins on the hash paths are real but those maps are too small a slice of a mapping run to surface.

**And the dependency ledger is slightly negative rather than positive.** `ahash` does not leave the tree (`bramble-rs`, `cf1-rs`, and `libradicl` all still pull it), and hashbrown 0.15 already vendors `foldhash` 0.1, so the workspace would carry `ahash` + two `foldhash` versions instead of just `ahash`.

So: a change whose only rationale is performance, with no measurable end-to-end performance change, touching hot-path code and adding a crate version. Closing. If a workload ever surfaces where these maps genuinely dominate (or `bramble`/`cf1`/`libradicl` migrate off ahash so the swap would actually consolidate hashers), this branch is the right starting point — and with #1128's divan harness the kernel-level claim will be easy to re-establish when that day comes.

PR #1127, opened 2026-08-14
perf(index): map refseq.bin instead of reading it into a Vec

BenjaminDEMAILLE:
The reference bases are the largest thing a run loads: a few hundred MB for a human transcriptome, several GB once a genome decoy is in the index. Both `load_ref_seqs` and `SalmonIndex::load_with_opts` pulled the whole `refseq.bin` into a `Vec<u8>` before the first fragment was processed, even though the bases are only ever read and bias correction touches a fraction of them.

`salmon_core::Bases` is now either `Owned(Vec<u8>)` (everything built in memory: annotation-reconstructed references, test fixtures) or `Mapped(Arc<Mmap>)`, deref'ing to `&[u8]` either way, so no accessor signature changes. `RefSeqs::from_mapped` runs the same offset-table validation as `from_concatenated`.

### Measured

Synthetic 509 MB / 250k-reference store, page cache warm, release build:

| mode | before | after | peak RSS before | peak RSS after |
|---|---|---|---|---|
| load only | 39.4 ms | **2.1 ms** | 496 MB | **10 MB** |
| load + read every 10th reference | 44.2 ms | **18.2 ms** | 496 MB | 420 MB |
| load + read every base | 47.3 ms | **34.4 ms** | 496 MB | 496 MB |

Checksums over the bases are identical in every mode. Even the read-everything case is faster, because the copy into anonymous memory is skipped. The mapped pages are also clean and reclaimable, unlike the anonymous buffer they replace, so several salmon processes on one machine share them instead of each paying the full resident cost.

### Safety

The mapping relies on the index directory being immutable for the duration of a run. salmon already depends on that: `sshash-lib` maps the sshash index the same way, and rebuilding an index under a running salmon was never supported. The `unsafe` block carries that note; it is the only one in `salmon-core`, and `memmap2` is already in the tree transitively via `sshash-lib`.

### Not included

`refseq_offsets.json` is still parsed as JSON. Converting it to a binary or zero-copy table needs an index-format bump and is a much smaller cost, so it is left for a separate change.

`cargo test --workspace`, `cargo clippy --workspace --all-targets`, `cargo fmt --check` all clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2721,6 +2721,7 @@ name = "salmon-core"
 version = "2.5.1"
 dependencies = [
  "flate2",
+ "memmap2",
  "niffler",
  "serde",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,9 @@ rand_pcg = "0.10"
 rand_distr = "0.6"
 # hashing
 xxhash-rust = { version = "0.8", features = ["xxh3", "xxh64"] }
+# memory-mapped reads of the index's reference-base blob; already in the tree
+# via sshash-lib, which maps the sshash index the same way.
+memmap2 = "0.9"
 # compression
 flate2 = "1"
 zstd = "0.13"
--- a/crates/salmon-core/Cargo.toml
+++ b/crates/salmon-core/Cargo.toml
@@ -9,6 +9,7 @@ authors.workspace = true
 description = "Shared core types for the Rust port of salmon (transcripts, library formats, log-space math)."
 
 [dependencies]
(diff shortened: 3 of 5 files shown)


rob-p:
Thank you for the suggestion, but given my previous experiences with `mmap`, the way that its behavior depends on the specific OS and the current page table load, and the way that it tends to perform poorly on NFS mounts, I'd prefer to take the hit in actually loading the relevant index components over which I have direct control into RAM via a read. I realize this might not always be possible, as sometimes other crates like epserde will make their own decisions, but we have control over this one ;P.

PR #1129, opened 2026-08-15
build(cli): add jemalloc/snmalloc diagnostic features and document the allocator choice

BenjaminDEMAILLE:
## What and why

salmon has shipped mimalloc as its `#[global_allocator]` since the port began, on the reasoning that the quant hot path is multithreaded and allocation-heavy. That reasoning was never checked against the alternatives. This PR adds the two missing ones behind features (matching the existing `sysalloc` diagnostic), and records what they actually measure.

**The default does not change.** The measurement says mimalloc stays. What this PR buys is that the next person to ask the question re-runs it in one command instead of re-arguing it.

## The measurement

GENCODE v50 (654,828 references), 10M simulated 2x100 bp pairs, `quant -l A -p 16`, 5 reps with allocators interleaved within each rep, Apple M4 Max. Median `mapping`-phase seconds (the only phase that allocates heavily; `em_bias` is compute-bound and serves as a control):

| allocator | median mapping | peak RSS |
|---|---|---|
| mimalloc (default) | 31.0 s | 3.84 GiB |
| snmalloc | 30.8 s | 4.22 GiB |
| jemalloc | 33.2 s | 3.66 GiB |
| system malloc | 42.7 s | 3.71 GiB |

- **System malloc is reliably ~38% worse** on mapping. This is the result that justifies shipping a replacement allocator at all, and it reproduced in every round.
- **jemalloc does not win.** Consistently ~7% behind mimalloc. Its strengths (fragmentation control, RSS on long-lived large heaps) are not what a batch quant run is bounded by.
- **mimalloc and snmalloc are tied.** Two independent clean rounds disagree in *direction* (one has snmalloc ahead, the other mimalloc ahead by 6%), so the gap is under the noise floor. snmalloc also costs ~0.4 GiB more peak RSS.

`docs/allocator-choice.md` has the full protocol and both traps this fell into first: the single-threaded microbenchmark ranks the allocators in an order the real workload reverses (it puts system malloc *ahead* of mimalloc), and a contaminated first round with a cold page cache over 4 GB of reads inverted the ranking entirely.

## Implementation

`crates/salmon-cli/src/global_alloc.rs` centralizes the selection. Cargo features are additive, so it encodes a fixed priority and resolves any combination to exactly one `#[global_allocator]`. `NAME` reports which one won, so a benchmark labels its own numbers rather than trusting the flags it was invoked with, and the binary logs it at debug level:

```
$ RUST_LOG=debug salmon swim
global allocator: mimalloc
```

`examples/cache_bench.rs` pulls the same module in via `#[path]`, so the microbenchmark and the real binary cannot drift apart on which allocator a given flag set selects.

## Cost to a normal build

Both new dependencies are `optional = true` and off by default, so a stock release build is unchanged. `snmalloc-sys` needs cmake, but only when `--features snmalloc` is passed.

`tikv-jemallocator` is pinned to 0.5 rather than 0.6 because piscem-rs -> sux -> rdst already links `tikv-jemalloc-sys` 0.5, and two versions of a `links = "jemalloc"` native library cannot coexist in one dependency graph.

## Testing

`cargo clippy --release -p salmon-cli --all-targets` clean on all four feature combinations (default, `sysalloc`, `jemalloc`, `snmalloc`); each binary verified to report the allocator it was built with.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -579,6 +579,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c8d4a3bb8b1e0c1050499d1815f5ab16d04f0959b233085fb31653fbfc9d98f9"
 
 [[package]]
+name = "cmake"
+version = "0.1.58"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c0f78a02292a74a88ac736019ab962ece0bc380e3f977bf72e376c5d78ff0678"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "coitrees"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2675,6 +2684,8 @@ dependencies = [
  "salmon-index",
  "salmon-map",
  "salmon-quant",
+ "snmalloc-rs",
+ "tikv-jemallocator",
  "tracing",
  "tracing-subscriber",
 ]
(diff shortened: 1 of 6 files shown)


rob-p:
Thanks Benjamin — the measurement here is genuinely useful, and it answers the question it set out to ask: mimalloc is the right default (system malloc ~38% worse on the mapping phase, jemalloc ~7% behind, snmalloc a tie at +0.4 GiB RSS). The single-threaded-microbenchmark trap and the cold-page-cache contamination you documented are exactly the kind of thing that saves the next person a wasted week, and having the numbers stated in this PR means the question is now answered on the record.

But I'm going to close this rather than merge it, on the machinery rather than the measurement: I don't want to carry alternative allocators in the tree, even feature-gated. The features are only worth their weight if we intend to re-run the comparison periodically or ship non-default allocators to users, and we don't — the measurement's own conclusion is that there is nothing to switch to. What the machinery costs is real, though: a feature-resolution priority scheme that has to stay correct as features combine, a cmake dependency lurking behind one of the flags, a pinned older `tikv-jemalloc-sys` to dodge a `links` conflict with the piscem-rs subtree (a constraint someone would eventually trip over when that stack moves), and one more axis of build configurations for CI and for anyone bisecting. That's ongoing cost for a question we now consider settled.

The PR itself — numbers, protocol, and pitfalls — stays linkable as the durable record, and this close is the "mimalloc, measured, settled" stamp on it. If the question ever genuinely reopens (a new allocator generation, or a workload shift toward long-lived heaps where jemalloc's strengths matter), the branch is the right starting point and we can revisit then.

PR #1131, opened 2026-08-20
feat(quant): report strand-incompatible fragments in lib_format_counts.json

### LINKED ISSUE #1130: Stranded libraries: expose the count of strand-incompatible fragments as a gDNA contamination estimator
## Context

Stranded paired-end libraries (`-l ISR`), ribo-depleted total RNA from human whole blood, quantified against a decoy-aware index (GRCh38 + Ensembl r116). In a subset of samples we identified genomic DNA carry-over from incomplete on-column DNase digestion, measured independently via RSeQC `infer_experiment`: the sense-strand fraction rises from a cohort median of 2.2% to 18.6% in the worst sample, implying up to ~41% double-stranded material in that library (solving `observed = 0.5*g + (1-g)*baseline`).

## What Salmon currently does, and what gets lost

With `-l ISR`, strand-incompatible fragments are excluded from assignment (default `--incompatPrior 0`). This silently removes the half of the gDNA fragments that land on the wrong strand — which is helpful — but:

1. **The other half of the gDNA lands on the correct strand by chance and enters the counts**, indistinguishable from RNA. In our data, genes called differentially expressed downstream correlated strongly with per-sample gDNA estimates (|rho| up to 0.86) until we added a contamination covariate.
2. **The evidence of the filtering is not surfaced.** `lib_format_counts.json` reports `compatible_fragment_ratio: 1.0` and `strand_mapping_bias: 0.0` even for the heavily contaminated sample, because these are computed after/over assigned fragments. The number of fragments *discarded for strand incompatibility* — which would be a direct, free estimator of gDNA contamination — does not appear anywhere in the output as far as we can tell.

## Questions

1. Is there a recommended way to obtain, per sample, the count of fragments discarded due to strand incompatibility with the declared library type? If it exists somewhere in `aux_info`, we could not find it documented.
2. Is there a recommended practice for handling the residual same-strand gDNA fraction in the counts (beyond modeling it as a covariate downstream, which is what we do now)? E.g., is there any merit to an option that estimates the contamination rate from the incompatible-fragment excess and down-weights ambiguous assignments accordingly?

## Feature request (if neither exists)

Expose in `lib_format_counts.json` the raw count of mappings/fragments filtered out for library-type incompatibility, alongside the existing `num_compatible_fragments`. For stranded libraries this single number would let users flag gDNA contamination at quantification time, with zero extra compute — we currently have to run a separate RSeQC pass on a STAR alignment to get it.

Happy to share the sample-level numbers or test a patch.

Setup: salmon 2.5.1 with `-l ISR --seqBias --gcBias`, decoy-aware index (whole GRCh38 primary assembly as decoy), tximport downstream.

BenjaminDEMAILLE:
Closes #1130.

## Problem

On a stranded library, fragments whose every mapping lands on the wrong strand are discarded (default `--incompatPrior 0`), but the count never reached the output. `lib_format_counts.json` hardcoded `compatible_fragment_ratio: 1.0` and `num_compatible_fragments = num_mapped`, so the only way to get the wrong-strand fraction was a second strandedness pass over an alignment (RSeQC `infer_experiment` on a STAR BAM in the reporter's case).

That number is a free contamination estimator. Double-stranded input (genomic DNA carry-over from incomplete DNase digestion) maps to either strand with equal probability, so an excess of incompatible fragments implies a comparable amount of contamination sitting *on* the expected strand, indistinguishable from RNA once assigned. Comparing the ratio across a cohort flags the affected libraries at quantification time.

## Change

Two per-worker counters in the mapping pass (merged into the shared atomics at thread end, like the existing orphan/decoy tallies, so the hot path touches no shared memory), written to `lib_format_counts.json`:

| field | meaning |
|---|---|
| `num_incompatible_fragments` **(new)** | fragments that mapped but had no strand-compatible placement |
| `num_compatible_fragments` | fragments with at least one strand-compatible placement, now measured instead of aliased to `num_mapped` |
| `compatible_fragment_ratio` | the two counts' ratio, instead of a hardcoded `1.0` |

Notes:

* `num_incompatible_fragments` is counted under any `--incompatPrior`. Under the default 0 the fragment is dropped and is therefore not in `num_assigned_fragments`; above 0 it is kept at reduced weight and is.
* `num_compatible_fragments` was wrong with `--incompatPrior > 0`: such fragments count as mapped without being compatible. With the default prior the value is unchanged from before.
* Every other field keeps its C++ name and meaning, so the file stays a drop-in for existing readers. The new key is additive.
* The `--deterministic` mapping pass writes this file and then hands off to a RAD requant that does not rewrite it, so it tallies the same way (from the expected format an explicit `-l` already fixes). Under `-l A` neither path counts the fragments consumed before the library type is inferred, so the two counts can sum to less than the mapped total.

Not addressed here: `strand_mapping_bias` is still `0.0`, and there is no modelling of the residual same-strand contamination (question 2 of the issue). Both are separate pieces of work; this PR is the "zero extra compute" number the issue asks for.

## Testing

`cargo test --workspace` green; `cargo clippy --workspace --all-targets` and `cargo fmt --check` clean.

Two new end-to-end tests in `crates/salmon-quant/tests/quant_end_to_end.rs`, on simulated inward-FR reads:

* `stranded_lib_format_counts_report_incompatible_fragments`: under `ISF` every fragment is compatible (ratio 1.0, count 0); under the opposite `ISR` the new count is the whole library and the ratio collapses to ~0, while `num_assigned_fragments` stays at the (near-zero) mapped total.
* `deterministic_mapping_pass_reports_incompatible_fragments`: the `--deterministic` phase-1 pass reports the same, which would otherwise silently be zero for every deterministic run.

Docs: field semantics and the contamination rationale added to `website/src/content/docs/reference/output-formats.md`.


rob-p (commented):
Thanks Benjamin — the core of this is exactly right, and verified on our side: the per-worker tally merged at thread end costs the hot path nothing (the compatible-placement count is a register increment inside the already-executing filter closure), the empty-placement guards make miscounting unmapped fragments impossible, and on a real unstranded library forced to `-l ISR` the regular path reports `compatible_fragment_ratio: 0.510` — the textbook half-split, with dropped fragments correctly excluded from `num_assigned_fragments`. Tests, clippy, fmt all green here too.

Two gaps keep it from merging as-is, both in mode coverage rather than in the mechanism:

1. **The `--deterministic` flow loses the numbers end-to-end.** The requant *does* rewrite `lib_format_counts.json`: phase 2 (`quantify_rad`, `salmon-align/src/rad.rs`) calls `write_outputs`, whose own LibCounts block (`salmon-align/src/lib.rs:2483`) rewrites the file into the same output directory with a hardcoded `compatible_fragment_ratio: 1.0` — and without the new key. The new deterministic test passes because it drives phase 1 directly; the full CLI flow ends with the clobbered file (confirmed empirically: same subset, `--deterministic -l ISR` finishes at ratio 1.0 with no `num_incompatible_fragments` field).

2. **Alignment-based (`-a`) and RAD-input quant still hardcode 1.0.** Both run the identical `is_compatible` filter with the same `incompat_prior` machinery (`rad.rs:542`, `rad.rs:1271`, and the BAM streaming pass), so the tally is exactly as free there — and after this PR the same JSON field would mean "measured" in reads mode but "fiction" in the other three modes.

Since the fix is the same "count what the filter already decided" pattern you used, we'll push the completion onto this branch rather than volley it back: tallies in the requant/BAM paths, the new field in `salmon-align`'s writer (which makes the phase-2 rewrite *correct* instead of lossy), and the end-to-end test extended to run the real two-phase flow.

One caveat we're deliberately *not* folding in: in sketch mode, paired placements carry `format: None` and auto-pass the strand filter (orientation is known — `is_fw`/`r2_fw` — but not filled into `format`), so the ratio stays ~1 there regardless of strandedness. Making sketch pairs strand-judged would change quantification behavior, not just accounting, so it's follow-up material; we'll note the limitation in the docs.

rob-p:
Pushed the completion as `f9916ffb`, on top of your commit:

- **RAD requant + BAM (`-a`) modes now measure the same tallies** — the per-placement `is_compatible` judgment both paths already made is now tallied per fragment (per-worker counters merged at thread end, same pattern as yours) and `salmon-align`'s `lib_format_counts.json` writer reports the measured values with the identical judged-fragment semantics as the reads-mode writer.
- **This closes the `--deterministic` gap**: phase 2's requant rewrite of the file now writes measured values into the final output directory. Verified on real data — a full `--deterministic -l ISR` run on an unstranded library now ends with `compatible_fragment_ratio: 0.512` / `num_incompatible_fragments: 83142` (regular path: 0.510 / 83317; the small difference is the two passes judging live mappings vs the RAD's stored orientations). Deterministic `quant.sf` is byte-identical to the pre-change build: accounting only.
- **New tests pin exact counts by construction** in both paths: a mixed-orientation RAD through `quantify_rad` and a mixed-orientation SAM through `quantify_alignments` (ISF / ISR / IU each asserting exact `num_compatible` / `num_incompatible` / ratio / assigned). Your phase-1 test stays, re-documented as covering the standalone `--writeRad --skipQuant` file.
- Docs now state the per-mode coverage and a sketch-mode caveat (paired sketch mappings aren't strand-judged at filter time; being looked at separately).

Full workspace suite: 282 passed / 0 failed; clippy and fmt clean. Thanks again — the mechanism and the evidence discipline in the original commit made completing it straightforward.

BenjaminDEMAILLE:
Thanks — and thanks for pushing the completion rather than volleying it back; `f9916ffb` covers both gaps, and I dropped my own version of the same two fixes in favour of it. I'd written the RAD/BAM tallies and a two-phase end-to-end test independently; yours splits the coverage better (a dedicated `rad_lib_format_counts` with exact counts by construction, rather than one test driving both phases), so there was nothing of mine worth keeping there.

One thing I did want to push back on: the sketch caveat. You're right that making sketch pairs strand-judged changes quantification behaviour — but that's true of the *filter*, not of the tally. `2ec60d6b` separates the two.

`is_compatible_for_counting` derives the missing `LibraryFormat` from `is_fw`/`r2_fw` via `observed_paired_format` — the same derivation the deterministic pass already uses to infer a sketch run's library type — and is used **only** by the two per-fragment tallies. `is_compatible` is untouched, so the filter still accepts every sketch pair and a sketch run assigns exactly the fragments it did before. Wherever the observed format is known the two functions agree; a test asserts that exhaustively over every (expected, observed, strand, mate status) combination.

The end-to-end test pins both halves: sketch `-l ISR` on ISF reads reports ≥90% of fragments wrong-strand, and `num_mapped` is identical to the `-l ISF` run.

That combination seems like the more useful output anyway — a low ratio *with* an unchanged mapping rate says "your declared library type is wrong" without also implying fragments were discarded over it. The doc note now says that instead of pointing users at selective alignment for the signal, and keeps the one asymmetry that remains: a `--sketch --writeRad` run requantified from its RAD *does* filter those pairs (the RAD stores both mates' orientations), so the two runs agree on the ratio but need not agree on abundances.

`cargo fmt`, `cargo clippy --workspace --all-targets` and the full workspace suite (31 targets) are green on top of your commit.

rob-p:
Thanks Benjamin — `2ec60d6b` was well-built (the exhaustive filter/tally agreement test in particular is exactly the right discipline), but it was premised on the sketch filter's accept-everything behaviour being intended, and that question got settled separately the same day you wrote it: #1136 ruled it a bug, and #1137 — merged to `develop` a few hours ago — fixes it at the mapping site. Sketch proper pairs now carry their observed format (`observed_paired_format(is_fw, mate_is_fw)`, the same derivation you reached for), so `is_compatible` judges them like every other mode.

The strongest argument for that ruling is actually in your own commit: the caveat that a `--sketch --writeRad` run requantified from its RAD *does* filter those pairs, so the two runs "agree on the ratio but need not agree on abundances." That asymmetry — the same reads through the same mapper giving different abundances depending on whether a RAD sat in the middle — is the bug. #1137 removes it; keeping the filter blind would have institutionalized it.

And the diagnostic you argued for survives mostly intact: the tally judges each fragment *before* the filter drops it, so a wrong `-l` sketch run still reports the low `compatible_fragment_ratio` — now alongside a collapsed mapping rate instead of an unchanged one, which is the same signal selective alignment has always given. One consistent story across modes beats a sketch-only special case.

So I've merged `develop` into this branch and reworked `2ec60d6b` on top (`71d8626c`) rather than reverting it:

- `is_compatible_for_counting` is dropped — post-#1137 every proper pair carries `Some(format)` in every path (SA and sketch fill it, the RAD and BAM paths derive/fill it), so the counting variant could only diverge from the filter in a state that no longer occurs, and a second judgment function that guesses where `is_compatible` deliberately doesn't is drift waiting to happen. The tally goes back to counting what the filter itself decided (one scan, counter inside the filter closure).
- Your sketch end-to-end test stays, expectations flipped to the fixed semantics: wrong-strand count still ≥90% of truth, ratio still ≤0.1, and the dropped fragments now excluded from `num_assigned_fragments` rather than pinned equal to the matching run. It complements #1137's test (that one pins mapping counts; yours pins the JSON against them).
- The "counted, but not filtered on" docs section is replaced by one sentence: every mode measures, every mode judges before filtering.

Full workspace suite 283 passed / 0 failed, clippy and fmt clean. With develop merged in and the modes now consistent, I think this is complete — planning to merge once CI agrees.

PR #1132, opened 2026-08-20
feat(degnorm): degradation normalization — coverage during quant, cohort fit after

BenjaminDEMAILLE:
## What

Degradation normalization, in the spirit of [DegNorm](https://nustatbioinfo.github.io/DegNorm/) (Xiong *et al.*, *Genome Biology* 2019), implemented over salmon's own data:

```sh
# 1. quantify as usual, plus a coverage dump
salmon quant -i index -l A -1 s1_1.fq -2 s1_2.fq --degCoverage -o quants/s1
# ... same for the other samples

# 2. fit across the cohort
salmon degnorm --quants quants/s1 quants/s2 quants/s3 -o degnorm
```

Out come one corrected `quant.sf` per sample plus the model's matrices:

```
degnorm/quants/<sample>/quant.sf   corrected counts, salmon's own format
degnorm/degradation_index.tsv      transcript × sample
degnorm/counts_raw.tsv
degnorm/counts_adjusted.tsv        raw / (1 − DI)
degnorm/degnorm_summary.json
```

The per-sample files are drop-in replacements for the originals, so this is a step inserted ahead of an existing analysis rather than a fork of it — only the path changes:

```r
files <- file.path("degnorm", "quants", samples, "quant.sf")
txi   <- tximport(files, type = "salmon", tx2gene = tx2gene)
dds   <- DESeqDataSetFromTximport(txi, coldata, ~ condition)
```

Each row keeps its `Name`, `Length` and `EffectiveLength` verbatim and in the original order; `NumReads` becomes the adjusted count and `TPM` is recomputed from it (a TPM column left at its old values would contradict the counts beside it). Transcripts the model skipped keep their original count rather than being dropped — a missing row would break the target set tximport expects. Only `quant.sf` is written, not a full output directory, so `tximeta` still needs the original quant dirs.

## Why two stages

Degradation is only identifiable by comparison: one sample's coverage curve is whatever it is, and it takes other samples to establish what it should have looked like. So the streaming half — turning fragment placements into coverage — happens during `quant`, where the placement is already in hand and the read is never visited again; the cross-sample fit happens afterwards over a handful of small dumps. `salmon degnorm` refuses a single sample rather than reporting zeros.

## The model

Per transcript, stack the samples' binned coverage curves into `K` (`m × n`) and fit the closest rank-one matrix that stays **above** the data (degradation only ever removes coverage):

```
minimise ‖K̃ − a bᵀ‖_F   s.t.   K̃ ≥ K,  a ≥ 0,  b ≥ 0
DIᵢ = 1 − (Σ_l K_il) / (Σ_l aᵢ b_l)
```

Block coordinate descent: least-squares rank-one fit, then project with `K̃ ← max(K, a bᵀ)`.

One addition that turned out to be load-bearing: entries that have fallen below `--maskFrac` of the current fit are held out of the *envelope fit* (never out of the index). Without it, damage in a region where only some samples carry evidence gets absorbed into a sagging envelope, and samples that lost nothing pick up an index they have not earned — on a three-sample fixture where one library lost its 5' half, the unmasked fit reports `0.095 / 0.452 / 0.095` where the truth is `0 / 0.5 / 0`; with masking it recovers `0 / 0.5 / 0` exactly. Pinned by a test.

## Implementation

* New crate `salmon-degnorm` (`coverage`, `nmf`, `cohort`), depending only on anyhow/flate2/rayon/serde.
* Accumulation hangs off the same per-fragment hook the bias models use, weighted by the same posteriors. Two deliberate differences: it is **shared** rather than per-thread (a per-worker copy would be a `bins × transcripts` array per thread), and it is **not** stopped at the end of the model-training window — a curve summarising the first few million fragments would describe the head of the FASTQ, not the sample.
* Cost on the hot path: four relaxed `fetch_add`s per compatible mapping. A fragment's span is written as a difference array, so the cost does not grow with how many bins it covers, and the slots are fixed-point `AtomicI64` rather than CAS-looping `AtomicF64` because a highly-expressed transcript has every worker hitting the same handful of bins.
* Memory: `(bins + 1) × transcripts × 8` bytes, ~160 MB for a human transcriptome at the default 100 bins, allocated only under `--degCoverage` and released as soon as the dump is written.
* Zero cost when the flag is off (one `Option` check per fragment).

## Honest about the differences from DegNorm

Same model, different data, and the numbers will not match:

* **transcripts, not genes** — salmon has no genome and no exon-union coverage;
* **posterior-weighted coverage, not primary alignments** — a fragment compatible with several transcripts contributes to each in proportion to its posterior;
* **binned, not per base** (bins are a fixed fraction of each transcript, which is what makes them comparable across samples);
* **per-entry masking, not DegNorm's baseline-selection procedure**;
* **no curve smoothing**.

Also documented, because both bear on how the output should be read: uniform loss is unidentifiable by construction (indistinguishable from lower abundance), and the index has a coverage-dependent positive floor (~0.1 at 20×, ~0.25 at 5× on simulated intact libraries), so it is a comparative measure rather than an absolute percentage. `--minCoverage` / `--minLength` drop the transcripts where the floor dominates.

## Scope

Reads mode only. Alignment mode and RAD input warn and carry on (they produce no per-fragment posteriors); `--deterministic` is rejected for the same reason.

## Tests

* `salmon-degnorm` unit tests: binning exactness (fractional end bins, clipping, mass scaling), dump roundtrip, and the model against known-answer matrices — exact rank one → 0, a half-truncated sample → 0.5, three samples damaged in different places → 0.4 each, monotonicity in severity, uniform loss → 0, degenerate/single-sample/cap/all-lost edge cases, plus the masking regression above.
* `salmon-degnorm`: the emitted `quant.sf` files carry the adjusted counts, a TPM column that renormalises to a million, untouched `Length`/`EffectiveLength`, every input row, and a sample name that cannot escape the output directory.
* `salmon-quant` end-to-end: simulate an intact and a 5'-degraded library of the same transcripts, quantify both with `--degCoverage`, fit the cohort, and check the degraded library is the one flagged, that the adjustment closes the count gap, and that the written `quant.sf` files parse and carry the adjusted counts. A second test checks the dumped curves directly — near-zero over the lost 5' bins, real coverage over the 3' ones.

`cargo fmt`, `cargo clippy --workspace --all-targets` and the full workspace test suite are clean.

## Docs

`docs/degnorm.md` (workflow, model, differences from DegNorm, how to read the numbers, cost, limitations), plus the CLI reference, the flag-status table (new "Rust-port only" status), and `MIGRATION.md`.


rob-p:
Marking this **postponed** rather than closed, and converting to draft — and I want to be explicit that this one is parked because it's *interesting*, not because it's unwanted.

Degradation-aware normalization is a real problem and the design here (cheap coverage accumulation during quant, cohort-level fit afterward) is a sensible decomposition. The hesitation is architectural, not technical: it's unclear the cohort-fit half belongs *inside* salmon rather than as a downstream tool that consumes what salmon emits — even a tightly integrated one, the way tximport or alevin-fry sit beside salmon rather than in it. Salmon's job is per-sample quantification; a cross-sample statistical correction is a different lifecycle (different inputs, different iteration cadence, different users), and folding each such method into the binary is the road to a mono-tool that's harder to maintain and harder to trust. The per-sample half — emitting the coverage summaries a downstream fitter would need — is the part with a natural claim to living in salmon.

So the likely future shape, if we take this up: salmon grows the (small) coverage-emission surface, and the cohort fit lives as its own tool that we'd happily link prominently. Parked with the `postponed` label and retargeted at `develop` so it stays alive for that conversation rather than closed outright.

PR #1133, opened 2026-08-20
Update README with corrected quick start

justbennet:
I added a step 0 to the quick start instructions to extract the sample data.

The prior text did not have the full name of the transcripts file and the fastq files in the sample data are not compressed.

I corrected all three file names, and the examples now seem to run without error.


rob-p:
Thanks!

PR #1139, opened 2026-08-21
Revert 1128 perf/dep divan

rob-p:


rob-p:
Closing as vestigial: this is the auto-generated *revert* of #1128 (it would remove the divan bench from `develop`, not add it). The intent behind it — getting #1128's content onto `develop` after it merged to `master` — was already handled by merging `master` into `develop` (d223e81b), and both branches now carry the bench identically. Nothing to revert, nothing to re-land.

Issue #1140, opened 2026-08-21
Proposal: make --deterministic the default in 2.6.0, deprecate the online one-pass modes


rob-p:
## Proposal

Make `--deterministic` the default quantification mode in **2.6.0**, across selective alignment, sketch, and alignment input. The online one-pass paths stay available for one release behind an explicit flag (e.g. `--onePass`), with a deprecation warning when used; removal would follow in the release after, contingent on no substantive issues surfacing during the deprecation window.

We have generally found the deterministic mode to be faster and at least as accurate, and it adds the property the online path can never have: byte-identical `quant.sf` at any thread count. But "generally found" isn't a basis for flipping a default, so here is the measurement.

## Evidence

**Performance** — 10M real read pairs, GENCODE v49, interleaved A/B reps, same machine:

| config | wall | CPU | peak RSS |
|---|---|---|---|
| SA one-pass `-p 16` | 22.3 s | 293 s | 1.4 GiB |
| SA det `-p 16` | 22.1 s | 281 s | 1.3 GiB |
| SA one-pass `-p 4` | 71.0 s | 264 s | 1.2 GiB |
| SA det `-p 4` | **65.4 s** | 245 s | 1.0 GiB |
| sketch one-pass `-p 16` | 13.3 s | 166 s | 0.6 GiB |
| sketch det `-p 16` | **11.6 s** | **132 s** | 0.8 GiB |
| sketch one-pass `-p 4` | 32.6 s | 120 s | 0.6 GiB |
| sketch det `-p 4` | **29.2 s** | **107 s** | 0.5 GiB |

Deterministic is faster or tied in every cell — up to 13% wall / 20% CPU in sketch — *including* the cost of writing and re-reading the intermediate RAD.

**Accuracy** — polyester sim over GENCODE v49 pc transcripts, ~5M fragments/sample, truth from read names, two samples:

| config | Spearman | log-Pearson | MARD |
|---|---|---|---|
| SA one-pass | 0.9183 / 0.9168 | 0.9598 / 0.9592 | 0.00435 / 0.00438 |
| SA det | 0.9178 / 0.9167 | 0.9595 / 0.9592 | 0.00437 / 0.00439 |
| sketch one-pass | 0.7702 / 0.7703 | 0.9408 / 0.9405 | 0.01412 / 0.01408 |
| sketch det | 0.7704 / 0.7709 | 0.9408 / 0.9407 | 0.01411 / 0.01404 |

[...]

PR #1141, opened 2026-08-21
fix(quant): make each --writeMappings/--writeBam record describe the alignment it reports

BenjaminDEMAILLE:
Split out of #1123 (the third slice of it). This PR carries only the part where the current output **states something false about salmon's own mappings**. The additive fields SAM allows to be absent are in the PR stacked on top of this one, and the genome projection stays in #1123.

`--writeMappings` has always been a diagnostic view of salmon's mappings rather than a general-purpose aligner output, and that is the right scope. No new output mode is added here. The view just stops misdescribing what it shows.

## What is false on `develop` today

| Statement the file makes | Why it is false |
|---|---|
| `CIGAR` = `<readLen>M` (plus a clip at a transcript end), synthesized from the read length | A placement with an indel is written as if it had none, so every base past the gap sits at the wrong reference offset. The `Cigar` type itself holds at most two operations and cannot express an indel. |
| `AS` = the mapper's chain score | Nothing in the record could be checked against it. Auditing `AS` against a recomputed `NM` over a whole output found **2.4% of records self-contradicting**: a `15M1I1D15M1I1D...` sawtooth at `NM:i:80` on a 100 base read, beside an `AS` claiming about three mismatches. Both cannot be true. |
| `MAPQ` = `1`, always | `samtools view -q 255`, the standard "uniquely placed" filter, discards the entire file. A constant that says "ambiguous" for a uniquely placed fragment is not an estimate, it is a wrong one. |

## What this PR changes

Records carry the CIGAR of an alignment actually realized for that placement, with real `I`/`D` operations, the `NM` and `MD` that describe it, the score of that same alignment, and `MAPQ` derived from the placement count on STAR's scale (255 unique, 3, 1, 0), which is what the tools downstream already filter on.

Three separate causes behind the self-contradicting records, each of which alone produces a false one:

- **The anchor search reached only `indel_margin`.** That bounds indel size and says nothing about how wrong a seed-derived position can be: a chain on a repeat or a paralogous stretch can imply a start most of a read away. One offending read belonged 69 bases earlier, where it matched with three mismatches. Out of reach, the banded DP still returned an alignment, the sawtooth above. The search now covers a read length either side.
- **The search gave up when the read did not fit inside the reference at its anchor**, which is exactly the case most worth looking at: an anchor near a transcript end becomes a soft-clipped record that throws bases away when the read often belongs further in and matches in full.
- **ksw2's extension pins the read's first base to the window's first base**, so a realized alignment can grow rightwards but never shift left, and a mismatch near the 5' end left the DP no way to say "the read started earlier" except by opening a leading insertion: `26I34M` at `NM 26` where the answer is `60M` at `NM 1`.

A repaired anchor is a hypothesis, not a verdict: a read with a genuine deletion also looks badly placed to an ungapped scan, so both anchors are realized and the better score wins. An alignment scoring below the mapper's own `min_accepted_score` for the bases it claims is refused, and the record falls back to the synthesized CIGAR with no `NM`/`MD`, which at least says "no base-level alignment here".

Also here, because `NM`/`MD` cannot be honest without it: the index now writes `ambiguous.bin`, the offsets of reference bases indexing had to replace (non-ACGT cannot live in the k-mer structure). Without it `NM` counted a match where the input had an `N` and `MD` named the substitute. 40 bytes for the whole human transcriptome, and **an index built before this file existed loads as "none recorded" and behaves exactly as it did**.

## Cost

Realization runs only for placements that are written, and only when mapping output is on: a run without `--writeMappings`/`--writeBam` never enters `salmon_map::realize`. Two things keep the written path affordable: an affine-scoring bound that proves the ungapped alignment optimal (no DP at all for up to one mismatch at salmon's defaults), and skipping the traceback when the banded optimum equals the ungapped score over the full read.

1M fragments, `-p 8`, order-alternated, best of 3, on 20k unique transcripts and on 5k genes x 4 isoforms (2.5 placements per fragment):

| | `master` | this branch |
|---|---|---|
| no mapping output, single-mapping | 0.89s | 0.89s |
| no mapping output, multi-mapping | 0.94s | 0.92s |
| `--writeBam`, single-mapping | 1.04s | 1.94s |
| `--writeBam`, multi-mapping | 1.36s | 3.28s |

The default path is unchanged. The cost is per written record, so it grows with multi-mapping rather than shrinking.

## Verification

Against samtools 1.24, on the two simulated fixtures, human chr22 (hg38) + GENCODE v44, and real reads (DRR028171, 457366 pairs of 151 bases):

- **`NM`/`MD` match `samtools calmd`** on every record that carries them, including for the transcript containing an `N`.
- **`AS` agrees with `NM` and the CIGAR** on every ungapped record: the score describes the alignment in its own record.
- Every record's CIGAR accounts for exactly as many read bases as `SEQ` holds, pinned by an integration test (`crates/salmon-quant/tests/mapping_output.rs`) alongside the header, the tag set and the single-end path.
- `samtools quickcheck` passes; SAM output round-trips through `samtools view -b`.
- Full workspace suite green.

## Not in this PR

`QUAL` (still `*`), `@SQ M5`/`UR`, `MC`/`MQ`, `ZW`, `@RG`/`--rgLine`, and unmapped records. Those are omissions the format allows, which is a different argument from a CIGAR that does not describe the alignment beside it, so they are in the stacked PR. Genome-coordinate output (`--spliced`) stays in #1123.

### DIFF
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -532,9 +532,10 @@ struct QuantArgs {
     /// Write the names of unmapped fragments to aux_info/unmapped_names.txt.
     #[arg(long = "writeUnmappedNames")]
     write_unmapped_names: bool,
-    /// Write per-mapping SAM records to this file (spoofed CIGAR, like salmon's
-    /// standard `--writeMappings`). `--writeSam` is an accepted alias, naming
-    /// the format explicitly to pair with `--writeBam`.
+    /// Write per-mapping SAM records to this file. Records carry a base-level
+    /// CIGAR with NM and MD, describing the alignment the record reports.
+    /// `--writeSam` is an accepted alias, naming the format explicitly to pair
+    /// with `--writeBam`.
     #[arg(
         short = 'z',
         long = "writeMappings",
@@ -542,7 +543,8 @@ struct QuantArgs {
         conflicts_with = "write_bam"
     )]
     write_mappings: Option<PathBuf>,
-    /// Write per-mapping BAM records to this file. Mutually exclusive with
+    /// Write per-mapping BAM records to this file: the same records as
+    /// --writeMappings, BGZF-compressed. Mutually exclusive with
     /// --writeMappings/--writeSam.
     #[arg(long = "writeBam", conflicts_with = "write_mappings")]
     write_bam: Option<PathBuf>,
@@ -3545,6 +3547,16 @@ mod tests {
         assert_eq!(q.dump_eq_weights, defaults.dump_eq_weights);
         assert_eq!(q.no_length_correction, defaults.no_length_correction);
         assert_eq!(q.no_frag_length_dist, defaults.no_frag_length_dist);
+    }
+
+    /// `--sampleUnaligned` keeps its salmon short form, since it is the flag
+    /// users already have in their scripts.
+    #[test]
+    fn sample_unaligned_has_both_spellings() {
+        for flag in ["--sampleUnaligned", "-u"] {
+            assert!(quant_args(&[flag]).unwrap().sample_unaligned, "{flag}");
+        }
+        assert!(!quant_args(&[]).unwrap().sample_unaligned);
     }
 
     /// `--writeSam` is a spelling of `--writeMappings`, giving SAM output a
--- a/crates/salmon-index/src/lib.rs
+++ b/crates/salmon-index/src/lib.rs
@@ -65,6 +65,18 @@ const INFO_FILE: &str = "info.json";
 const REFSEQ_FILE: &str = "refseq.bin";
 /// Per-reference offsets into `REFSEQ_FILE` (`num_refs + 1` entries).
 const REFSEQ_OFFSETS_FILE: &str = "refseq_offsets.json";
+/// Reference offsets whose base was not ACGT in the input and was replaced to
+/// make the sequence k-mer-able. Pairs of `(tid, offset)`, little-endian `u32`,
+/// sorted, behind a `u64` count.
+///
+/// The index has to align against a cleaned sequence, but a record describing
+/// that alignment should not claim the reference had a base it never had. Keeping
+/// the positions costs one entry per ambiguous base (four across the whole
+/// GENCODE v44 human transcriptome) and lets `NM`, `MD` and the `@SQ M5` speak
+/// about the reference the user actually supplied.
+///
+/// Absent in indexes built before this existed, which load as "none recorded".
+const AMBIGUOUS_FILE: &str = "ambiguous.bin";
 
 /// On-disk salmon index *format* version written into `info.json` by this build.
 /// Distinct from `salmon_version` (the software release string): this is bumped
@@ -387,6 +399,10 @@ struct RefHashes {
 struct PreprocessResult {
     /// non-ACGT bases replaced with pseudo-random ACGT
     replaced: u64,
+    /// per written reference, the offsets whose base was replaced. Keyed by name
+    /// rather than by position because the index assigns its own reference order
+    /// later, and matching on names cannot drift from it.
+    ambiguous: Vec<(Vec<u8>, Vec<u32>)>,
     /// `(retained_name, duplicate_name)` pairs in discovery order. Without
     /// `--keepDuplicates` the duplicate was *not* written to the cleaned FASTA (so
     /// it is absent from the index); with `--keepDuplicates` every transcript is
@@ -462,6 +478,7 @@ fn preprocess_fasta(
             .with_context(|| format!("creating {}", out_path.display()))?,
     );
     let mut replaced = 0u64;
+    let mut ambiguous: Vec<(Vec<u8>, Vec<u32>)> = Vec::new();
     let mut polya_clipped = 0u64;
     let mut polya_dropped: Vec<Vec<u8>> = Vec::new();
     let mut short_decoys_dropped: Vec<Vec<u8>> = Vec::new();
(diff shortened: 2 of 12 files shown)


rob-p:
This is a careful piece of work, and splitting it out of #1123 into "false statements" vs "permitted omissions" was the right decomposition discipline. But I want to push back on the premise before the implementation, because I think the categories come out differently under the contract this output has always had.

**`--writeMappings` is a pseudobam.** It has been, deliberately, since the C++ days: a diagnostic view of *salmon's mappings* — positions, pairing, and the scores quantification actually used — with known fields spoofed or unwritten, for brevity and for computational convenience. The main path does not compute base-level alignments (score-only DP, no traceback), and that is a feature: the compute follows what quantification needs, not what an output format could in principle carry. Under that contract, re-examining your table:

- **The `<readLen>M` CIGAR is a tradeoff, not a falsehood** — a nominal placeholder in a view that never promised base-level alignments. Where I'll concede the point entirely: *we never wrote the contract down.* The output-formats page says nothing about nominal CIGARs, constant MAPQ, or what `AS` means. That's our defect, not the format's — a "pseudoalignment output contract" paragraph is landing with the 2.6.0 docs pass, and with it stated, a reader who treats the CIGAR as real has been warned rather than misled.
- **`AS` is not wrong — it's the most correct value this file can carry.** It is the score quantification used for that placement. Recomputing it against a freshly realized alignment would make the output *disagree with the quant run it documents*, which for a diagnostic view is strictly worse. Your 2.4% "self-contradiction" is a contradiction against the nominal CIGAR — i.e., against the field the contract says is nominal.
- **`MAPQ=1` is the one item where you've found something real** on my criterion (a value that misleads at zero computational justification): deriving MAPQ from the placement count costs nothing — the count is already in hand — and fixes the genuine `samtools view -q` footgun. I'd take *that* as a tiny standalone PR (MAPQ scale + docs), fully decoupled from realization.

**The most valuable thing in this PR may be a question it raises but doesn't answer.** Your three anchor pathologies (seed-derived positions off by most of a read, edge give-ups, ksw2's left-pinning) were found in the realization DP — but the same seeding and chaining feed the *scoring* path that quantification uses. If those pathologies distort scores on the same reads, that's a quantification-accuracy bug and jumps every queue, BAM or no BAM. If they don't (e.g., such placements score below `min_accepted_score` and drop out), then the 2.4% is an artifact of auditing a nominal field. Please open an issue with your fixtures and the sawtooth reads — answering *that* is worth more than any output field.

**Disposition: postponed, not closed** — same standard as #1123/#1129. Full realization (2–2.4× on written output, unconditionally, plus a new index artifact in `ambiguous.bin`) is a real capability some users may genuinely want — but it must be *opt-in*: something like `--writeMappings=nominal|realized`, where `nominal` stays the cheap default contract and `realized` buys NM/MD/real-CIGARs knowingly. That's a design discussion for after 2.6.0 lands — we are mid-flip and not taking on a second behavioral surface — and #1142's additive fields belong to the same discussion. The engineering here (the affine bound proving ungapped optimality, the both-anchors-realized hypothesis testing, the refuse-below-`min_accepted_score` fallback) is exactly what that opt-in mode would be built from, so nothing is wasted by parking it.

Concretely: (1) docs contract lands with 2.6.0; (2) MAPQ fix welcome now as its own small PR; (3) the scoring-path question becomes an issue; (4) this PR and #1142 get the `postponed` label pending the opt-in design.

BenjaminDEMAILLE:
Taking the disposition as written, and conceding the framing on two of the three items.

**The CIGAR.** You are right that a nominal placeholder in a view that never promised base-level alignments is a tradeoff rather than a falsehood, and that the missing piece is the written contract. I will say where I still differ, once, and then drop it: my objection was never that the CIGAR is nominal, it is that `<readLen>M` is *indistinguishable from a real one*. A reader cannot tell it apart without knowing the contract, which until 2.6.0 was unwritten. If the docs pass says so plainly, the field stops being a trap, and that resolves it. `*` would resolve it too and cost nothing, but it would break every reader that tolerates a nominal CIGAR today, so it is not obviously better.

**`AS`.** Conceded, and I had it backwards. It is the score quantification used, and recomputing it against a realized alignment would make the file disagree with the run it documents. For a diagnostic view that is worse, not better. My 2.4% was a contradiction against the nominal field, exactly as you say.

**MAPQ.** #1151, standalone, no realization, no `ambiguous.bin`: the scale, its docs, and a unit test. On the fixture, `-q 255` goes from selecting 0 records to selecting 433027.

**The scoring-path question: #1152, and the answer is no.** I built a fixture against the three pathologies (half the transcripts paralogues, indels in 15% of reads, mismatches concentrated in the first 30 bases) and compared the mapper's per-mate `AS` against a realized alignment at the repaired anchor, joined per record so the units match, then compared the per-fragment weight vectors the EM would see. 99.3% of records agree; 0.24% of multi-placement fragments would rank their placements differently; nothing survives the `exp(-(best - score))` decay into a materially different apportionment. The pathologies live in the realization DP, as you suspected.

The audit did find one real thing, and it is in your category rather than mine: **every right-orphan record reports `AS:i:0`**. The mapper stores an orphan with `score` and `r1_score` both set to the anchor's score, and the writer reports read2's as `score - r1_score`. Quantification is unaffected (it weights by `score`), but 0.47% of records in the file claim a zero-scoring alignment that actually scores 182 to 200. Zero computational justification, one line. I have not touched it, since you scoped #1151 to MAPQ; say the word and it joins that PR or becomes its own.

On the opt-in design for after 2.6.0: `--writeMappings=nominal|realized` is the right shape, and I would rather have the discussion then than argue it now. One thing worth carrying into it, from #1140's measurement: under `--deterministic` the mapping output is byte-different on every run while the record set is identical, so whatever `realized` ends up promising, it should be explicit that it does not promise a reproducible file.

rob-p:
Scoping decision recorded (2026-08-24): the remainder of this PR is deferred to post-2.6.0.

One slice was carved out and taken into 2.6.0 as **#1180** — the right-orphan `AS:i:0` bug you flagged as being in our category. Confirmed in full (717,526 right-orphan records at `AS:i:0`, both SAM and BAM, `quant.sf` byte-identical before/after) and fixed as its own surgical PR. Thanks for the precise diagnosis — the `score - r1_score` collapse was exactly it.

The rest here (CIGAR synthesized from read length, `AS`-vs-`NM` self-consistency) needs the design work it always did and is deferred to post-2.6.0, tracked in #1179. The MAPQ slice already landed via #1151 and #1171. Not closing this PR — it remains the home for the deferred work.

PR #1144, opened 2026-08-22
fix(cli): forward the inference knobs --deterministic was dropping (#1140 PR A)

BenjaminDEMAILLE:
PR A of #1140. `run_deterministic` hand-copies the reads-mode request into the `AlignQuantOptions` the requant pass takes, and three knobs never made the trip: the flag was accepted, the run succeeded, and nothing happened.

## What was dropped

| flag | effect it should have | status before |
|---|---|---|
| `--biasSpeedSamp` | fragment-length stride of the GC effective-length convolution | forwarded nowhere; `AlignQuantOptions` default (5) used |
| `--noBiasLengthThreshold` | accept a bias-corrected effective length outright instead of flooring it at the lower barrier | same |
| `--initUniform` | start the EM from a uniform abundance vector | `QuantOptions` had no such field at all, so reads mode never carried it in either path |

`--initUniform` is now a `QuantOptions` field, per your note that the flip will want it there anyway.

## The seam, since the block was the root cause

The forwarding is now `requant_options(&map_opts, rad_path, out_dir, bias_targets) -> AlignQuantOptions`, with two unit tests: one sets every forwarded knob to a value different from its `AlignQuantOptions` default and asserts each one arrives, the other asserts an unset request leaves them at the defaults, so the first test is measuring forwarding rather than a coincidence. A field dropped from that mapping now fails a test instead of failing silently at runtime.

`ref_seqs` stays with the caller: it reads the index from disk, and only the caller knows whether bias correction asked for it.

## End-to-end, and what it can honestly show

`--biasSpeedSamp 25` with `--gcBias` under `--deterministic`, 3000-transcript fixture, 300k pairs:

```
before   quant.sf md5 57985fee3ebe9037e05071f3789efc06   (identical with and without the flag)
after    quant.sf md5 d5058db964800db47fa0146ddba4d74a   (differs from the no-flag run)
```

The other two produce no observable diff on that fixture, and I would rather say why than quietly omit them:

- `--noBiasLengthThreshold` removes a floor that only binds when a bias-corrected length falls below the uncorrected one. That never happens on this fixture, so the corrected lengths are the same either way.
- `--initUniform` changes where the EM starts, and it converges to the same optimum from either start on a well-conditioned input.

Both are forwarded and covered by the unit test. Neither absence of a diff is evidence that they are not.

## Out of scope, deliberately

`--initUniform` still does nothing in one-pass reads mode: there is no initialisation step there, since the online EM starts from the abundances it accumulated while mapping. Per your scoping this PR does not invent behaviour for a path headed for deprecation; it only stops the flag from being silently inert, with a warning naming where the flag does apply. Say the word if you would rather drop even the warning from this PR and let the flip handle it.

### DIFF
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -1085,8 +1085,30 @@ fn write_gene_level(
 /// passed silently did nothing. That is how `--biasSpeedSamp` and
 /// `--noBiasLengthThreshold` came to be ignored under `--deterministic`.
 ///
-/// `ref_seqs` is deliberately not set here; it needs to read the index from disk
-/// and only the caller knows whether bias correction asked for it.
+/// # Deliberately not forwarded
+///
+/// The point of the function is to make "audited and correctly omitted"
+/// distinguishable from "forgotten", which a test cannot do for an absence. Every
+/// field of `AlignQuantOptions` this leaves alone is here, with why:
+///
+/// * `ref_seqs`: needs to read the index from disk, and only the caller knows
+///   whether bias correction asked for it. Set immediately after this returns.
+/// * `explicit_fld_args`, `fld_policy`: phase 1 already folded the user's FLD
+///   prior into the bake. Forwarding them would have phase 2 warn that a baked
+///   FLD overrides the run's own `--fldMean`, against a bake that came from that
+///   very argument.
+/// * `forgetting_factor`, `num_aux_model_samples`, `num_pre_aux_model_samples`:
+///   online-inference knobs. Phase 2 has no online phase to apply them to.
+/// * `num_error_bins`, `discard_orphans`, `no_error_model`,
+///   `deterministic_error_model`: BAM-side knobs. Phase 2 reads a RAD whose
+///   placements were already scored and filtered during mapping.
+/// * `bam`, `output_dir`, `progress`: set by the caller, which owns the paths and
+///   the run's lifecycle.
+///
+/// One absence here is a bug rather than a decision: `skip_quant` is set on
+/// phase 1 unconditionally (its job is to map and write the RAD) and never
+/// reaches phase 2, so `--skipQuant --deterministic` quantifies anyway. Being
+/// able to see that from this list is the argument for the list.
 fn requant_options(
     map_opts: &QuantOptions,
     rad_path: &std::path::Path,
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -1076,6 +1076,52 @@ fn write_gene_level(
     Ok(())
 }
 
+/// The phase-2 options for a deterministic reads-mode run: every knob from the
+/// reads-mode request that the RAD quantifier can honour, carried across.
+///
+/// This is a separate function so the forwarding can be asserted field by field
+/// in a test. A knob that is simply forgotten here does not fail loudly: the
+/// second pass runs with the default, the run succeeds, and the flag the user
+/// passed silently did nothing. That is how `--biasSpeedSamp` and
+/// `--noBiasLengthThreshold` came to be ignored under `--deterministic`.
+///
+/// `ref_seqs` is deliberately not set here; it needs to read the index from disk
+/// and only the caller knows whether bias correction asked for it.
+fn requant_options(
+    map_opts: &QuantOptions,
+    rad_path: &std::path::Path,
+    out_dir: &std::path::Path,
+    bias_targets: Option<PathBuf>,
+) -> AlignQuantOptions {
+    let mut q = AlignQuantOptions::new(rad_path.to_path_buf(), out_dir.to_path_buf());
+    q.lib_type = map_opts.lib_type.clone();
+    q.em = map_opts.em.clone();
+    q.range_factorization_bins = map_opts.range_factorization_bins;
+    q.transcripts = bias_targets;
+    q.seq_bias = map_opts.seq_bias;
+    q.gc_bias = map_opts.gc_bias;
+    q.pos_bias = map_opts.pos_bias;
+    q.bias_seed_em_iters = map_opts.bias_seed_em_iters;
+    q.score_exp = map_opts.map_config.score.score_exp;
+    q.incompat_prior = map_opts.incompat_prior;
+    q.sig_digits = map_opts.sig_digits;
+    q.fld_mean = map_opts.fld_mean;
+    q.fld_sd = map_opts.fld_sd;
+    q.fld_max = map_opts.fld_max;
+    q.gc_bins = map_opts.gc_bins;
+    q.cond_gc_bins = map_opts.cond_gc_bins;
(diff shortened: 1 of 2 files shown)


rob-p (approved):
Verified end to end on real data (200k pairs, GENCODE v49): `--biasSpeedSamp 25` under `--deterministic --gcBias` now changes `quant.sf` where it previously didn't, the two forwarding tests do what they claim (the defaults-stay-default companion is the right guard against the first test passing by coincidence), fmt/clippy clean. The seam is exactly the durable fix — the hand-copied block was unreviewable, and a function with a field-by-field test isn't. Approved; three notes, none blocking merge.

**1. Your seam immediately caught another one.** Auditing what `requant_options` *doesn't* forward: `--skipQuant --deterministic` silently ignores `--skipQuant` — `run_quant` sets it before the deterministic branch, `run_deterministic` unconditionally overwrites it for phase 1, never forwards it to phase 2, and the run writes a full `quant.sf` despite the flag (confirmed empirically on this branch). Same class as the three you fixed. Don't grow this PR for it — but it needs an owner. Suggested semantics for the follow-up: `--skipQuant --deterministic` = phase 1 only, keep the RAD as the deliverable, which is exactly what `-a --deterministic --skipQuant` already does (its keep-condition includes `opts.skip_quant`). If you want it, it's yours; otherwise it goes on the flip PR's checklist.

**2. Document the deliberate omissions where the test can't see them.** The doc comment covers `ref_seqs`; the other absences deserve a sentence each, because the next reader can't tell "audited and correctly omitted" from "forgotten" — and this function's whole reason to exist is making that distinction visible:
- `explicit_fld_args` / `fld_policy`: correctly omitted — phase 1 already folded the user's FLD prior into the bake, so forwarding them would make phase 2 emit a bogus "baked FLD overrides your `--fldMean`" warning against the run's own bake.
- `forgetting_factor`, `num_aux_model_samples`, `num_pre_aux_model_samples`, `num_error_bins`, `discard_orphans`: online-/BAM-side knobs with no phase-2 meaning (orphan discarding already happened during mapping).
- `skip_quant`: see note 1.

A short "deliberately not forwarded" block listing these turns the function into the complete audit record.

**3. Merge coordination**: #1143/#1146 edit the same `run_deterministic` region (they append `prior_seconds`/`external_start_time` after your extracted block, and #1146 rewires the `rad_path` selection above it). Whichever lands second has a small textual conflict; when you rebase, `prior_seconds`/`external_start_time` belong *inside* `requant_options`' remit eventually but must keep being set by the caller (they're computed from the caller's clock), so the right resolution is your function plus the two caller-side assignments after it — and they should then join the forwarding test.

BenjaminDEMAILLE:
Thanks. All three notes handled.

**1.** Took `--skipQuant`: #1149, phase 1 only with the RAD kept, matching what `-a --deterministic --skipQuant` already did. Test drives the binary and asserts both directions.

**2.** Pushed the audit block. Every `AlignQuantOptions` field this function leaves alone is now listed with why: `ref_seqs` (caller reads the index), `explicit_fld_args`/`fld_policy` (phase 1 already folded the prior into the bake, forwarding them would warn about the run's own bake), the online knobs, the BAM-side knobs. `skip_quant` is in the list too, marked as the one absence that was a bug rather than a decision, with the pointer to #1149. That was the argument for the list, so it seemed worth showing it earning its keep immediately.

**3.** Understood on `prior_seconds`/`external_start_time`: caller-computed, so caller-assigned after the call, and they join the forwarding test when I rebase. Happy to do that rebase whenever #1143/#1146 land.

rob-p:
Pushed the develop-conflict resolution onto this branch (`742495b6`) rather than volleying it back, since the conflict was against our #1143/#1146 merges: `requant_options` stays the single source of forwarded knobs, develop's caller-side additions (the `prior_seconds`/`external_start_time` timing handoff and the cleanup-wrapped phase 2) sit on top, and develop's now-redundant hand-copied field block is gone. The two timing fields join your "deliberately not forwarded" list — they're computed from the caller's clock after phase 1, so they can't be part of a mapping built before it starts — and the `skip_quant` entry now records that #1149 fixes it.

The same push also settles the `--initUniform` question that came out of reviewing this PR, and the answer surprised us: **in this port, the flag is already honoured in both one-pass and `--deterministic` reads mode — by construction.** The offline EM starts uniform in every reads-mode path (`optimize_packed_with_init(..., None, ...)`); the C++ online-blended warm start was deliberately not carried over (measured output-neutral at convergence, and dropping it removed an order-dependent step — the `--meta` preset comment has recorded this fact all along). The RAD quantifier likewise always starts uniform; its baked `initial_abundances` feed only the bias expected model, never the EM seed. The one place the flag toggles anything is online alignment mode (`-a`), whose EM warm-starts from the online estimates — and that path honoured it before this PR.

So the defect was never behavior — it was *documentation claiming the opposite of the code*, in both directions: the one-pass warning said the flag "applies to `--deterministic`" (where it's vacuous) and couldn't apply in one-pass (where it's the permanent behavior). The push removes the warning and rewrites the flag's help text and both options docs to state the per-path truth. Your forwarding of `init_uniform` through `requant_options` stays: it's what makes the flag *stay* honoured if the RAD path ever grows a warm start.

Workspace suite on the branch: 295 passed / 0 failed; clippy/fmt clean. Mergeable again.

BenjaminDEMAILLE:
Thanks for resolving it on the branch rather than volleying it back, and for digging into `--initUniform` far enough to find that the defect was the documentation.

I checked the claim rather than taking it, since my warning asserted the opposite: reads mode calls `optimize_packed_with_init(&packed, …, None, …)` at both sites (`salmon-quant/src/lib.rs:986,988`), the RAD quantifier likewise at `salmon-align/src/rad.rs:2036,2109`, and the only warm start in reads mode is the bias burn-in continuing its own alphas (`lib.rs:1196`), which is a within-run continuation rather than an initialisation the flag could toggle. The single place `init_uniform` branches on anything is online `-a` (`salmon-align/src/lib.rs:2335`). So the flag has been honoured in reads mode all along, by construction, and my warning was wrong in both halves. Your rewrite is right and the forwarding still earns its place as the thing that keeps it true if the RAD path ever grows a warm start.

Worth noting what this says about the exercise: the seam surfaced two real defects (`--skipQuant`, now #1149) and one documentation defect that had been asserting the opposite of the code in two places. The false-positive rate of "list every field and say why" is apparently zero so far.

Rebases done on my side, both onto the new `develop` with your #1143/#1146 in it: **#1147** (dump-eq) and **#1148** (RAD length knobs) are pushed and green, and **#1149** is retargeted from `feat/det-disk-story` back to `develop` now that it has merged. Only conflict in either was the `AlignQuantOptions` field block against `prior_seconds`/`external_start_time`, resolved by keeping both.

The consolidation duty is still mine and still owed: the moment this PR lands, #1147's two dump flags and #1148's two length knobs move into `requant_options` and its field-by-field test, and come off the "deliberately not forwarded" list. I cannot stack them on this branch in the meantime (a fork branch cannot be a base), so they sit on `develop` until then.

PR #1145, opened 2026-08-22
feat(align): warn when a BAM has no usable AS to score by (#1140 PR D)

BenjaminDEMAILLE:
PR D of #1140. Deterministic alignment mode without `--errorModel` weights each placement by the BAM's `AS`. A missing tag parses as score 0, indistinguishable from a genuine zero, so a BAM with no `AS` leaves every placement of a fragment tied and the EM apportions multireads by effective length alone. Same exit code, same output shape, quietly worse numbers.

## The cost of staying silent

Same BAM, once as produced and once with the tags stripped, against the truth in the read names (3000-transcript fixture, 300k pairs):

| input | mode | Spearman | MARD |
|---|---|---|---|
| with `AS` | det | 0.92081 | 0.02294 |
| no `AS` | det | **0.76235** | **0.06890** |
| no `AS` | det `--errorModel -t` | 0.90438 | 0.02443 |

Three times the MARD, and the remedy already exists.

## What this adds

Presence is counted where the tag is read, per your note that `record_to_frag` maps a missing `AS` to 0: by the time scores exist, absent and zero are the same number. Each worker tallies `(records, records_with_AS, min, max)`; the tallies merge alongside the existing summary reduction.

Three states draw a complaint, each naming `--errorModel -t` as the way out:

- no record carries `AS`
- some records carry it and some do not, so zeros compete with real scores
- every record reports the same value

The third is why the count is per record rather than per placement, as you suggested optionally: a proper pair sums to twice an orphan's score, so a BAM whose every record says `AS:i:100` looks varied if you judge the sums.

Observed on the three fixtures:

```
with_as   (no warning)
no_as     no alignment in this BAM carries an AS tag (890098 records read): scoring falls back to
          equal weights for every placement of a fragment, so multireads are apportioned by
          effective length alone; pass --errorModel with -t <transcripts.fa> to score the
          alignments from their own bases instead
const_as  every alignment in this BAM reports the same AS (100): [same advice]
```

## Scope

The error-model path never reads `AS`, so it says nothing (structurally: the tally only exists in the branch that scores by the tag). Genome projection goes through `project_genome_bam_to_rad` and scores by bramble similarities, so it is untouched, per your exemption.

## Testing

The verdict is returned on `AlignRadSummary` as well as logged, so the tests assert on a value rather than capturing log output. Five cases in `crates/salmon-align/tests/deterministic_bam.rs`: absent, constant, partial, varying (silent), and the error-model path (silent). The existing fixture in that file carries no `AS` at all, which is how the absent case gets a realistic input for free.

### DIFF
--- a/crates/salmon-align/src/bam_rad.rs
+++ b/crates/salmon-align/src/bam_rad.rs
@@ -72,6 +72,13 @@ const ERROR_MODEL_ALPHA: f64 = 1.0;
 pub struct AlignRadSummary {
     pub num_processed: u64,
     pub num_mapped: u64,
+    /// What was wrong with the BAM's `AS` tags, when the run scored by them and
+    /// they could not carry that weight. `None` when the tags were usable, or
+    /// when `--errorModel` scored the alignments instead.
+    ///
+    /// Returned as well as logged so a caller (and a test) can see the judgement
+    /// without capturing log output.
+    pub score_tag_warning: Option<String>,
 }
 
 /// Project a name-grouped transcriptome BAM into a fully-baked salmon RAD at
@@ -186,7 +193,22 @@ pub fn write_alignment_rad(
         let sums = stream_grouped(&opts.bam, nthreads, || {
             WriteWorker::new(&writer, &fld, naive.as_ref())
         })?;
-        reduce_summary(sums)
+        // The scores these records carry are only as good as the tags they came
+        // from, and a BAM without them fails silently: say so before the numbers
+        // are used for anything.
+        let mut tags = ScoreTagTally::new();
+        let mut plain = Vec::with_capacity(sums.len());
+        for (processed, mapped, worker_tags) in sums {
+            tags.merge(&worker_tags);
+            plain.push((processed, mapped));
+        }
+        let complaint = tags.warning();
+        if let Some(text) = &complaint {
+            tracing::warn!("{text}");
+        }
+        let mut summary = reduce_summary(plain);
+        summary.score_tag_warning = complaint;
+        summary
     };
 
     // ---- end-of-pass bake (single quant pass) ----------------------------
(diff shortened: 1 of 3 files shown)


rob-p (approved):
Reviewed on the branch: all 8 tests pass, clippy/fmt clean, and the design choices hold up to poking. Counting presence at the single point where the tag is parsed (rather than inferring from scores) is the only correct structure given the `unwrap_or(0)`; the per-record (not per-placement-sum) constancy judgement is right and the pair-sums-to-200 rationale in the test doc is exactly the trap it avoids; the three-state warning covers the partial case, which is genuinely worse than absent and easy to have missed. Returning the verdict on `AlignRadSummary` so tests assert on a value instead of scraping logs is the kind of testability decision that keeps paying. Approved; two notes, neither blocking.

**1. One wording consideration for the constant-`AS` case.** There is a legitimate way to hit it: error-free reads of uniform length through an end-to-end aligner produce genuinely identical `AS` on every record — clean simulation data does this routinely. The warning's *content* is still true (a constant score column carries zero information, whatever its provenance), and the advice still applies, so this is not a false positive in substance — but the current phrasing reads as "your BAM is defective." A clause like "this can be legitimate for error-free reads of uniform length, but it means the scores cannot distinguish placements" would keep sim users from doubting their aligner. Take or leave.

**2. Follow-up worth considering: make the verdict machine-readable in `meta_info.json`.** Right now the warning reaches humans (log) and tests (`AlignRadSummary`), but a pipeline that only keeps the output directory can't see it — and `meta_info.json` already has the `diagnostics` array built for exactly this. The wrinkle is that the verdict is born in phase 1 while `meta_info.json` is written by phase 2, so it needs a hand-off (an `extra_diagnostics` on `AlignQuantOptions` that `quantify_rad` appends, or similar). That dovetails with the deprecation-mechanics work on #1140 rather than belonging here — flagging it so it lands on that checklist instead of getting lost.

BenjaminDEMAILLE:
Took note 1, pushed. The constant-`AS` message now reads:

> every alignment in this BAM reports the same AS (100), which can be legitimate for error-free reads of uniform length, but means the scores cannot distinguish one placement from another: scoring falls back to equal weights [...]

You are right that it is not a false positive in substance, and a simulation user should not come away doubting their aligner.

Note 2 (the verdict into `meta_info.json`'s `diagnostics` array) I have left alone: the phase-1-to-phase-2 hand-off makes it yours on the deprecation-mechanics work rather than mine here, and it is on the #1140 list now.

PR #1147, opened 2026-08-22
feat: dump equivalence classes from the RAD path, not before they exist (#1140 PR C)

BenjaminDEMAILLE:
PR C of #1140. `--dumpEq` under `--deterministic` wrote a well-formed `eq_classes.txt.gz` declaring **zero** classes: the mapping pass wrote it, and at that point there are none, because the classes are assembled by the quantification pass from the RAD. Nothing overwrote it afterwards, so the run produced a file whose header says `2996` transcripts and `0` classes beside a `quant.sf` derived from 4469 of them. Worse than a missing file, since nothing about it looks wrong.

## Shape, as you suggested

The serializer moves to `salmon-eqclass` as `write_eq_classes(dir, names: &[String], collapsed, with_weights)`, beside the `CollapsedEqClasses` it writes. It could not be shared before because it took a `SalmonIndex`, which the RAD path does not have; taking the names is all it needed. Both callers are now thin, and the two paths cannot drift into writing different bytes for the same classes. `flate2` as a direct dependency of `salmon-eqclass` rather than routing through `salmon-core::compress`, since the format is line-oriented text straight into a `GzEncoder`.

- `quantify_rad` dumps after collapsing, where the classes the EM consumes exist.
- `run_deterministic` takes the two flags off phase 1 and sets them on phase 2, so phase 1 also stops forcing `build_eq` to fill a builder nothing reads.
- `-a`, `--rad` and genome projection gain `--dumpEq`/`--dumpEqWeights`: same quantifier, the flags were simply never wired to it.

## Measured

Same 300k-fragment input, `--dumpEq`, before and after:

```
before   one-pass  header "2996 4468"      det  header "2996 0"      (5599 bytes of nothing)
after    one-pass  header "2996 4468"      det  header "2996 4469"
```

The one-class difference between the modes is real and not a defect of either: the fixed FLD moves a weight across the acceptance boundary. In both modes the declared count equals the number of class lines, and the per-class counts sum to the mapped-fragment total from `meta_info.json`.

## Testing

Rather than assert the two modes produce identical class sets (they do not, see above), the end-to-end test asserts the invariants that must hold in each mode: the declared count is non-zero, it equals the number of lines, the counts account for every mapped fragment, and the transcript-name block is the index's. It drives the built binary, because the bug was in how the two passes were wired rather than in the writer.

A unit test in `salmon-eqclass` pins the two output shapes: collapsed by transcript set without weights (two range-factorized sub-classes over the same set merging into one line, counts summed), verbatim with them.

## Overlap warning

This touches `run_deterministic` in `crates/salmon-cli/src/main.rs`, which #1146 rewrites substantially. The overlap is four lines (taking `dump_eq`/`dump_eq_weights` off `map_opts`, setting them on the phase-2 options), so whichever lands second is a trivial rebase. Happy to rebase this onto #1146 rather than the other way round, since yours is the larger change there. Same applies to #1144, which extracts the phase-2 option building into a named function.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2734,6 +2734,7 @@ version = "2.5.1"
 dependencies = [
  "anyhow",
  "clap",
+ "flate2",
  "indicatif",
  "mimalloc",
  "piscem-rs",
@@ -2768,7 +2769,9 @@ name = "salmon-eqclass"
 version = "2.5.1"
 dependencies = [
  "dashmap",
+ "flate2",
  "salmon-core",
+ "tempfile",
  "xxhash-rust",
 ]
 
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -198,6 +198,12 @@ pub struct AlignQuantOptions {
     /// start time (asctime) of that earlier phase; when set, reported as the
     /// run's `start_time` instead of this pass's own
     pub external_start_time: Option<String>,
+    /// `--dumpEq`: write `aux_info/eq_classes.txt.gz`, the equivalence classes
+    /// the EM consumed, collapsed by transcript set.
+    pub dump_eq: bool,
+    /// `--dumpEqWeights`: the same file with each class's per-transcript
+    /// combined weights, and without collapsing range-factorized sub-classes.
+    pub dump_eq_weights: bool,
     /// significant digits for the EffectiveLength and NumReads columns of
     /// `quant.sf` (`--sigDigits`, salmon default 3)
     pub sig_digits: u32,
@@ -265,6 +271,8 @@ impl AlignQuantOptions {
             init_uniform: false,
             prior_seconds: 0.0,
             external_start_time: None,
+            dump_eq: false,
+            dump_eq_weights: false,
             sig_digits: 3,
             num_error_bins: 4,
             discard_orphans: false,
--- a/crates/salmon-align/src/rad.rs
+++ b/crates/salmon-align/src/rad.rs
@@ -1974,6 +1974,20 @@ pub fn quantify_rad(opts: &AlignQuantOptions, rad_path: &Path) -> Result<AlignQu
     // alignment input; analogous to "mapping" in the reads driver).
     timer.mark("rad_read");
 
+    // The classes as the EM will consume them. Written from here rather than
+    // from the mapping pass because this is the only place they exist: a
+    // deterministic run's first pass maps and writes the RAD without ever
+    // assembling a class (COMBINE-lab/salmon#1140).
+    if opts.dump_eq || opts.dump_eq_weights {
+        salmon_eqclass::write_eq_classes(
+            &opts.output_dir,
+            &names,
+            &collapsed,
+            opts.dump_eq_weights,
+        )
+        .context("writing eq_classes.txt.gz")?;
+    }
+
     // Deterministic EM from a uniform start. With a fixed FLD the collapsed
     // classes are independent of fragment/worker order, so the EM fixed point is
     // reproducible; there is no online warm-start (it was proven not to affect the
--- a/crates/salmon-cli/Cargo.toml
+++ b/crates/salmon-cli/Cargo.toml
@@ -33,6 +33,7 @@ sysalloc = []
 
 [dev-dependencies]
 tempfile = { workspace = true }
+flate2 = { workspace = true }
 
 [lints]
 workspace = true
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -1156,6 +1156,12 @@ fn requant_options(
     q.bias_speed_samp = map_opts.bias_speed_samp;
     q.no_bias_length_threshold = map_opts.no_bias_length_threshold;
     q.init_uniform = map_opts.init_uniform;
+    // Phase 2 is where the equivalence classes exist, so it is where they are
+    // dumped. Phase 1 has none to dump (COMBINE-lab/salmon#1140), which is why
+    // this mapping has to be built from the request as the user made it, before
+    // the driver clears the flags for phase 1.
+    q.dump_eq = map_opts.dump_eq;
+    q.dump_eq_weights = map_opts.dump_eq_weights;
     q
 }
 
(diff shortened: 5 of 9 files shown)


rob-p (approved):
Reviewed on the branch: all tests pass (the new e2e, the shape-pinning unit test, and the full crates involved), clippy/fmt clean, and the structure is exactly what was asked — the serializer is a faithful *move* (same format strings, same `BTreeMap` collapse, names in index order), so reads-mode output is byte-compatible by construction, and the shared home makes divergence impossible. Verified the wiring corners too: `--rad --skipQuant --dumpEq` dumps correctly (the collapse precedes the skip-quant early-out), and phase 1 no longer pays for `build_eq` it never reads. The e2e's decision to assert per-mode invariants (declared == listed, counts sum to `num_mapped`) rather than cross-mode identity — and to document *why* the modes legitimately differ by one class — is the right call.

**One fix before this merges — a misattached doc comment.** `write_eq_classes` was inserted between `NAIVE_NO_FMT`'s doc comment and the constant itself, so rustdoc now renders:

- `write_eq_classes` with a doc that *opens* with "A sentinel value is used instead of `Option<u8>` because these placements are stored in the millions…" — two orphaned lines from the constant's doc, followed by your actual text;
- `NAIVE_NO_FMT` (and the preceding sentinel rationale) stripped of the doc that explained it.

Moving the function below the `NAIVE_NO_FMT` block (or above the whole sentinel doc) restores both. Two-minute fix; everything else is ready and we'll merge as soon as it's pushed.

Two take-or-leave notes:
- The `GzEncoder` wraps the `File` directly; a `BufWriter` between them would batch the many small `write!` calls per class line. Same as the code it moved, so not a regression — just cheap if you're touching the function anyway.
- Merge-order coordination stands as you described: this touches the same `run_deterministic` lines as #1144/#1146; land in whatever order, the rebases are trivial, and once #1144's `requant_options` exists these two flags belong in it (and in its forwarding test).

BenjaminDEMAILLE:
Fixed and pushed: `write_eq_classes` now sits below `NAIVE_NO_FMT`, so the sentinel's rationale documents the sentinel again and the writer's doc opens with its own text. Took the `BufWriter` note too, since I was in the function: one `write!` per field per class was a lot of small writes reaching the encoder. Behaviour unchanged, tests still green.

On merge order: agreed, and once #1144's `requant_options` is in, these two flags belong inside it and in its forwarding test. I will fold them there in whichever rebase comes second.

rob-p:
Merge train update: #1144/#1145/#1149 are on `develop`, #1148 goes next, and this PR lands last — deliberately, because its final rebase is the one with all interacting pieces in view. Three things for it:

**1. The consolidation has a design wrinkle worth solving cleanly.** Your current pattern `mem::take`s the dump flags off `map_opts` *before* phase 1 (correctly — phase 1 must not pay for `build_eq` or write the empty file), but `requant_options(&map_opts, …)` reads `map_opts` *after* that mutation, so a naive "move the fields into `requant_options`" would forward the cleared values. The clean resolution: **hoist the `requant_options` call above the phase-1 flag mutations** — it's a pure function of the original request, and `rad_path`/`bias_targets` are already known at that point — then clear `map_opts.dump_eq`/`dump_eq_weights` for phase 1 as plain assignments (the take-locals dance disappears). The fields then join the forwarding test like any other, and the function keeps its meaning: "the phase-2 options, from the request as the user made it."

**2. A new cross-PR interaction is now yours to close**, since this rebase is where both halves meet: with #1149 in, `--deterministic --skipQuant --dumpEq` stops after phase 1 — and with this PR's (correct) phase-1 suppression, *nobody* dumps: the run silently produces no `eq_classes.txt.gz` at all, while one-pass `--skipQuant --dumpEq` produces a real one. Silently dropping a flag is the exact pattern this batch exists to eradicate, so add a warning next to #1149's early return, along the lines of: "`--dumpEq` with `--skipQuant --deterministic`: no quantification pass runs, so there are no equivalence classes to dump; quantify the kept RAD with `--rad --dumpEq` to produce them." A line in your e2e (or #1149's) pinning the warning-and-no-file behavior would seal it.

**3. Housekeeping**: the two dump flags come off nothing (they were never on the omissions list) — just don't add them there.

Everything else rebases clean.

BenjaminDEMAILLE:
All three done, rebased on the current `develop` (with #1144/#1145/#1149 in it; one more trivial rebase due once #1148 lands).

**1. The hoist.** You were right that a naive move would have forwarded the cleared values. `requant_options` is now called *before* phase 1 touches anything, so it reads the request as the user made it, and the `mem::take` dance is gone: phase 1's suppression is two plain assignments. The function's doc says why the ordering matters, since it is the kind of thing a later refactor would happily undo. The two dump flags now live inside it and in both tests.

**2. The interaction.** Confirmed before fixing: with #1149 in, `--deterministic --skipQuant --dumpEq` produced no file and said nothing, while one-pass `--skipQuant --dumpEq` produces a real one. Warning added next to #1149's early return:

```
--dumpEq/--dumpEqWeights has nothing to write under --skipQuant --deterministic: equivalence
classes are built by the quantification pass, which --skipQuant skips. Quantify the RAD that
was kept with `--rad <path> --dumpEq` to produce them.
```

It names the actual RAD path rather than a placeholder, since the run already knows it. The e2e pins all of it: no `eq_classes.txt.gz`, the warning present, `--rad` named in it, and the RAD it points at on disk.

**3.** Nothing added to the omissions list.

Verified on the rebased branch: `--deterministic --dumpEq` declares `2996 4469` classes as before, workspace suite green (all crates), clippy/fmt clean.

One observation worth recording, since it is the third of its kind in this batch: the failure mode was again a flag silently doing nothing, and again it appeared only where two correct changes met. #1149 is right to stop after phase 1, this PR is right to keep phase 1 from writing an empty dump, and the gap between them belonged to neither. The audit list and the seam catch the single-PR cases; the cross-PR ones seem to need exactly what you did here, which is someone sequencing the merges with all the halves in view.

rob-p:
Verified the rebase on the branch — all three items landed, and well:

- The **hoist** is the right shape and the comment articulates the invariant better than my instruction did ("what keeps 'the phase-2 options' meaning the user's request rather than whatever phase 1 left behind"). The take-locals dance is gone; phase-1 clearing is two plain assignments.
- The **skipQuant×dumpEq warning** names the kept RAD's actual path — better than the generic advice I suggested — and the e2e pins both halves: no file written, and the RAD the warning points at is present.
- Consolidation into `requant_options` + both tests: done.

22/0 locally, clippy clean. Merge-ready from my side once #1148 lands and this takes its final trivial rebase (adjacent lines in `requant_options`).

rob-p:
Pushed the final rebase (`7291df55`) so the train keeps moving — all eight conflicts with #1148 were both-additive (the two flag pairs meeting in `AlignQuantOptions`, `requant_options`, the two `run_quant` wiring blocks, and the three forwarding-test sites); both sides kept everywhere. This completes the consolidation the merge train owed: every knob the requant honours now flows through `requant_options` and its field-by-field test. 88/0 on the affected crates, clippy/fmt clean, mergeable.

PR #1148, opened 2026-08-22
feat(align): --noLengthCorrection and --noFragLengthDist in the RAD path (#1140 PR B)

BenjaminDEMAILLE:
PR B of #1140, the feature work. Neither knob existed in the RAD quantifier, so both were accepted and dropped under `--deterministic`: the run corrected effective lengths and weighted by fragment length whatever the user asked for. `--noLengthCorrection` is the one with a constituency, since 3' tagged-end protocols do not sample a transcript's interior uniformly and the corrected length describes something the data never did.

## Mirroring reads mode, not reinventing it

| knob | reads mode | this PR |
|---|---|---|
| `--noLengthCorrection` | effective length is the raw `ref_len`; bias correction disabled (`bias_on && !no_length_correction`) | same, in `quantify_rad`'s eff-length block and its `bias_on` gate |
| `--noFragLengthDist` | `frag_log_prob` returns `LOG_1`, suppressing both the proper-pair PMF term and the orphan/single-end ambiguous term | same, in the eq-class pass and in the bias pass's posterior, the two places reads mode applies it |

Per your warning not to guess on the orphan term: `processor::frag_log_prob` returns early before either branch, so the ambiguous term goes too. I mirrored that rather than only zeroing the proper-pair PMF.

Both flags are wired for `--deterministic`, `-a` and `--rad`, since all three run the same quantifier and none of them had them.

## Measured

3000-transcript fixture, 300k pairs, `quant.sf` rows where `EffectiveLength == Length`:

```
det                                   0 / 2996
det --noLengthCorrection           2996 / 2996      (one-pass --noLengthCorrection: 2996 / 2996)
det --noLengthCorrection --gcBias  2996 / 2996      byte-identical to the line above
det --noFragLengthDist                0 / 2996      lengths untouched, estimates changed
```

The `--gcBias` line is the gate working: bias is disabled rather than silently correcting lengths the run was told not to correct.

## Testing

Two tests in `crates/salmon-align/tests/rad_input.rs`, both at the RAD level so they pin the semantics rather than the CLI wiring:

- `no_length_correction_uses_raw_lengths`: the default shortens every transcript; with the flag, `eff_lengths == lengths` exactly.
- `no_frag_length_dist_changes_weights_not_lengths`: two transcripts of equal length (so the divisor cannot explain anything), each fragment ambiguous between them with placements reporting fragment lengths of 190 (the FLD mean) and 700 (its far tail). With the term, the near-mean placement takes the mass (>290 of 300); without it, the two are indistinguishable and it splits evenly (within 1). Effective lengths identical in both runs.

Worth recording that the obvious fixture does not work: with two long transcripts the proper-pair term is `pmf(flen) - cmf(txpLen)` and `cmf` is 0 for anything much longer than the fragment distribution, so the term cancels between placements and the flag has no visible effect. Making the transcripts short instead makes the effective-length divisor dominate and the EM goes winner-take-all. Equal lengths plus differing per-placement fragment lengths is what isolates the term.

## Left out

The online `-a` path (`quantify_alignments`) still supports neither, which C++ salmon does. Kept out to hold this to the RAD path the default flip needs, as you offered; happy to follow up if you want the parity before 2.6.0.

## Overlap

Touches `run_deterministic` (two forwarded fields) and the `-a` / `--rad` option blocks in `main.rs`, so it overlaps #1146 and my #1144 / #1147 by a few lines each. Trivial rebase whichever order they land in.

### DIFF
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -198,6 +198,18 @@ pub struct AlignQuantOptions {
     /// start time (asctime) of that earlier phase; when set, reported as the
     /// run's `start_time` instead of this pass's own
     pub external_start_time: Option<String>,
+    /// `--noLengthCorrection`: use the raw reference length as the effective
+    /// length, instead of the fragment-length-derived one.
+    ///
+    /// For protocols where a fragment does not sample a transcript's interior
+    /// uniformly (3' tagged-end libraries above all), the usual correction
+    /// describes something the data never did, and the raw length is the honest
+    /// divisor. Bias correction is disabled with it, for the same reason: its
+    /// whole job is to adjust an effective length this mode is not computing.
+    pub no_length_correction: bool,
+    /// `--noFragLengthDist`: drop the fragment-length term from a placement's
+    /// weight, leaving score and compatibility to decide it.
+    pub no_frag_length_dist: bool,
     /// significant digits for the EffectiveLength and NumReads columns of
     /// `quant.sf` (`--sigDigits`, salmon default 3)
     pub sig_digits: u32,
@@ -265,6 +277,8 @@ impl AlignQuantOptions {
             init_uniform: false,
             prior_seconds: 0.0,
             external_start_time: None,
+            no_length_correction: false,
+            no_frag_length_dist: false,
             sig_digits: 3,
             num_error_bins: 4,
             discard_orphans: false,
--- a/crates/salmon-align/src/rad.rs
+++ b/crates/salmon-align/src/rad.rs
@@ -451,6 +451,8 @@ struct FragCfg<'a> {
     ignore_incompat: bool,
     incompat_prior: f64,
     paired_lib: bool,
+    /// `--noFragLengthDist`: leave the fragment-length term out of the weight.
+    no_frag_length_dist: bool,
     range_factorization_bins: u32,
     eq_builder: &'a EquivalenceClassBuilder,
 }
@@ -521,7 +523,12 @@ fn process_rad_fragment(
         let proper = p.status == MateStatus::PairedEndPaired
             && p.frag_len != salmon_rad::FRAG_LEN_UNPAIRED
             && p.frag_len > 0;
-        let log_frag_prob = if proper {
+        let log_frag_prob = if cfg.no_frag_length_dist {
+            // --noFragLengthDist: neither the proper-pair PMF term nor the
+            // orphan ambiguous-length term, matching what the reads path
+            // suppresses (`processor::frag_log_prob`).
+            0.0
+        } else if proper {
             // length-conditioned proper-pair probability. The fixed PMF is
             // normalized by `log P(L ≤ txpLen)` (the CMF at the transcript length)
             // so short transcripts — where only the left tail of insert sizes can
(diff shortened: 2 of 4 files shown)


rob-p (approved):
Reviewed on the branch: both new tests pass, the full salmon-align suite is green (60/0), clippy/fmt clean. The fidelity audit comes back clean at every point I could check it against reads mode:

- `no_length_correction`: raw-length divisor mirrors lib.rs:887, the `bias_on` gate mirrors lib.rs:907, and — the one you couldn't have known to check without digging — reads mode's *third* gate at lib.rs:552 (`want_rough_abund`, phase 1's naive-eq/rough-seed builder) is already `!no_length_correction`-gated, so the whole seed pipeline composes correctly with your change without either of us touching it. The `--gcBias` byte-identity line in your measurement is exactly the right demonstration that the gate works.
- `no_frag_length_dist`: suppressing both the proper-pair PMF term *and* the orphan ambiguous term, in both the eq-class pass and the bias-pass posterior, is the correct mirror of `processor::frag_log_prob`'s early return — and FLD derivation/baking correctly continues (it's reporting, not weighting), matching reads mode.

The second test's design note — recording *why* the obvious two-long-transcripts fixture can't show the effect (the CMF term cancels) and why short transcripts go winner-take-all — is the kind of negative result that saves the next person a day. Equal lengths + divergent per-placement fragment lengths is the minimal isolating fixture and the near-mean/tail assertion pair pins it hard. Approved; two notes.

**1. Rebase landing spot**: once #1144's `requant_options` merges, these two forwards belong inside it and in its field-by-field test (and off the "deliberately not forwarded" list where they'd now be wrong). Same for #1147's two dump flags. Whoever rebases last inherits that consolidation.

**2. On the left-out online `-a` parity — recommend *not* doing the follow-up.** Online `-a` is headed for `--online`-gated deprecation in 2.6.0 and removal after; adding these knobs there would be building on a floor we're tearing out. The right treatment is the flip PR's warn-and-ignore sweep: `--noLengthCorrection`/`--noFragLengthDist` with `--online -a` should say "not supported on the deprecated online path; the default (deterministic) path honours it" rather than gaining an implementation. I'll put it on the #1140 checklist so the gap is handled by policy instead of parity work.

rob-p:
Merge train update: #1144, #1145, and #1149 are on `develop`; you're up next, then #1147 last. This PR now conflicts in `main.rs` exactly where expected — your two `q.*` lines target the hand-copied block #1144 deleted. The conflict resolution is consolidation round 1, all in one commit:

- `q.no_length_correction` / `q.no_frag_length_dist` move **into `requant_options`**.
- Both join the field-by-field forwarding test (values different from the `AlignQuantOptions` defaults) and the defaults-stay-default companion.
- Nothing to remove from the "deliberately not forwarded" list — they were never on it.

No other changes needed; everything else in the PR rebases clean.

BenjaminDEMAILLE:
Rebased, and consolidation round 1 is in the same commit as you asked.

- `q.no_length_correction` / `q.no_frag_length_dist` now live in `requant_options` rather than in the caller, next to the three knobs #1144 moved there, with a line saying why they were absent (the RAD path had no such fields at all until this PR, so `--deterministic` accepted the flags and quantified as though they had not been passed).
- Both join the field-by-field forwarding test with non-default values, and the defaults-stay-default companion.
- Nothing removed from the omissions list, since they were never on it.

The conflict was exactly where you predicted: my two `q.*` lines against the hand-copied block your merge deleted. Resolved by dropping the block wholesale rather than reconciling it line by line, since `requant_options` is now the only place those assignments should exist.

End-to-end still holds on the rebased branch: `--deterministic --noLengthCorrection` gives `EffectiveLength == Length` for all 2996 rows, plain `--deterministic` for none. Workspace suite green, clippy/fmt clean.

#1147 is next and I am on it now: the `requant_options` hoist above the phase-1 mutations, the two dump flags into it and its test, and the `--deterministic --skipQuant --dumpEq` warning with an e2e pinning it.

rob-p:
Verified the rebase on the branch: consolidation round 1 is exactly as specified — both knobs in `requant_options`, both forwarding tests extended, everything green locally (23/0, clippy clean). Merge-ready; #1147 rebases once more on top (trivial adjacent-line conflict in `requant_options`) and the batch is closed.

PR #1149, opened 2026-08-22
fix(cli): --skipQuant --deterministic maps and stops (#1140 follow-up)

BenjaminDEMAILLE:
Taking note 1 from your #1144 review. Phase 1 sets `skip_quant` for its own purposes (map, write the RAD, do not run the online EM), so the user's `--skipQuant` was overwritten there and never reached phase 2: the run quantified anyway, wrote a full `quant.sf` from a flag asking for no quantification, and then deleted the RAD it had just been asked to produce.

Semantics as you suggested, and as `-a --deterministic --skipQuant` already behaves: take the request before phase 1 overwrites it, run phase 1, stop. The RAD is the deliverable rather than an intermediate, so the keep-condition no longer applies to it.

```
--deterministic              quant.sf written, RAD removed        (unchanged)
--deterministic --skipQuant  no quant.sf, RAD kept                (was: quant.sf written, RAD removed)
```

The test drives the binary, since the bug was in how the two passes were wired rather than in either of them, and asserts both directions: a full run still quantifies and still cleans up, a mapping-only run does neither.

This touches the same handful of lines in `run_deterministic` as #1144, #1147, #1148 and your #1146. Trivial rebase in any order; happy to fold it into whichever of mine you merge last if you would rather have one PR there than two.

### DIFF
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -549,7 +549,9 @@ struct QuantArgs {
     /// distribution, so the result is byte-identical across runs and thread
     /// counts. Avoids a second mapping pass. The intermediate RAD is written under
     /// the output directory and deleted on success unless --keepRad (or use
-    /// --writeRad PATH to choose its location and keep it).
+    /// --writeRad PATH to choose its location and keep it). With --skipQuant the
+    /// run stops after mapping and the RAD is kept as its output, in the output
+    /// directory even when --radScratchDir is given.
     #[arg(long = "deterministic")]
     deterministic: bool,
     /// Keep the intermediate RAD produced by --deterministic (by default it is
@@ -1194,9 +1196,25 @@ fn run_deterministic(
     if explicit && scratch_dir.is_some() {
         tracing::warn!("--radScratchDir is ignored when --writeRad names an explicit path");
     }
+    // Scratch placement is for intermediates. Under `--skipQuant` there is no
+    // phase 2 and the RAD is the run's output, so it belongs in the output
+    // directory under its stable name rather than pid-suffixed on a scratch
+    // volume nothing was meant to be kept in. An explicit `--writeRad` still
+    // wins over both.
+    let deliverable = map_opts.skip_quant;
+    if deliverable && !explicit && scratch_dir.is_some() {
+        tracing::info!(
+            "--skipQuant makes the RAD this run's output, so it is written to the output \
+             directory rather than to --radScratchDir"
+        );
+    }
     let rad_path = match rad_out {
         Some(p) => p,
-        None => intermediate_rad_path(out_dir, scratch_dir, "intermediate_mappings")?,
+        None => intermediate_rad_path(
+            out_dir,
+            if deliverable { None } else { scratch_dir },
+            "intermediate_mappings",
+        )?,
     };
     // The RAD writer opens its file before the mapping pass, so the output
     // directory (where the default intermediate lives) must exist first.
--- a/crates/salmon-cli/src/main.rs
+++ b/crates/salmon-cli/src/main.rs
@@ -1215,6 +1215,12 @@ fn run_deterministic(
 
     // Phase 1 — map once, write the RAD (bakes the deterministic FLD + resolved
     // library format), skipping the online EM: quantification happens in phase 2.
+    //
+    // `skip_quant` is about to be set for phase 1 whatever the user asked, since
+    // phase 1's job is to map and write the RAD. Take the request first: with
+    // `--skipQuant` there is no phase 2, and the RAD is the deliverable rather
+    // than an intermediate. Alignment mode already worked this way.
+    let stop_after_mapping = map_opts.skip_quant;
     map_opts.write_rad = Some(rad_path.clone());
     map_opts.skip_quant = true;
     // Derive the FLD + library format order-independently during this pass and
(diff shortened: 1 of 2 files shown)


rob-p (approved):
Verified on the branch: the capture happens before phase 1's overwrite (the one place it can), the e2e asserts both directions (full run still quantifies and cleans up; mapping-only run does neither, with a non-empty RAD), clippy/fmt clean. The semantics match `-a --deterministic --skipQuant` exactly, which was the ask. Approved; two notes.

**1. One interaction to resolve at rebase time, against #1146.** Once `--radScratchDir` exists, a `--skipQuant` run's RAD — now a *deliverable*, per this PR — would land in the scratch directory under a pid-suffixed name, which is the wrong place and the wrong name for an output the user asked to keep. The principled rule: scratch placement is for *intermediates*; when `--skipQuant` promotes the RAD to the output, it belongs in the output directory (explicit `--writeRad` still wins over everything). That's one conditional in the placement logic, and whichever of this PR and #1146 rebases onto the other should carry it — flagging it now so it doesn't fall between them. Worth one line in the e2e too: `--skipQuant --radScratchDir X` → RAD in the output dir, not in `X`.

**2. Cosmetic, take or leave**: the deliverable keeps the name `intermediate_mappings.rad`, which now slightly misdescribes it. The log line naming the path covers discoverability, and `--writeRad PATH` exists for anyone who wants a chosen name, so I wouldn't churn the name — but the `--deterministic` flag's help text ("deleted on success unless --keepRad") deserves a clause noting `--skipQuant` keeps it as the output.

Also acknowledging the response commits on #1144/#1145/#1147 — all three do exactly what the reviews asked (and the #1144 omissions list documenting `skip_quant` as "a bug, see the list" while this PR fixes it is the audit trail working as intended). All four are mergeable from my side once the #1146 interaction has an owner.

BenjaminDEMAILLE:
I took the #1146 interaction, since it is cheaper to own it here than to make your branch carry a rule about a flag it does not know about yet.

This PR is now **rebased onto `feat/det-disk-story` and retargeted at it**, so the diff shows only my two commits; GitHub will retarget it to `develop` when #1146 merges.

The rule, as you framed it: scratch placement is for intermediates. Under `--skipQuant` the RAD is the run's output, so it goes to the output directory under its stable name, with an `INFO` line saying so when `--radScratchDir` was given. An explicit `--writeRad` still wins over both. Before the fix, on my fixture:

```
--skipQuant --radScratchDir /tmp/scratchE
  before   /tmp/scratchE/intermediate_mappings-34073.rad
  after    <out>/intermediate_mappings.rad,  /tmp/scratchE empty
```

The e2e now asserts both halves of that (RAD in the output directory, scratch directory left with nothing in it), on top of the two directions it already checked.

Note 2 taken as the help-text clause rather than a rename, as you preferred: `--deterministic` now says that with `--skipQuant` the run stops after mapping and keeps the RAD as its output, in the output directory even when `--radScratchDir` is given. Name unchanged.

One consequence worth naming: if #1146 lands first this is a normal merge, but if you would rather merge this one first, say so and I will re-cut it against `develop` with the conditional held back for your rebase.

rob-p:
The update does what the review asked and the rebase choice is the right one (stacking on #1146's tip resolves the interaction *in* the PR instead of promising it) — the `deliverable` rule, the info-line explaining the placement override, and the test asserting the scratch directory is left **empty** are all exactly right.

One parity gap in it: the rule landed in reads-mode `run_deterministic` only. `run_deterministic_align`'s keep-condition is `explicit || keep_rad || opts.skip_quant` — the RAD is equally the deliverable of `-a --deterministic --skipQuant` — but its `rad_path` still passes `scratch_dir` unconditionally, so that spelling keeps the output pid-suffixed in scratch. Same one-conditional fix, same `deliverable = opts.skip_quant`, and its keep-message then names a stable path.

(Considered whether `--keepRad --radScratchDir` deserves the same treatment and concluded no: `--keepRad` keeps *the intermediate*, and the user explicitly chose scratch placement — both requests are honoured as stated. `--skipQuant` is different because it changes what the file *is*. Worth a line in the comment if you agree, so the asymmetry reads as decided rather than missed.)

With the `-a` mirror pushed this is merge-ready from my side.

BenjaminDEMAILLE:
Parity gap closed and pushed. `run_deterministic_align` now takes the same `deliverable = opts.skip_quant` conditional, the same info line, and its keep-message names a stable path. Its e2e is a second test: `-a --deterministic --skipQuant --radScratchDir X` writes `intermediate_alignments.rad` into the output directory and leaves X empty.

You are right that its keep-condition already said the RAD was the deliverable there; the path selection was the half that had not caught up, which is a good argument for the rule living next to the placement rather than next to the cleanup.

Agreed on `--keepRad --radScratchDir`, and I have put the reasoning in the commit message so the asymmetry reads as decided: `--keepRad` keeps the *intermediate*, and the user chose scratch placement for intermediates, so both requests are honoured exactly as stated. `--skipQuant` is different in kind because it changes what the file *is*, not merely how long it lives. If a user wants a kept intermediate somewhere specific, `--writeRad PATH` is the flag that says so and still overrides everything.

Both tests green, clippy/fmt clean.

rob-p:
Verified the `-a` mirror on the branch: both e2e tests pass (including the new one pinning the deliverable to the output directory with scratch left empty), clippy clean. Merge-ready from my side — thanks for the fast turnarounds across this whole batch.

PR #1151, opened 2026-08-22
fix(quant): derive MAPQ from the placement count instead of a constant 1

BenjaminDEMAILLE:
The MAPQ half of #1141, as its own PR per your review, fully decoupled from realization: no re-alignment, no `ambiguous.bin`, no new index artifact. The count of placements is already in hand when the records are written, so this is arithmetic on a value the writer holds.

## What changes

Every record carried `MAPQ 1`, which says "probably wrong" about placements salmon chose to keep. `samtools view -q 255`, the standard "uniquely placed" filter and the one STAR output trains people to use, therefore discarded the whole file.

salmon has no phred-scaled probability to offer, and this does not invent one. It reports what salmon does know, on the scale STAR already established:

| placements (`NH`) | `MAPQ` |
|---|---|
| 1 | 255 |
| 2 | 3 |
| 3 or 4 | 1 |
| 5 or more | 0 |

On a 300k-fragment fixture (890098 records):

```
before   890098 records at MAPQ 1;      -q 255 selects 0
after    433027 at 255, 199088 at 3, 168502 at 1, 89481 at 0;   -q 255 selects 433027
```

## Docs

The output-formats page gains a short `MAPQ` section stating the scale and, more importantly, that a salmon `MAPQ` counts placements rather than estimating a probability. That belongs in the pseudobam contract paragraph you are landing for 2.6.0; treat this as a first draft of that one field's entry and fold it in if it fits better there.

## Scope

Unit test for the scale, including the boundary at four placements and the `nh == 0` case a record built without a placement would hit. Full workspace suite green, clippy and fmt clean. Nothing else from #1141 is here.

I have taken the rest of your review as written: #1141 and #1142 wait for the opt-in design discussion after 2.6.0, and the scoring-path question goes to its own issue with the fixtures, which I am putting together now. I do want to answer one substantive point there rather than here, since it changes what the issue should measure: the anchor pathologies were found by the realization DP, but the reason I think they are worth chasing is that the *positions* they disagree with come from the same chaining that feeds scoring. Whether that disagreement reaches the scores is exactly the open question, and I would rather measure it than assert it.

### DIFF
--- a/crates/salmon-quant/src/mapping_record.rs
+++ b/crates/salmon-quant/src/mapping_record.rs
@@ -165,6 +165,31 @@ pub fn complement(base: u8) -> u8 {
     }
 }
 
+/// The `MAPQ` to report for a fragment with `nh` placements.
+///
+/// SAM's `MAPQ` is a phred-scaled probability that the placement is wrong, and
+/// salmon does not compute one: its whole design is to keep every plausible
+/// placement and let the EM apportion them. What it does know is *how many*
+/// placements a fragment has, which is the same thing STAR reports, so this uses
+/// STAR's mapping: a uniquely placed fragment gets 255 and the value falls off
+/// as the fragment becomes more ambiguous. Matching STAR matters because the
+/// RNA-seq tools downstream of a salmon BAM are the ones already tuned to it: to
+/// all of them `-q 255` means "uniquely placed".
+///
+/// The previous hardcoded `1` made every record look ambiguous, so any such
+/// filter discarded the whole file. It is the one placeholder in this output
+/// that costs nothing to replace: the count is already in hand when the records
+/// are written, so no alignment work is implied (COMBINE-lab/salmon#1140,
+/// #1141).
+pub fn mapping_quality(nh: usize) -> u8 {
+    match nh {
+        0 | 1 => 255,
+        2 => 3,
+        3..=4 => 1,
+        _ => 0,
+    }
+}
+
 /// The CIGAR and clamped position for a read placed at `pos` on a transcript of
 /// length `txp_len`.
 ///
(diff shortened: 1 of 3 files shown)


rob-p (approved):
Verified on the branch: unit test + the SAM/BAM equivalence e2e pass, clippy clean, and the diff is exactly the invited scope — arithmetic on a count the writer already holds, `mapping_record.rs` only, quantification untouched, no realization, no index artifact. The STAR scale is reproduced correctly (255 / 3 / 1 for 3–4 / 0 for 5+), and applying the fragment-level `nh` uniformly to both mates' records and to orphans is the right consistency. Approved; two notes, one with a small ask.

**1. `mapping_quality(0) == 255` — consider 0 instead.** Unreachable today (`maps.is_empty()` skips the loop), so this is future-proofing, but the future it proofs against is exactly #1142's unmapped records: an unmapped record carrying MAPQ 255 would be actively wrong, and whoever adds unmapped output is unlikely to remember this function's edge case. `0` is the value samtools convention expects on unmapped records, and no mapped record can reach the arm. One-character change plus the test line.

**2. Docs coordination, mine to do**: #1150's pseudobam-contract paragraph currently says "`MAPQ` is nominal; do not filter on it" — true on `develop` today, false the moment this merges. Whichever lands second gets the reconciliation; if this merges first (it should — it's small and standalone), I'll fold your MAPQ entry into the contract paragraph on #1150's rebase, replacing the nominal-MAPQ line with the derived-scale description, as your PR body anticipated.

Also noted and agreed on the scoring-path issue framing: measuring whether the chaining disagreements reach the scores, rather than asserting either way, is exactly the right shape — looking forward to the fixtures.

Issue #1162, opened 2026-08-23
Group 5: error/message quality items from the pre-2.6.0 readiness sweep


rob-p:
Tracking issue for the **Group 5** items from the pre-2.6.0 release-readiness sweep (follow-up to #1140). Groups 1–4 are being addressed now, one PR each; this is the deferred remainder, to attack once those land.

None of these produces a wrong answer — they are about what happens when the user gets something wrong, and how clearly salmon says so.

### Deferred items

| Case | Current | Proposed |
|---|---|---|
| `-a --gcBias` (or `--seqBias`/`--posBias`) without `-t` | Runs the **entire** BAM→RAD pass — potentially hours on real data — then fails with the `--rad` wording ("with `--rad` require -t/--targets") for an invocation containing no `--rad`. The correctly-worded twin is reachable only from deprecated `--online` | Check at invocation, before any work; fix the message to name the flag the user actually passed |
| FASTQ parse errors | Name neither the file nor the record: `mapping failed: FASTX error: Invalid FASTQ separator`. `-1/-2/-r` all accept many files, so the user cannot tell which. This is the only input-error class in the tree that does not chain a path | Chain the path and the record identifier |
| `--rad` with an unwritable `-o` | Reads the RAD and runs the whole EM before failing, and the message carries no path. Reads and `-a` both check first and name the directory | Check the output directory up front, like the other drivers |
| `-2` without `-1` | "no reads provided", though `-2` was provided | Say what is actually missing |
| `--decoyThreshold` sketch warning | Fires in `-a`/`--rad`, asserting sketch is running when `--sketch` was itself silently dropped. The check sits before mode dispatch, unlike the main sketch block | Move inside the reads dispatch |
| Single-flag plural verb | Three joined-list warnings interpolate a plural verb for one flag ("`--writeRad` shape an intermediate RAD"); a fourth site already handles this with `if inert.len() == 1` | Reuse the existing singular/plural branch |
| `--skipQuant --dumpEq` | Writes eq classes in `-a`/`--rad` but not default reads (which emits a good explanatory warning) — a help-text mismatch rather than a behaviour bug | Reconcile the help text with the behaviour |

### Also noted, upstream

[...]

PR #1166, opened 2026-08-23
fix(groups 4+5): metadata that tells the truth, and errors that arrive in time

rob-p:
**Groups 4 and 5** (#1140, #1162). You asked me to parallelize these — five agents staged disjoint files simultaneously, and I assembled and verified the result. Because the work interleaves in the same files (`main.rs` carries both groups), it lands as one PR rather than two, sectioned by group below. Stacked on #1165.

## Group 4 — metadata asserting something the run did not do

| Field | Was | Now |
|---|---|---|
| `library_types` | Reported what the **detector observed**: `-l ISR` on ISF data said `["ISF"]`, contradicting `cmd_info.json` and `lib_format_counts.json` in the same directory, on a run that really did filter as ISR and mapped **0** fragments | Reports the **applied** format. Per your call the metadata reports **both** — applied in `library_types`, inferred in `detected_library_type` |
| `num_decoy_targets` | Hardcoded `0` while `num_decoy_fragments` beside it was real, with `meta_info_complete: true` | The real count, derived from the same row split as `quant.sf` so it cannot contradict `num_valid_targets` |
| `frag_length_source` | `rad_baked` for a plain `-a` run — naming an internal intermediate the user never asked for | `alignments` for alignment input; genuine `--rad` keeps `rad_baked`, where it is correct |
| `logs/salmon_quant.log` | Labelled a default reads run "alignment mode", dropped the documented mapping-type/rate lines, and called unmapped fragments "strand-incompatible" on a run with `num_incompatible_fragments = 0` | Describes the mode the user chose |
| `num_bias_bins` | `0` | **Still `0`, now with the reason recorded** |

The `frag_length_source` fix needed a real signal rather than path-sniffing, so `AlignQuantOptions` gained `input_is_alignments`, set by the two alignment drivers.

On `num_bias_bins` — two agents independently traced it to C++'s `readBias(FORWARD).counts.size()`, the legacy fixed 4096-bin k-mer histogram behind `observed_bias.gz` that this port replaced. The GC and positional bin counts *are* in scope at the write site, but filling either in would hand a C++-schema reader a different model's number in the wrong unit. Left alone deliberately.

## Group 5 — errors that arrived too late or named the wrong flag

- **`-a` + bias without `-t` read the entire BAM before failing** — hours on real data — with a message hard-coded to say `--rad` for an invocation containing no `--rad`. Now fails at invocation (**0.00 s**), naming `-a`, the bias flags actually passed, and `-t`.
  Genome projection is deliberately **excluded**: it reconstructs reference sequences from `--genome` plus the annotation, so a blanket gate would have broken a working invocation. That catch came from the agent pushing back on my brief.
- **`quantify_rad`'s backstop message no longer asserts which input flag you typed.** It serves four user-facing modes; it now names the requirement, not the mode. Its module docs say so too, since that premise is why the bug existed.
- **FASTQ parse errors name the file.** With several files the message states the list is the *candidate set* rather than claiming a culprit — the pinned paraseq discards reader identity before the error surfaces, so naming the exact file needs an upstream change. Recorded in #1162.
- `-2` without `-1` says what is missing instead of "no reads provided".
- The `--decoyThreshold`/`--sketch` warning moved inside reads dispatch; it used to fire in `-a`, asserting sketch was running on a run that never mapped.
- `--noSingleFragProb` and `--skipQuant` help text, `docs/cli-flag-status.md` and the migration page corrected where they described pre-fix behaviour.

## One correction to Group 1

A FASTQ whose final record is **malformed rather than short** was told to "re-transfer the file". salmon cannot distinguish a truncated transfer from a corrupted one at that point, so it now describes what it observed and says as much. My own Group 1 test caught this when the wording changed.

## Tests

Seven new end-to-end tests covering each metadata field and each error path, including the controls that matter: `-l A` still agrees with the detector, genuine `--rad` still reports `rad_baked`, and the sketch warning still fires where sketch really runs.

Suite **338/0**, clippy clean.

https://claude.ai/code/session_016cfXr9P5JDqoEtafmTGpNs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2744,6 +2744,7 @@ dependencies = [
  "salmon-index",
  "salmon-map",
  "salmon-quant",
+ "serde_json",
  "tempfile",
  "tracing",
  "tracing-subscriber",
--- a/crates/salmon-align/src/lib.rs
+++ b/crates/salmon-align/src/lib.rs
@@ -203,6 +203,15 @@ pub struct AlignQuantOptions {
     /// invocation (index, read files, threads); phase 2's own record would
     /// replace it with a RAD-centric one and lose all of that (#1140)
     pub preserve_cmd_info: bool,
+    /// this run converted alignments into the RAD it is quantifying (`-a`
+    /// `--deterministic`, genome projection) rather than being handed one.
+    /// Only metadata depends on it: `frag_length_source` must say `alignments`
+    /// — the distribution came from the input alignments — instead of leaking
+    /// `rad_baked`, which names an intermediate the user never asked for. It
+    /// cannot be inferred: `preserve_cmd_info` marks the *reads* driver, and a
+    /// standalone `--rad` (where `rad_baked` is correct) looks identical here
+    /// otherwise. Left false by anything reading a user-supplied RAD.
+    pub input_is_alignments: bool,
     /// the read files behind this RAD, when the driver knows them (reads-mode
     /// `--deterministic`), for `lib_format_counts.json`'s `read_files` — which
     /// otherwise reports `[]` because a RAD does not name its reads (#1140)
@@ -317,6 +326,7 @@ impl AlignQuantOptions {
             prior_seconds: 0.0,
             external_start_time: None,
             preserve_cmd_info: false,
+            input_is_alignments: false,
             read_files: Vec::new(),
             dump_eq: false,
             dump_eq_weights: false,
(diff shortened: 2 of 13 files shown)


rob-p:
Added `a0b43d4b`: **the bias models' shapes are now recorded**, answering your question — nothing captured them anywhere before.

`num_bias_bins` still cannot carry them (it names C++'s legacy 4096-bin k-mer histogram, which this port replaced) and stays `0` for schema compatibility. Three new salmon-rs fields report the models that actually run, in **both** writers so every driver agrees:

| Field | Meaning |
|---|---|
| `seq_bias_context_length` | context length in nucleotides |
| `gc_bias_bins` | total GC bins (`--numGCBins` × `--conditionalGCBins`) |
| `pos_bias_bins` | `{ length_classes, bins_per_class }` |

Two deliberate choices worth a look in review:

- **Measured from the produced models** (the dump arrays' own lengths), not from the requested options — so a configured-but-unbuilt model cannot be described as though it ran.
- **Absent, not zero,** when the correction was not requested. A `0` would read as a real model with no bins; absence is how a consumer tells "off" from "on with N bins".

Verified across drivers — reads, `-a` and `--online` all report `9` / `75` / `{5, 20}` with all three corrections on — and `gc_bias_bins` tracks the knobs rather than a constant (`--numGCBins 30 --conditionalGCBins 4` → `120`). Documented in the output-format spec, including why `num_bias_bins` stays 0, and in the release notes.

Suite **339/0**, clippy clean.

PR #1170, opened 2026-08-23
ci: let Dependabot open version updates, grouped

BenjaminDEMAILLE:
There is no `.github/dependabot.yml` in the repository. Security updates run without one (that is where #1121's js-yaml bump came from), but *version* updates do not, and the workflows show what that has cost:

| action | pinned at |
|---|---|
| `actions/checkout` | `v4` in `ci.yml`, `docker.yml`, `docs.yml`; `v6` in `release.yml` |
| `actions/upload-artifact` | `v4` in `docker.yml`; `v7` in `release.yml` |
| `actions/download-artifact` | `v4` in `docker.yml`; `v8` in `release.yml` |

Nobody chose that split, and nothing surfaces it today.

## What the config covers

Four ecosystems, which is what the repository actually has: the cargo workspace at `/`, the npm project under `/website`, the workflows, and the `Dockerfile`'s base image.

## Grouping, because ungrouped is worse than nothing

A Rust workspace this size produces roughly a dozen individual bumps a week. A wall of one-line PRs gets reviewed by nobody and trains everyone to ignore the label, so every ecosystem here batches:

- **`cargo-minor`** — one weekly PR for routine patch and minor movement.
- **`combine-lab-upstream`** — `piscem-rs`, `paraseq`/`paraseq-temp`, `libradicl`, `cf1-rs`, `sshash-lib`, `ksw2rs` in their own PR. These move with work happening in those repositories, and their bumps have carried deliberate decisions (the piscem-rs 0.9.2 "parked pool as a floor" commit, the paraseq-temp 0.5.0-pre.2 republication). A bump like that should never arrive buried in a batch of unrelated ones.
- **cargo majors** stay ungrouped: each is a decision, and each deserves its own diff and its own CI run.
- **`website-minor`** and **`actions`** batch wholesale. An action major is a workflow change, visible in the diff, and CI is its own test for it.
- **docker** is monthly and ungrouped, since `rust:1.91-bookworm` is a toolchain decision rather than a dependency refresh.

## Two details worth checking against your intent

- **`target-branch: develop`** on every entry. The default would be `master`, which carries releases, and everything else lands through `develop` first. Say if you would rather they target `master`.
- **Commit prefixes** are set explicitly (`chore(deps)`, `ci(deps)`, `build(deps)`) rather than left to Dependabot's inference from history.

Both are one-line changes if you want them differently, as is the whole grouping policy: if the lab-upstream group turns out to be noise rather than signal, an `ignore` entry for those crates replaces it.

Config validated as YAML; nothing else in the repository changes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -53,11 +53,17 @@ updates:
       # and paraseq-temp 0.5.0-pre.2 commits). Kept in their own PR so such a
       # bump is never buried in a batch of unrelated ones.
       combine-lab-upstream:
-        # Minor and patch only, so a major from one of these lands on its own
-        # like every other major. Grouping a piscem-rs 1.0 with unrelated
-        # bumps is exactly what the separate group exists to prevent.
+        # Patch only, because all six are 0.x and Cargo's compatibility
+        # boundary there is the MINOR component: ^0.9.2 is >=0.9.2 <0.10.0, so
+        # 0.9.2 -> 0.9.3 is compatible and 0.9.2 -> 0.10.0 is breaking.
+        # Dependabot classifies by moved field, not by that boundary, which
+        # makes the breaking one `minor` to the grouping logic. Verified against
+        # dependabot-core rather than inferred: for a dependency at 0.9.2, the
+        # patch level covers `> 0.9.3, < 0.10` and the minor level starts at
+        # `>= 0.10.a`, with no 0.x special case anywhere. Allowing `minor` here
+        # would therefore batch a breaking upgrade of our own upstream crates
+        # into a routine PR, which is the case this group exists to isolate.
         update-types:
-          - minor
           - patch
         patterns:
           - piscem-rs
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -61,18 +61,13 @@ updates:
           - patch
         patterns:
           - piscem-rs
-          # Both `paraseq` spellings are listed, and BOTH ARE INERT TODAY.
-          # `paraseq = { package = "paraseq-temp", ... }` is a rename, and
-          # dependabot-core's cargo parser skips a renamed manifest entry
-          # outright (`next unless name == name_from_declaration(...)` in
-          # cargo/lib/dependabot/cargo/file_parser.rb). Running that parser over
-          # this workspace confirms it: `paraseq` is not parsed at all, and
-          # `paraseq-temp` is seen only through Cargo.lock, with no requirement
-          # attached, which makes it indirect and therefore outside the default
-          # `allow`. So paraseq bumps stay manual, as they have been. The
-          # patterns are kept for the day the temporary republication goes away
-          # and the entry becomes an ordinary direct dependency again.
-          - paraseq
+          # `paraseq-temp`, not `paraseq`: the manifest used to rename it, and
+          # dependabot-core's cargo parser skips a renamed entry outright
+          # (`next unless name == name_from_declaration(...)` in
+          # cargo/lib/dependabot/cargo/file_parser.rb), which left the crate
+          # visible only through Cargo.lock as an indirect dependency no update
+          # would be raised for. The rename is gone, so this pattern matches a
+          # direct dependency; do not reintroduce it.
           - paraseq-temp
           - libradicl
           - cf1-rs
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -61,9 +61,17 @@ updates:
           - patch
         patterns:
           - piscem-rs
-          # `paraseq` in the manifest is the `paraseq-temp` republication, so
-          # both spellings are listed rather than guessing which one Dependabot
-          # reports for a renamed package.
+          # Both `paraseq` spellings are listed, and BOTH ARE INERT TODAY.
+          # `paraseq = { package = "paraseq-temp", ... }` is a rename, and
+          # dependabot-core's cargo parser skips a renamed manifest entry
+          # outright (`next unless name == name_from_declaration(...)` in
+          # cargo/lib/dependabot/cargo/file_parser.rb). Running that parser over
+          # this workspace confirms it: `paraseq` is not parsed at all, and
+          # `paraseq-temp` is seen only through Cargo.lock, with no requirement
+          # attached, which makes it indirect and therefore outside the default
+          # `allow`. So paraseq bumps stay manual, as they have been. The
+          # patterns are kept for the day the temporary republication goes away
+          # and the entry becomes an ordinary direct dependency again.
           - paraseq
           - paraseq-temp
           - libradicl
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,13 +10,21 @@
 # size produces a dozen individual bumps a week otherwise, and a wall of PRs is
 # reviewed by nobody, which is worse than no updates at all.
 #
-# NONE OF THIS APPLIES TO SECURITY UPDATES. GitHub only uses a `dependabot.yml`
-# entry for security updates when it sets no `target-branch`, and every entry
-# here sets one. So a security bump keeps arriving the way it does today:
-# against the default branch (`master`), ungrouped, without these commit
-# prefixes. That is deliberate (a security fix belongs on the release branch
-# directly, not queued behind `develop`), but it is worth knowing before
-# wondering why one showed up looking nothing like the PRs described below.
+# Every entry targets `develop`. Nothing here should ever open a PR against
+# `master`, which carries releases.
+#
+# NONE OF THIS APPLIES TO SECURITY UPDATES, and that is a constraint rather
+# than a preference. GitHub only uses a `dependabot.yml` entry for security
+# updates when it sets no `target-branch`, and every entry here sets one. So a
+# security bump keeps arriving against the *default* branch (`master`),
+# ungrouped and without these commit prefixes. The two settings cannot both be
+# satisfied from this file: dropping `target-branch` would bring security
+# updates under this config but send every version update to `master` too,
+# which is worse.
+#
+# The only way to get security PRs onto `develop` is to make `develop` the
+# repository's default branch. That is a repository setting, not a config
+# change, and it is deliberately not assumed here.
 version: 2
 updates:
   # ---- Rust workspace ------------------------------------------------------
@@ -32,9 +40,23 @@ updates:
     commit-message:
       prefix: chore(deps)
     groups:
-      # One PR a week for the routine movement. Patch and minor bumps of
-      # well-behaved crates are reviewed as a batch or not at all.
-      cargo-minor:
+      # One PR a week for the routine movement: patch level only, which is the
+      # tier that is mergeable on a glance.
+      #
+      # Minor is deliberately NOT batched here, for the same reason it is not
+      # batched in `combine-lab-upstream` below: 34 of this workspace's 55
+      # direct dependencies are 0.x, and for a 0.x requirement Cargo's
+      # compatibility boundary is the MINOR component, so `noodles-*`,
+      # `rapidgzip-core`, `sux`, `hashbrown`, `jiff`, `rand*`, `crossbeam*` and
+      # friends all break at a bump Dependabot classifies as `minor`. Batching
+      # those would hide a breaking upgrade inside a routine PR.
+      #
+      # Unmatched updates are not suppressed — they simply arrive ungrouped,
+      # one PR each, with their own diff and their own CI run. That is the
+      # point: a minor is a change to explain, and a major is a decision to
+      # make, neither of which is served by a batch. `open-pull-requests-limit`
+      # above bounds how many can be open at once, so the queue cannot run away.
+      cargo-patch:
         patterns:
           - "*"
         exclude-patterns:
@@ -46,7 +68,6 @@ updates:
           - sshash-lib
           - ksw2rs
         update-types:
-          - minor
           - patch
       # The lab's own upstream crates move with work happening in those repos,
       # and their bumps have carried deliberate floors (see the piscem-rs 0.9.2
@@ -100,6 +121,10 @@ updates:
     commit-message:
       prefix: chore(deps)
     groups:
+      # Minor IS batched here, unlike the cargo groups, and that is a choice
+      # rather than an oversight: this is the documentation site, so nothing it
+      # depends on can affect quantification results. Tightening it to patch
+      # would buy noise instead of safety.
       website-minor:
         patterns:
           - "*"
@@ -109,9 +134,15 @@ updates:
 
   # ---- GitHub Actions ------------------------------------------------------
   #
-  # The reason this file exists. One weekly PR that keeps every workflow on the
-  # same action versions, including the majors: an action major is a workflow
-  # change, visible in the diff, and CI is its own test.
+  # The reason this file exists: the workflows had drifted to three different
+  # `actions/checkout` majors with nothing to surface it. One weekly PR keeps
+  # every workflow on the same minor/patch versions.
+  #
+  # Majors are excluded from the group on purpose. An action major is a
+  # breaking workflow change — v4 to v5 of `checkout` or `upload-artifact` has
+  # changed defaults and outputs before — and "the diff is visible and CI is
+  # green" is not the standard we want for adopting one. Each arrives as its
+  # own PR, to be taken deliberately or closed.
   - package-ecosystem: github-actions
     directory: /
     schedule:
@@ -125,6 +156,9 @@ updates:
       actions:
         patterns:
           - "*"
+        update-types:
+          - minor
+          - patch
 
   # ---- container base image ------------------------------------------------
   #
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -61,13 +61,20 @@ updates:
           - patch
         patterns:
           - piscem-rs
-          # `paraseq-temp`, not `paraseq`: the manifest used to rename it, and
-          # dependabot-core's cargo parser skips a renamed entry outright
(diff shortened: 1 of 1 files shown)


rob-p:
Reviewed. I verified every factual claim in the description against the repo rather than taking them on trust, and they all hold:

- The action drift is real and exactly as described — `ci.yml`/`docker.yml`/`docs.yml` on `actions/checkout@v4`, `release.yml` on `@v6`, with `upload-artifact` at v4/v7 and `download-artifact` at v4/v8.
- All four ecosystems exist: cargo workspace at `/`, `website/package.json`, `.github/workflows/`, root `Dockerfile`. There is no existing `.github/dependabot.yml`.
- The default branch really is `master`, so `target-branch: develop` is a deliberate override rather than a redundant line.
- All seven `combine-lab-upstream` crates exist. Worth noting `sshash-lib` and `ksw2rs` are declared in `crates/salmon-map/Cargo.toml`, not the root manifest — they are still in scope because Dependabot resolves the whole workspace from `/`, but the patterns would silently cover nothing if that ever changed.
- Schema-validated: `version: 2`, four valid ecosystems, no unknown keys, `update-types` values legal.

Listing both `paraseq` and `paraseq-temp` is the right call given `paraseq = { package = "paraseq-temp", ... }` — matching on renamed packages is exactly where a single spelling would silently miss.

Two things before this merges, one of which I think you want to know about explicitly.

### 1. `target-branch` means this config will not apply to security updates

GitHub's docs are explicit:

> "In order for Dependabot to use this configuration for security updates, the `directory` must be the path to the manifest files … and **you should not specify a `target-branch`**."

Since every entry here sets `target-branch: develop`, none of this configuration reaches security updates. In practice that means security PRs keep behaving exactly as they do today — raised against `master`, ungrouped, without the `chore(deps)`/`ci(deps)` prefixes. That is not a regression (it is the status quo, and it is where #1121's js-yaml bump came from), and arguably it is what you want: a security fix should go to the release branch directly rather than queue behind `develop`.

But it does mean the grouping and prefix policy described here silently does not cover the security PRs, which is the category most likely to arrive unannounced. Worth one line in the file saying so, so the next person doesn't wonder why a security bump showed up on `master` unprefixed and ungrouped. **This matters right now** — there is currently an open high-severity Dependabot alert on the default branch, so that path is live.

### 2. The description and the config disagree about majors

The body says *"cargo majors stay ungrouped: each is a decision, and each deserves its own diff and its own CI run."* That holds for everything caught by `cargo-minor` (which is correctly restricted to `minor`/`patch`), but **`combine-lab-upstream` has no `update-types`**, so it matches all update types — including majors. A `piscem-rs` 1.0 would therefore arrive grouped with any other lab-crate bumps rather than alone.

Given the reasoning in the PR — that these bumps "have carried deliberate decisions" — I suspect grouping their majors is the *opposite* of the intent. Two ways to resolve it:

```yaml
      combine-lab-upstream:
        patterns: [...]
        update-types:
          - minor
          - patch
```

which makes lab-crate majors ungrouped like every other major, and matches the stated policy. Or leave the behaviour and fix the description. Either is a one-line change; I just don't want the file and its rationale to say different things, since the rationale is what the next person will read.

Neither point blocks. The config is well-constructed and the grouping argument is right — a wall of one-line PRs really does get reviewed by nobody.

BenjaminDEMAILLE:
Both taken, pushed.

**1. Security updates.** You are right, and I had not checked it: with a `target-branch` set on every entry, none of this reaches security updates. The behaviour is what I would choose anyway (a security fix belongs on the release branch, not queued behind `develop`), but the file gave no way to know that, which is the actual problem given there is a live high-severity alert on the default branch right now. The header now says it in as many words, including what a security PR will look like instead: against `master`, ungrouped, unprefixed.

**2. Majors in the lab group.** Caught a real contradiction: no `update-types` on `combine-lab-upstream` meant it matched majors, so a `piscem-rs` 1.0 would have arrived grouped with whatever else moved that week. That is the opposite of why the group exists. Fixed the config rather than the description, since the description was the intent: `minor` and `patch` only, and lab majors now land alone like every other major.

Thanks for checking the claims against the repo rather than taking them. Your note on `sshash-lib`/`ksw2rs` living in `crates/salmon-map/Cargo.toml` is worth keeping in mind: the patterns work today because Dependabot resolves the workspace from `/`, and they would silently match nothing if either crate ever moved out of the workspace. Not something the config can guard against, but it is the kind of silent-no-op this batch has spent a week eradicating, so I would rather it be written down somewhere than only in this thread.

YAML revalidated; nothing else changed.

rob-p (approved):
Both points addressed exactly, and I re-validated rather than eyeballing the diff.

**Grouping now matches the stated policy.** Re-checked the full matrix:

| ecosystem | group | update-types |
|---|---|---|
| cargo | `cargo-minor` | minor, patch (7 lab crates excluded) |
| cargo | `combine-lab-upstream` | minor, patch |
| npm | `website-minor` | minor, patch |
| github-actions | `actions` | all — deliberate, an action major is a workflow change visible in the diff |

The two cargo sets are exact complements — the seven crates excluded from `cargo-minor` are precisely the seven the lab group matches, so nothing falls between them and nothing is double-covered. Lab minor/patch batch together, lab majors land alone like every other major. Schema still valid.

**The security-update note is the right call**, and I'd single out that you put it at the top of the file rather than beside one entry. It is a property of every entry, and the reader who needs it is the one wondering why a bump arrived on `master` looking nothing like what the file describes — that person is skimming the header, not reading the cargo section.

One thing worth stating for the record since it is now load-bearing: this is a real constraint, not a preference. GitHub only applies a `dependabot.yml` entry to security updates when that entry sets no `target-branch`, so keeping `develop` as the version-update target and letting security updates go to `master` is the only combination that gets both. The alternative — dropping `target-branch` to bring security updates under this config — would send every routine dependency bump straight to the release branch, which is clearly worse.

Approving. The action drift this fixes is the kind of thing that stays invisible until a release workflow breaks on an action that has been deprecated for a year.

BenjaminDEMAILLE:
Correction to my own PR, and it takes back something you and I both let stand.

You checked my claims against the repository. I had checked them against the repository and GitHub's documentation, but not against `dependabot-core`, and one of them does not survive contact with the implementation.

**The `paraseq` patterns match nothing.** `paraseq = { package = "paraseq-temp", ... }` is a rename, and the cargo parser skips a renamed manifest entry outright:

```ruby
# cargo/lib/dependabot/cargo/file_parser.rb, manifest_dependencies
next unless name == name_from_declaration(name, requirement)
```

`name` is `paraseq`, `name_from_declaration` returns `paraseq-temp`, so the entry is dropped before a dependency is ever built from it. I ran that parser over this workspace in the official updater image rather than trusting the read:

```
parsed 374 dependencies
  piscem-rs      0.9.2       direct (1 req)
  paraseq        -           NOT PARSED AT ALL
  paraseq-temp   0.5.0-pre.2 TRANSITIVE (no manifest requirement)
  libradicl      0.17.0      direct (1 req)
  cf1-rs         0.5.1       direct (1 req)
  sshash-lib     0.6.4       direct (1 req)
  ksw2rs         0.2.0       direct (1 req)
```

`paraseq-temp` reaches the parser only through `Cargo.lock`, with no requirement attached, which makes it indirect and therefore outside the default `allow`. So no version-update PR will ever be opened for it, whatever the group says. Your note that listing both spellings was "the right call" was right in spirit and wrong in outcome, and it was my claim to check.

I have kept the patterns and written the finding next to them, rather than deleting them: the manifest itself calls the republication temporary, and the day it goes away the entry becomes an ordinary direct dependency and the patterns start working. What I did not want was a config line that reads like coverage while matching nothing, which is precisely the class of defect this batch has spent a week removing from the CLI. It seemed poor form to introduce one in the file that is supposed to keep the repository honest.

Two consequences worth having on the record:

- **`paraseq-temp` bumps stay manual** until the rename is dropped. That is the status quo, not a regression, but it is now stated rather than implied.
- Everything else in the lab group is a genuine direct dependency and will be grouped as described. Verified above rather than assumed.

The rest of the config is unchanged, and the two points from your review (the security-update caveat, majors ungrouped) stand as pushed.

BenjaminDEMAILLE:
Better than documenting it: **#1173 removes the rename**, so the pattern matches a direct dependency instead of nothing.

The rename turned out to be unnecessary in the first place. `paraseq-temp` declares `[lib] name = "paraseq"` in its own manifest, so `use paraseq::...` resolves the same with or without the key: #1173 changes no code and no `Cargo.lock` line, only the two manifest entries. The comment's stated reason for the rename (keeping `use paraseq::...` unchanged) does not hold, and the hazard it warns about is unrelated to the key, since piscem-rs resolves `paraseq-temp` as well.

Same parser, same workspace, before and after:

```
before   paraseq-temp   0.5.0-pre.2   TRANSITIVE (no manifest requirement)
after    paraseq-temp   0.5.0-pre.2   direct (1 req)
```

This PR now lists only the `paraseq-temp` spelling, with a note saying why it is spelled that way and asking that the rename not come back. If you would rather not touch the manifest, say so and I will close #1173 and restore the "inert, kept for later" note here instead. But given that this is the one pinned pre-release in the tree, waiting on an upstream PR to land, it seemed like the wrong dependency to leave invisible to the tool whose job is to tell us it moved.

BenjaminDEMAILLE:
Note 2 taken: `paraseq` is back in the pattern list, alongside `paraseq-temp`, and the comment now gives each spelling its own reason. `paraseq-temp` is what the manifest declares and what the parser sees today; `paraseq` is the entry that starts working the day the pool lands in noamteyssier/paraseq and the manifest moves back to the official crate.

That is a better argument than my reason for dropping it. I removed it because it matched nothing, which is the right instinct for a config line generally, but this one has a specific future in which it matches, and the alternative is depending on someone remembering to edit this file at exactly the moment they are busy switching crates.

Merge order noted: #1173 first, or the two together. Nothing else on this branch is waiting on anything.

rob-p (approved):
Re-approving after the paraseq comment.

Re-validated: schema parses, the `cargo-minor` exclusions and the `combine-lab-upstream` patterns are still exact complements over the same seven crates, and the lab group is still `minor`/`patch` only.

The comment is better than what I suggested. I proposed keeping the `paraseq` spelling as insurance; you also wrote down *why the current spelling must not be reverted*, with the parser line that makes it true. Those are the two failure modes, and they pull in opposite directions — someone tidying the manifest would remove `paraseq-temp` for looking ugly, and someone tidying this file would remove `paraseq` for matching nothing. Each now has its reason attached at the place the edit would happen, which is the only place a comment gets read.

Worth noting the pair only works because of the ordering: with #1173 in, `paraseq-temp` matches a direct dependency, and `paraseq` is dormant until the pool lands upstream. Landing this without #1173 would leave both patterns matching nothing, which is the state the earlier revision of this file had to describe. Merging #1173 first, or the two together, avoids ever committing that.

rob-p:
Change requested on the `combine-lab-upstream` group — @rob-p's call, and having checked the versions I think it is sharper than it first looks.

**Every one of those six crates is pre-1.0:**

| crate | requirement |
|---|---|
| `piscem-rs` | 0.9.2 |
| `libradicl` | 0.17 |
| `sshash-lib` | 0.6 |
| `cf1-rs` | 0.5 |
| `paraseq-temp` | 0.5.0-pre.2 |
| `ksw2rs` | 0.2.0 |

For a `0.x` requirement Cargo treats the **minor** component as the compatibility boundary: `^0.9.2` is `>=0.9.2, <0.10.0`, so `0.9.2 → 0.9.3` is compatible and `0.9.2 → 0.10.0` is a **breaking change**. Dependabot, meanwhile, classifies update types by version component — `0.9 → 0.10` moves the minor field, so it is `semver-minor` to the grouping logic.

The consequence is that `update-types: [minor, patch]` on this group does the opposite of what its comment promises: a breaking `piscem-rs 0.9 → 0.10` would arrive **batched with unrelated lab bumps**, which is precisely the case the separate group was created to isolate. The `cargo-minor` group has the same exposure for every other 0.x crate in the tree, but the lab crates are the ones whose breaking changes have actually carried deliberate decisions this cycle — the 0.9.2 parked-pool floor, the paraseq-temp republication.

Rob's framing is that salmon should track these deliberately, "adopting the latest *compatible* version where feasible" — which for a 0.x requirement means patch-level only. Concretely:

```yaml
      combine-lab-upstream:
        # These are all 0.x, where Cargo's compatibility boundary is the MINOR
        # component: 0.9.2 -> 0.9.3 is compatible, 0.9.2 -> 0.10.0 is breaking.
        # Dependabot classifies the latter as `minor` (the field moved), so
        # allowing `minor` here would batch a breaking upgrade of our own
        # upstream crates into a routine PR. Patch only; a 0.x minor is a
        # deliberate adoption, not a dependency refresh.
        update-types:
          - patch
        patterns: [...]
```

If you would rather these never be proposed automatically at all, an `ignore` entry with `update-types: ["version-update:semver-minor", "version-update:semver-major"]` on the six achieves that more explicitly, and reads as intent rather than as a side effect of the grouping. Either is fine by me; the first keeps genuinely-compatible patch bumps flowing, which I read as what Rob wants.

One caveat on my own claim: GitHub's reference does not document 0.x handling explicitly — it only says versions are assumed to be `major.minor.patch`, which implies literal component comparison. So the classification above is inference, not a quoted guarantee. It does not change the recommendation: if the inference is right, patch-only is necessary; if it is wrong and Dependabot already treats `0.9 → 0.10` as major, patch-only costs nothing.

The rest of the file — the security-update header, the exact-complement grouping, the paraseq pair — is unaffected and still good.

BenjaminDEMAILLE:
Applied: `combine-lab-upstream` is `update-types: [patch]`.

**Your caveat is now discharged.** You flagged the 0.x classification as inference rather than a quoted guarantee, so I measured it in `dependabot-core` instead. Constructing an `IgnoreCondition` per level for a dependency at 0.9.2 and reading back the ranges each level covers:

```
patch   > 0.9.3, < 0.10
minor   >= 0.10.a, < 1
major   >= 1.a
```

No 0.x special case anywhere in the resolution. 0.9.2 to 0.10.0 is `minor` to the grouping logic, exactly as you inferred, so `[minor, patch]` would have batched a breaking upgrade of our own upstream crates into a routine PR. Same probe on 0.5.0-pre.2: the patch level covers `> 0.5.0, < 0.6`, so the pre-release to release step is patch, which is the one movement this group should carry automatically.

I took the `update-types` form rather than the `ignore` form. `ignore` would also stop these from being *proposed*, and I read the intent as "propose them, do not batch them": a 0.x minor from piscem-rs still deserves a PR, just its own, where the diff and the decision are visible. If you meant the stronger reading, the ignore block is a two-line swap.

**One thing your comment raises that I have not changed, with a number attached.** You noted `cargo-minor` has the same exposure. It is not marginal here:

```
direct dependencies        55
  of those 0.x             34
  0.x, third-party         29   (ahash, bramble-rs, crossbeam*, hashbrown, indicatif, jiff, libc,
                                 mimalloc, needletail, noodles-*, rand*, rapidgzip-core, sha2,
                                 statrs, sux, thread-broker, tracing*, xxhash-rust, zstd, ...)
```

So more than half the direct dependencies are 0.x, and for every one of them a breaking bump reads as `minor` and would land in the weekly batch. Restricting `cargo-minor` to patch as well would be consistent, and would turn those into individual PRs rather than suppressing them, but it changes the weekly volume from roughly one PR to potentially several, which is the trade the grouping exists to avoid. The lab crates are the ones whose breaking changes have carried decisions this cycle, which is why I have done what you asked and no more. Say if you want the same treatment across the board and I will push it; noodles and rapidgzip-core in particular are 0.x crates where a minor bump is a real API change.

rob-p:
Pushed `120d3b83` applying @rob-p's policy for automated PRs. Answering your open question directly: yes, the same treatment across the board.

The policy, in three tiers:

| tier | what | how it arrives |
|---|---|---|
| routine | patch only | one grouped PR per ecosystem per week |
| explain | any minor, **including 0.x minors, which are breaking** | its own PR, own diff, own CI |
| decide | any major | its own PR, never batched |

**`cargo-minor` → `cargo-patch`, `update-types: [patch]`.** Your number was the argument: 34 of 55 direct dependencies are 0.x, 29 third-party, so for more than half the tree a breaking bump reads as `minor` and would have landed in the routine batch. The group name also no longer describes what it does. Not suppressed, just ungrouped — and `open-pull-requests-limit: 5` already bounds the queue, which is the answer to the volume worry that motivated grouping in the first place.

**`actions` gains `update-types: [minor, patch]`.** This one is a change to your stated rationale, not just its scope. "An action major is a workflow change, visible in the diff, and CI is its own test" is precisely the green-CI-equals-decision reasoning being ruled out here — `checkout` v4→v5 and `upload-artifact` v3→v4 both changed behaviour in ways a passing build would not catch. Majors now break out individually.

**`website-minor` deliberately keeps batching minors**, now with a comment saying why: it is the docs site, nothing it depends on can reach quantification, and tightening it buys noise rather than safety. Worth being explicit so it does not read as an oversight next to the two groups that were tightened.

**One thing I changed in the header that is worth your eye.** Rob's instruction is that *all* PRs target `develop`, never `master`. Every entry already does. But your security-update caveat had a rationale attached — "that is deliberate, a security fix belongs on the release branch" — and that is no longer the stated intent, so I replaced it with the constraint as it actually stands: the two settings cannot both be satisfied from this file, since dropping `target-branch` would drag every version update onto `master` as well. The only lever that puts security PRs on `develop` is making `develop` the repository default branch, which is a repo setting and is explicitly not assumed by the file.

Schema revalidated: 4 entries, all `target-branch: develop`, groups now `cargo-patch=[patch]`, `combine-lab-upstream=[patch]`, `actions=[minor,patch]`, `website-minor=[minor,patch]`, docker ungrouped.

Merge order unchanged: #1173 first, then this.

PR #1171, opened 2026-08-23
fix(quant): a record with no placement gets MAPQ 0, not 255

BenjaminDEMAILLE:
Note 1 from your #1151 review, which merged before I could apply it, so it comes as a follow-up rather than a response commit.

The `nh == 0` arm returned 255. No mapped record reaches it, since a fragment with no placements never enters the record loop, so the arm is only about record kinds that have no placement by construction: an unmapped read, once #1142 or its successor writes them. 255 there claims certainty about a placement that does not exist, and 0 is what samtools convention expects on an unmapped record.

Your reasoning for doing it now rather than then is the part I would underline: whoever adds unmapped output is unlikely to go looking for this function's edge case, and the failure would be a plausible-looking 255 rather than anything that breaks.

One line, its test line, and a row in the docs table so the `MAPQ` section covers the no-placement case rather than only `NH >= 1`.

Note 2 (the `#1150` contract paragraph reconciliation) is yours and I have left it alone.

### DIFF
--- a/crates/salmon-quant/src/mapping_record.rs
+++ b/crates/salmon-quant/src/mapping_record.rs
@@ -183,7 +183,14 @@ pub fn complement(base: u8) -> u8 {
 /// #1141).
 pub fn mapping_quality(nh: usize) -> u8 {
     match nh {
-        0 | 1 => 255,
+        // No placements is not a case a mapped record can reach (the writer
+        // skips a fragment with none), so this arm exists for the record kinds
+        // that have no placement by construction: an unmapped read, once such
+        // records are written. 255 there would claim certainty about a
+        // placement that does not exist; 0 is what samtools convention expects
+        // on an unmapped record.
+        0 => 0,
+        1 => 255,
         2 => 3,
         3..=4 => 1,
         _ => 0,
@@ -438,6 +445,7 @@ mod tests {
         assert_eq!(mapping_quality(3), 1);
         assert_eq!(mapping_quality(4), 1);
         assert_eq!(mapping_quality(5), 0);
-        assert_eq!(mapping_quality(0), 255);
+        // An unmapped record has no placement to be confident about.
+        assert_eq!(mapping_quality(0), 0);
     }
 }
--- a/website/src/content/docs/reference/output-formats.md
+++ b/website/src/content/docs/reference/output-formats.md
@@ -491,6 +491,7 @@ is what STAR reports, so `MAPQ` follows STAR's mapping:
 | 2 | 3 |
 | 3 or 4 | 1 |
 | 5 or more | 0 |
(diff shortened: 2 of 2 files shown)


rob-p (approved):
Approving. Verified on the branch: 343/0 workspace, and the unit test covers the changed arm.

I checked the unreachability claim rather than taking it, because the change is only safe if it is true. It is, with one wrinkle worth writing down: neither caller (`sam.rs:134`, `bam.rs:329`) guards against an empty `maps`, so `mapping_quality(0)` genuinely *is* called today — `mapq` is computed before the emit loop. What saves it is that the loop body never runs with no placements, so the value never reaches a record. The arm is live code whose result is currently discarded, not dead code.

That makes the fix slightly more worth doing than "unreachable edge case" suggests. The moment a record kind is emitted outside that loop — which is exactly what unmapped-read output does — the existing value would have been consumed silently. And 255 is the worst possible wrong answer here: `samtools view -q 255` is the idiom for "uniquely placed", so an unmapped record would have passed the filter that exists to select confident alignments.

The docs row is the right call too. The `MAPQ` table previously enumerated `NH >= 1` and simply had no row for the no-placement case, so a reader could reasonably have concluded it could not arise.

Note 2 (#1150's pseudobam contract paragraph) is indeed mine — that paragraph now needs to stop saying `MAPQ` is nominal and point at the STAR scale instead. I will fold it in.

rob-p:
Note 2 is already done, so nothing is pending on me — flagging it so you are not holding a follow-up for it.

The reconciliation happened during #1150's rebase, where your MAPQ change and the pseudobam contract paragraph collided directly. On `develop` today the contract bullet reads:

> - **`MAPQ`** reflects only the placement count, not alignment confidence — see below.

pointing at the `MAPQ` section with the STAR table, rather than the old "MAPQ is nominal; do not filter on it with `samtools view -q`" line. The two would otherwise have contradicted each other in the same document — the contract telling readers not to filter on `MAPQ`, the new section telling them `-q 255` means uniquely placed.

Your PR's docs row slots into that same table, which is now the single place `MAPQ` semantics are defined.

BenjaminDEMAILLE:
Good, thanks for closing the loop. Nothing was being held for it on my side, but I had assumed the reconciliation was still ahead rather than already on `develop`.

Confirmed the shape it landed in: the contract bullet now points at the section instead of contradicting it, and the STAR table is the single definition of the semantics. That is the right split, and better than what my PR body imagined (a MAPQ entry living inside the contract paragraph), because the contract stays a list of what the pseudobam does not promise while the table stays the place the one derived field is specified.

My row for the no-placement case slots into that table unchanged; rebased on the current `develop` to be sure, no conflict.

PR #1173, opened 2026-08-23
deps: declare paraseq-temp under its real name

BenjaminDEMAILLE:
Follow-up to the finding on #1170: the manifest renamed this dependency, and the rename made it invisible to Dependabot.

## The rename was never load-bearing

```toml
paraseq = { package = "paraseq-temp", version = "0.5.0-pre.2" }
```

`paraseq-temp`'s own manifest declares `[lib] name = "paraseq"`, so `use paraseq::...` resolves identically with or without the rename key. Not one line of code changes in this PR, and `Cargo.lock` is untouched: same crate, same version.

The comment above the entry says the rename exists so that `use paraseq::...` is unchanged. That reason does not hold, and the hazard the comment warns about (two paraseq copies either side of piscem-rs's API) is unrelated to the key: piscem-rs resolves `paraseq-temp` too, which is visible in `Cargo.lock`, and that is what keeps the two sides on one crate.

## What it cost

Dependabot's cargo parser drops a renamed manifest entry before building a dependency from it:

```ruby
# cargo/lib/dependabot/cargo/file_parser.rb, manifest_dependencies
next unless name == name_from_declaration(name, requirement)
```

`name` is `paraseq`, `name_from_declaration` returns `paraseq-temp`, so the entry is skipped. The crate then reaches the parser only through `Cargo.lock`, with no requirement attached, which makes it indirect and puts it outside the default `allow`.

Measured by running that parser over this workspace in the official updater image, before and after:

```
before   paraseq-temp   0.5.0-pre.2   TRANSITIVE (no manifest requirement)
after    paraseq-temp   0.5.0-pre.2   direct (1 req)
```

Everything else is unchanged in both runs (374 dependencies parsed either way; piscem-rs, libradicl, cf1-rs, sshash-lib, ksw2rs all direct as before).

## Why it matters beyond Dependabot

This is a pinned pre-release of a republished crate, waiting on noamteyssier/paraseq#78 to land upstream. It is exactly the dependency whose movement someone should be told about, and it was the one dependency no tool could see. The comment now records why it is not renamed, so the next person who finds `paraseq-temp = ...` uglier than `paraseq = ...` does not quietly undo it.

With this in, the `paraseq-temp` pattern in #1170's `combine-lab-upstream` group starts matching, which is the state that PR's description assumed. I will update its inline note once one of the two lands; whichever order suits you.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,11 +101,18 @@ divan = "0.1"
 # TEMPORARY: a republication of paraseq's released 0.5.0 plus the resizable
 # worker pool offered upstream as noamteyssier/paraseq#78 (fixed spawn set with
 # parking; worker count bounded by construction). `ThreadPool` is what lets the
-# mapping thread count change mid-run, which the broker requires. The library
-# target is still named `paraseq`, so `use paraseq::...` is unchanged. `paraseq`
-# is the official crate -- move back once the pool lands there. Must match what
+# mapping thread count change mid-run, which the broker requires. `paraseq` is
+# the official crate -- move back once the pool lands there. Must match what
 # piscem-rs resolves, or two paraseq copies end up either side of its API.
-paraseq = { package = "paraseq-temp", version = "0.5.0-pre.2" }
+#
+# Declared under its real name rather than renamed to `paraseq`: the rename was
+# never needed (the crate's library target is itself named `paraseq`, so
+# `use paraseq::...` works either way) and it made the dependency invisible to
+# tooling. Dependabot's cargo parser skips a renamed manifest entry outright
+# (`next unless name == name_from_declaration(...)`), so this crate reached it
+# only through Cargo.lock, as an indirect dependency no version update would
+# ever be raised for.
+paraseq-temp = "0.5.0-pre.2"
 
 # Workspace-wide lints. Member crates opt in with `[lints] workspace = true`.
 [workspace.lints.rust]
--- a/crates/salmon-quant/Cargo.toml
+++ b/crates/salmon-quant/Cargo.toml
@@ -17,7 +17,7 @@ salmon-eqclass = { workspace = true }
 salmon-infer = { workspace = true }
 salmon-rad = { workspace = true }
 piscem-rs = { workspace = true }
-paraseq = { workspace = true }
+paraseq-temp = { workspace = true }
 thread-broker = { workspace = true }
 rapidgzip-core = { workspace = true }
 anyhow = { workspace = true }


rob-p (approved):
Approving. This one deserved verification rather than agreement, because if the central claim were wrong every `use paraseq::` in the tree would stop resolving. It holds — checked directly against the vendored crate:

```
$CARGO_HOME/registry/src/.../paraseq-temp-0.5.0-pre.2/Cargo.toml
  [package] name = "paraseq-temp"
  [lib]     name = "paraseq"
```

So the library target really is `paraseq` regardless of the manifest key, and the rename bought nothing. Built the workspace on your branch to confirm rather than infer: **343/0**, zero errors, `use paraseq::{fastx, parallel::ThreadPool, prelude::*}` in `pool_resize_flush_safety.rs` untouched, and `Cargo.lock` still clean after a build — cargo did not consider anything to have changed.

I also checked the hazard the original comment warns about, since that is the failure that would actually hurt: the lock contains exactly one `paraseq-temp v0.5.0-pre.2`, and `cargo tree -i` shows piscem-rs resolving the same package. One copy, both sides of the API. That invariant is preserved and — importantly — the comment you kept still states it, so it stays checkable.

The parser evidence is the part I would not have found. Confirming it by running the real updater over this workspace and diffing `TRANSITIVE (no manifest requirement)` → `direct (1 req)`, with the other 373 dependencies unchanged in both runs, is a much stronger argument than reading the Ruby and reasoning about it. It also explains something otherwise invisible: a pinned pre-release of a republished crate, blocked on an upstream PR, was the single dependency no tooling could report on.

Two notes.

**Merge order matters slightly.** This should land before or with #1170. Until it does, that config's `paraseq-temp` pattern matches a dependency Dependabot classifies as indirect, so the `combine-lab-upstream` group silently would not cover it — the exact state your #1170 description assumes is already true. Landing this first makes that assumption real.

**Keep the `paraseq` spelling in #1170's pattern list anyway.** It costs nothing, and it is insurance for the day the pool lands in noemteyssier/paraseq and the manifest moves back to the real `paraseq` crate. At that point the entry stops being renamed and the group keeps matching without anyone remembering to edit it.

BenjaminDEMAILLE:
Thanks for verifying rather than agreeing, and specifically for checking the invariant rather than the claim I made: `cargo tree -i` showing piscem-rs on the same `paraseq-temp v0.5.0-pre.2` is the check that actually matters here. I had reasoned from `Cargo.lock` containing exactly one such package, which is weaker evidence for the same conclusion.

Both notes handled.

**Merge order**: agreed, this before or with #1170. Nothing on either branch is waiting on anything else.

**`paraseq` kept in the pattern list**: pushed to #1170. Your reason is better than mine for dropping it. I removed it because it matched nothing, which is the right default for a config line, but this one has a specific future in which it matches, and the alternative is relying on someone editing that file at exactly the moment they are busy moving back to the official crate. The comment now gives each spelling its own reason so the pair does not read as a duplicate someone should tidy.

On the parser evidence: the reason I ran it rather than reading it is that I had already been wrong once in this thread by reasoning from documentation, and the failure mode was invisible, which is the combination that makes reading insufficient. Worth keeping the harness in mind for the flip PR too, if any of its warn-and-ignore behaviour ends up depending on what Dependabot or another tool does with a manifest rather than on what we believe it does.

PR #1174, opened 2026-08-23
ci: auto-merge the Dependabot updates that are not decisions

BenjaminDEMAILLE:
Pairs with #1170. That PR groups the updates so they arrive as a few PRs a week; this one merges the ones that are not decisions, because a queue of green patch bumps waiting for a human is how a dependency policy quietly stops working.

## Policy

| | auto-merged |
|---|---|
| cargo `cargo-minor` group (patch/minor) | yes |
| npm `website-minor` group (patch/minor) | yes |
| `github-actions` group, patch/minor | yes |
| any group containing a **major** | no |
| `combine-lab-upstream` group | no |
| docker base image | no |
| anything not targeting `develop` | no |

The excluded PRs get a comment saying which rule applied, rather than sitting there looking forgotten.

Two of those deserve their reasoning stated:

- **`combine-lab-upstream`** is excluded because those bumps have carried decisions, not just numbers (the piscem-rs 0.9.2 "parked pool as a floor" commit, the paraseq-temp republication).
- **Anything not targeting `develop`** means security updates, which land on the default branch: a `target-branch` in `dependabot.yml` excludes an entry from security updates, as you noted on #1170. A security fix merging into the release branch unread is not something a workflow should decide.

## What I verified rather than copied

Most auto-merge examples in circulation are subtly stale, so:

- **`update-type` on a grouped PR** is the *highest* semver change in the group, not the first dependency's. Checked in `fetch-metadata`'s source rather than inferred: `UPDATE_TYPES_PRIORITY.find(semverLevel => semverLevels.has(semverLevel))`, with the priority array ordered major, minor, patch. That is what makes "one major anywhere holds the group" work.
- **`dependency-group`** comes from the commit trailer (`dependency-group:\s(?<name>\S*)`), so gating on the group name is reliable.
- **`fetch-metadata@v3.1.0`** is the current release. The examples everywhere, including some official docs pages, still say `@v2`.
- **actionlint** passes on the new file (the repo's existing shellcheck findings in `release.yml` are untouched and pre-existing).

## One prerequisite I cannot check from here, and it matters

`gh pr merge --auto` only *waits* if the PR is not already mergeable. **If no status checks are required on `develop`, an auto-merge request on a mergeable PR merges immediately, before CI finishes.** That would be the opposite of the intent.

My token cannot read this repository's merge settings (`allow_auto_merge` comes back null) or its branch protection (`/branches/develop/protection` returns 404, which is either "no protection" or "no access"). So before this is useful, two things need to be true on the repo side:

1. **Allow auto-merge** enabled in repository settings.
2. **Required status checks on `develop`**, at minimum the four `build-test` matrix jobs.

If you would rather not depend on branch protection, say so and I will switch the workflow to wait for CI itself (`gh pr checks --watch --fail-fast` before a plain `gh pr merge --squash`). That version is self-contained and needs no repo settings; it costs a runner sitting idle for the length of CI, which for a handful of PRs a week is nothing. I did not default to it because `--auto` is the mechanism GitHub built for this, and holding a runner for 20 minutes to reimplement it is the sort of thing that looks clever and ages badly.

Merge order: after #1170, or with it; on its own this workflow simply never fires, since without that config Dependabot only raises security PRs, which this deliberately does not touch.

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,0 +1,70 @@
+# Auto-merge the Dependabot updates that are not decisions.
+#
+# The grouping in .github/dependabot.yml exists so the routine movement arrives
+# as a few PRs a week rather than a dozen. This finishes that thought: a grouped
+# patch/minor bump with green CI does not need a human to press the button, and
+# a queue of them waiting for one is how a dependency policy quietly stops
+# working.
+#
+# What is deliberately NOT auto-merged, and why:
+#
+#   * majors, in any ecosystem. `update-type` on a grouped PR is the highest
+#     semver change in the group (dependabot/fetch-metadata resolves it against
+#     `['major', 'minor', 'patch']` in priority order), so one major anywhere in
+#     a group holds the whole group for review.
+#   * the `combine-lab-upstream` group. piscem-rs, paraseq-temp, libradicl and
+#     friends move with work happening in those repositories, and their bumps
+#     have carried deliberate floors. Those are decisions.
+#   * anything targeting a branch other than `develop`. Security updates land on
+#     the default branch (a `target-branch` in dependabot.yml excludes an entry
+#     from security updates), and a security fix going straight to the release
+#     branch unread is not something a workflow should decide.
+#   * the container base image, which is a toolchain choice.
+name: Dependabot auto-merge
+
+on: pull_request_target
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  auto-merge:
+    # `pull_request_target` runs in the base repository's context with a
+    # writable token, which is what Dependabot's own documentation prescribes
+    # for this: a workflow triggered by Dependabot gets a read-only token
+    # otherwise. Nothing here checks out or executes the PR's code, which is the
+    # condition that makes `pull_request_target` safe.
+    if: github.actor == 'dependabot[bot]' && github.event.pull_request.base.ref == 'develop'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Read what Dependabot is proposing
+        id: metadata
+        uses: dependabot/fetch-metadata@v3.1.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Enable auto-merge for routine updates
+        if: >-
+          steps.metadata.outputs.update-type != 'version-update:semver-major'
+          && steps.metadata.outputs.dependency-group != 'combine-lab-upstream'
+          && steps.metadata.outputs.package-ecosystem != 'docker'
+        run: gh pr merge --auto --squash "$PR_URL"
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Say why this one waits for a human
+        if: >-
+          steps.metadata.outputs.update-type == 'version-update:semver-major'
+          || steps.metadata.outputs.dependency-group == 'combine-lab-upstream'
+          || steps.metadata.outputs.package-ecosystem == 'docker'
+        run: |
+          gh pr comment "$PR_URL" --body "Not auto-merged: ${REASON}. Merge it by hand once you have looked at it."
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REASON: >-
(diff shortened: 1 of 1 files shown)


rob-p:
I have the access you don't, so I checked the two prerequisites. **Both are currently false**, and the combination is more awkward than either alone.

```
repos/COMBINE-lab/salmon        allow_auto_merge: false
repos/.../branches/develop/protection   404 "Branch not protected"
```

So as things stand:

- **`gh pr merge --auto` would fail outright**, not merge early — auto-merge is disabled at the repository level, so the API rejects the request. Every run of this workflow would error. Annoying, but fail-safe, which is the right direction to fail in.
- **The dangerous state is one settings toggle away.** `develop` has no protection at all, so there are no required status checks. If someone enables "Allow auto-merge" to make this workflow start working — the obvious response to seeing it fail — then `--auto` on an already-mergeable PR merges it **immediately, before CI runs**. Your caveat is exactly right, and the trap is that the fix for the visible symptom creates the invisible hazard.

That ordering makes the `--auto` version the wrong default here. It is not just "needs two settings"; it is "needs two settings enabled in the right order, where doing the intuitive one first is unsafe."

**I would take the variant you offered**: `gh pr checks --watch --fail-fast` followed by a plain `gh pr merge --squash`. Your reasoning against it — that reimplementing a platform mechanism looks clever and ages badly — is normally right, and I would agree if `--auto` worked here. But it makes the workflow self-contained: correct on today's repository settings, with no dependency on branch protection that a future settings change could silently invalidate. A runner idling through CI for a handful of PRs a week is not a cost worth optimising against correctness.

If @rob-p would rather enable branch protection on `develop` — required `build-test` matrix checks, which is worth having regardless of this PR — then `--auto` becomes the better mechanism and this PR is right as written. That is a repository-policy decision affecting every contributor and every PR, not something to adopt as a side effect of a dependency-automation change.

Worth noting the same gap explains something from this cycle: I have been checking CI green by hand before every merge, because nothing enforces it. Branch protection would be an improvement on its own merits — I just don't think it should arrive bundled with auto-merge.

The rest of the PR I have no concerns about. The policy table is right, excluding `combine-lab-upstream` matches where this cycle's real decisions actually landed, and excluding anything not targeting `develop` correctly keeps security updates out of an unread merge. Verifying that grouped-PR `update-type` is the *highest* semver level in the group — from `fetch-metadata`'s source rather than from an example — is the detail most copies of this workflow get wrong.

rob-p:
Closing this rather than merging it, after @rob-p's call on the automated-PR policy in #1170.

**The policy changed under it.** #1170 landed with patch-only grouping: minors are no longer batched, they arrive as individual ungrouped PRs. This workflow gates on `update-type != semver-major && dependency-group != combine-lab-upstream`, which is name-agnostic and so survived the group rename — but that generality is now the problem. An ungrouped minor satisfies both conditions (`update-type` is minor, `dependency-group` is empty), so it would be auto-merged.

That matters more than it sounds: **34 of the 55 direct dependencies are 0.x**, where minor *is* Cargo's breaking boundary. This workflow would therefore silently auto-merge breaking bumps of `noodles-*`, `rapidgzip-core`, `sux`, `hashbrown` — exactly the case #1170 was changed to prevent, and exactly what Rob ruled out: a breaking upgrade should be a decision, not something adopted because CI is green.

**The cost/benefit no longer works either.** After the policy change the auto-mergeable set is three grouped PRs a week (`cargo-patch`, `website-minor`, `actions`); the two categories anyone actually worries about — `combine-lab-upstream` and docker — this workflow already excluded. So it would save about three clicks a week on the least consequential PRs in the repo, in exchange for 70 lines of workflow, an `allow_auto_merge` repo setting, required status checks on `develop`, and a path that merges into `develop` with no human in the loop.

None of this is a criticism of the implementation. Your verification was the good kind — reading `fetch-metadata`'s source for the grouped-PR `update-type` priority rather than trusting the examples, catching that `@v3.1.0` is current when the docs still say `@v2`, and flagging the `--auto`-merges-immediately hazard yourself as a prerequisite you could not check. That last point is what made the risk legible in the first place, and it is why the settings never got flipped blind.

Two things from this PR outlive it and are being kept:

- **The `--auto` hazard.** `gh pr merge --auto` only waits if the PR is not already mergeable, so enabling "Allow auto-merge" without required status checks merges before CI. That is now on record here for whoever revisits this.
- **Branch protection is being considered on its own merits**, decoupled from auto-merge. Nothing currently stops a force-push to `develop`, its deletion, or a merge with red CI. That gap does not go away with this PR, and it should not have been framed as a prerequisite *for* it.

If the dependency queue ever does start silently stalling, the argument for reviving this gets much stronger — but narrowed to `update-type == version-update:semver-patch`, which is the only tier the policy treats as routine.

PR #1185, opened 2026-08-24
deps: adopt libradicl 0.18 + hashbrown 0.17 for 2.6.0

rob-p:
Per @rob-p's call in the pre-2.6.0 dependency scan: adopt libradicl 0.18 (COMBINE-lab upstream, a deliberate 0.x-minor decision).

0.18 requires `noodles ^0.115`, which the tree already resolves to `0.115.0`, so **no noodles bump is forced** (the coupling that usually accompanies a libradicl bump does not bite here).

Verified against the release contract, not assumed:

- **Builds clean.** Our full RAD API surface (`chunk`/`header`/`readers`/`writers`/`record`/`rad_types`) is source-compatible; no code changes needed.
- **`cargo test --workspace --release`: 355 passed, 0 failed** (2 ignored), including all RAD round-trip and output-contract tests.
- **Determinism intact and output unchanged.** `quant.sf` is byte-identical across `-p 1/16/64` at `--sigDigits 9` on the 26M-fragment RAD, and the hash (`2a86764f…`) is **the same one the libradicl-0.17 build produced** — so 0.18 alters no quantification output.
- fmt + `clippy --workspace --all-targets -D warnings` clean.

Diff is two files: `Cargo.toml` (the pin) and `Cargo.lock` (libradicl + libradicl-macros 0.17 → 0.18). This lands before the 2.6.0 version bump.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2815,7 +2815,7 @@ version = "2.5.1"
 dependencies = [
  "ahash",
  "divan",
- "hashbrown 0.15.5",
+ "hashbrown 0.17.1",
  "ksw2rs",
  "piscem-rs",
  "salmon-core",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1656,15 +1656,16 @@ dependencies = [
 
 [[package]]
 name = "libradicl"
-version = "0.17.0"
+version = "0.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7518c918544220380ff5ff33b243771a8757872150dd280e6cc9355cddf3043b"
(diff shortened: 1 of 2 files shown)


rob-p:
Added **hashbrown 0.15 → 0.17** (the other out-of-range direct dep from the scan). Used directly only in salmon-map (align.rs, score.rs); 0.17 is source-compatible with no changes.

Re-validated both bumps together:
- build + fmt + clippy -D warnings clean
- `cargo test --workspace --release`: **355 passed, 0 failed**
- determinism intact and **output byte-identical** (`quant.sf` `2a86764f…` across `-p 1/16/64`, same as before either bump)

The cascade dry-run (`bump_and_publish.sh 2.6.0 --dry-run`) validated all 10 crates package cleanly in dependency order.

PR #1197, opened 2026-08-30
Release 2.7.0

rob-p:
Brings master up to develop for the 2.7.0 release: the sshash v6 / index-format-v2 change (#1196) plus the dependabot updates that accumulated on develop. Release notes: docs/release-notes-2.7.0.md. The version bump, tag, crate publishes, and dist binaries follow on master via scripts/bump_and_publish.sh 2.7.0 --publish.

https://claude.ai/code/session_01NknHxdevZi243yQ5sKaffu

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -63,7 +63,7 @@ jobs:
       # tag yet — the merge job assembles the tagged multi-arch manifest).
       - name: Build and push by digest
         id: build
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v7
         with:
           context: .
           platforms: ${{ matrix.platform }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -61,4 +61,4 @@ jobs:
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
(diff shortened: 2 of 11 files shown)


BenjaminDEMAILLE:
@rob-p Question on the 2.7.0 scope: should complete alignment-mode BAM support land in 2.7?

Per `docs/cli-flag-status.md`, `-a/--alignments` and `-t/--targets` are implemented, but the posterior-sampled BAM output path (`-s/--sampleOut`, `-u/--sampleUnaligned`, `--writeQualities`) is still accepted-and-warned only (`crates/salmon-cli/src/main.rs:1311`), and `--eqclasses` input mode is likewise unimplemented.

Do you want the sampled-BAM writer finished for a 2.7.x point release, or deferred to 2.8?

rob-p:
Hi @BenjaminDEMAILLE: The 2.7.0 point release was centered around the updated (and backward incompatible) index bump (based on the new sshash v6.0.0 design).

Feature additions could land as either a point release (2.7.x) or could wait until 2.8.0. I think one thing that would be useful to do is to understand what features users are still actively taking advantage of (so that carrying the extra implementation complexity with us would be worthwhile), versus which should be slated for removal.

PR #1198, opened 2026-08-30
refactor(cli): move the global allocator into its own module, document the gates for changing it

BenjaminDEMAILLE:
## What this is now

Scoped down to the half you said you'd take: the allocator placement, plus the
documentation of what would gate changing it. No new dependency, no new feature,
no behavior change.

* `#[global_allocator]` moves out of `main.rs` into
  `crates/salmon-cli/src/global_alloc.rs`, alongside a `NAME` constant reporting
  which allocator a build actually got. The binary logs it at `debug`.
* `examples/cache_bench.rs` pulls that same file in via `#[path]`, so the
  microbenchmark and the real binary cannot disagree about what is in play.
* mimalloc stays the default; `sysalloc` stays the diagnostic escape hatch it
  already was on `master`. `Cargo.lock` is back to `master`'s.

Dropped from the previous revision: the `jemalloc` and `snmalloc` features and
their dependencies.

## The documentation half

`docs/allocator-choice.md` now leads with the negative result rather than the
microbenchmark, and states the bar rather than implying the features were their
own justification. It records:

1. **The gates.** An end-to-end win on a real quant run (faster wall clock
   without a non-trivial RSS increase, or the same wall clock with a
   considerable RSS decrease), and evidence that the win holds across the
   input, output and option surfaces (selective alignment and sketch, with and
   without decoys, with and without mapping output, across a realistic `-p`
   range). Plus the supporting discipline: interleaved arms, spread reported
   with the median, `quant.sf` hashed to prove the arms compute the same thing.
   It notes #1126 and #1041 as the same bar applied elsewhere.

2. **The measurement behind the current default**, which was already in the
   file and which agrees with you: a four-way end-to-end round (GENCODE v50,
   10M pairs, `-p 16`, 5 interleaved reps) where system malloc is +38% on the
   mapping phase, jemalloc is ~7% behind mimalloc, and snmalloc is inside the
   noise floor while costing ~0.4 GiB more peak RSS. Nothing there justifies a
   switch. The doc now also says explicitly what that round does *not*
   establish under gate 2: one machine, one reference, one simulated read set,
   one option configuration.

3. **The two traps**, since both produced confident wrong answers first: the
   single-threaded microbenchmark ranks the four allocators in an order the real
   workload contradicts, and a cold-page-cache first round produced the exact
   reverse of the settled ranking.

## Verification

`cargo fmt --all --check` clean. `cargo build -p salmon-cli --all-targets`
clean, and clean again with `--features sysalloc`.
`cargo clippy -p salmon-cli --all-targets` clean (the only remaining note is the
pre-existing future-incompat warning from the third-party `proc-macro-error2`).

### DIFF
--- a/crates/salmon-cli/Cargo.toml
+++ b/crates/salmon-cli/Cargo.toml
@@ -28,7 +28,9 @@ tracing-subscriber = { workspace = true }
 indicatif = "0.18"
 
 [features]
-# Diagnostic: build with the system allocator instead of mimalloc.
+# Diagnostic: build with the system allocator instead of mimalloc. See
+# src/global_alloc.rs for how the choice is made and docs/allocator-choice.md
+# for what would have to be measured to change it.
 sysalloc = []
 
 [dev-dependencies]
--- a/crates/salmon-cli/examples/cache_bench.rs
+++ b/crates/salmon-cli/examples/cache_bench.rs
@@ -3,10 +3,16 @@
 //! AHashMap cap-256 + Vec cap-64) vs `clear()` (reuse), under whichever global
 //! allocator the binary is built with. Run with the default (mimalloc) and with
 //! `--features sysalloc` to contrast.
+//!
+//! Note what this does *not* measure: it allocates and frees on one thread,
+//! which is every allocator's fast path and not what the quant hot path does.
+//! Its ranking does not survive contact with a real run; see
+//! `docs/allocator-choice.md`.
 
-#[cfg(not(feature = "sysalloc"))]
-#[global_allocator]
-static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
+// Shared with the real binary so the two can never disagree about which
+// allocator a given set of feature flags selects.
+#[path = "../src/global_alloc.rs"]
+mod global_alloc;
 
 use piscem_rs::mapping::cache::MappingCache;
 use piscem_rs::mapping::sketch_hit_simple::SketchHitInfoSimple;
@@ -14,11 +20,7 @@ use std::hint::black_box;
 use std::time::Instant;
 
 fn main() {
-    let alloc_name = if cfg!(feature = "sysalloc") {
-        "system"
-    } else {
-        "mimalloc"
-    };
+    let alloc_name = global_alloc::NAME;
     // Enough iterations that per-op timing is not dominated by clock resolution.
     let n: u64 = 30_000_000;
 
--- a/crates/salmon-cli/src/global_alloc.rs
+++ b/crates/salmon-cli/src/global_alloc.rs
@@ -1,0 +1,31 @@
+//! Which global allocator the `salmon` binary is built with.
+//!
+//! The quant hot path is highly multithreaded and allocation-heavy (per-read
+//! scratch buffers allocated on a parser thread and freed on a worker thread),
+//! so the allocator is a real, measurable choice rather than a detail. mimalloc
+//! is the default, and `docs/allocator-choice.md` records both the measurement
+//! behind that default and what a candidate replacement would have to show
+//! before it is worth carrying:
+//!
+//! | build flags           | allocator     |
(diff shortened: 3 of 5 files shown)


rob-p:
Hi @BenjaminDEMAILLE,

  I already closed the alternative feature allocator request because it lacked end-to-end measurement benefit.  There are two specific parts to this PR. One seems like it may be a good idea, but I'd prefer to avoid the other.
  
## What seems interesting

Perhaps `main.rs` isn't the ideal place for the allocator placement. If we can move the allocator into `crates/salmon-cli/src/global_alloc.rs` to allow potentially easier swaps in the future, this seems like a clear refactoring win, so I'm happy to take it.

## What I'd rather not take right now

We already rejected the system allocator and jemalloc in the past. Having those, even as non-default features, complicates the build and increases the surface area that needs to be tested without breaking.  Currently, there doesn't seem to be evidence that something apart from `mimalloc` registers a tangible, meaningful, benefit in end-to-end runs. In order to consider either adding an additional allocator as a feature, or perhaps swapping the default, what we'd need is measured evidence that it (1) beats mimalloc on end-to-end performance (i.e. being faster without increasing RSS in a non-trivial way, or being just as fast wile decreasing RSS considerably) and (2) that this benefit holds across different input/output and option surfaces.  Otherwise, this is complexity without measured benefit.


If you want to edit this PR to be just the first part of moving the allocator placement and add documentation about what gates would be required to consider a different allocator, I'm happy to take that PR.

BenjaminDEMAILLE:
Trimmed to the first part, as scoped.

**What's left**: the move to `crates/salmon-cli/src/global_alloc.rs`, the `NAME` constant so a build labels its own numbers, the `#[path]` include that keeps `cache_bench.rs` and the binary in agreement, and the docs. `jemalloc`, `snmalloc` and their dependencies are gone; `Cargo.lock` is back to `master`'s.

**One scoping question**: I kept `sysalloc`, since it is already a feature on `master` (`crates/salmon-cli/Cargo.toml`, `main.rs`) rather than something this PR introduces, and it is useful for isolating an allocator-specific bug. Your comment reads as being about the features this PR adds, but if you want the pre-existing escape hatch gone too, say so and I will drop it in a follow-up rather than smuggle the removal in here.

**On the evidence**: the end-to-end round you're asking for was already in `docs/allocator-choice.md`, and it reaches your conclusion, not the one the PR description implied. Four-way, interleaved, 5 reps, 10M pairs at `-p 16`: system malloc +38% on the mapping phase, jemalloc ~7% behind mimalloc, snmalloc inside the noise floor (the two rounds disagree in direction) while costing ~0.4 GiB more peak RSS. Leading the description with the single-threaded microbenchmark table was the wrong emphasis on my part: that microbenchmark ranks the four in an order the real workload contradicts, which is now recorded in the doc as a trap rather than a result.

So the doc now states the negative result first, then your gates as the admission criteria for revisiting it: (1) an end-to-end win on a real quant run, faster without a non-trivial RSS increase or equal-speed with a considerable RSS decrease; (2) evidence the win holds across input/output/option surfaces. It cites #1126 and #1041 as the same bar applied elsewhere, and it is explicit that the round above would not itself satisfy gate 2 had a candidate looked good: one machine, one reference, one simulated read set, one option configuration.

**Follow-up, not in this PR**: the reason that round is one configuration is that we have no harness for the full-quant case. `scripts/bench_em_rad.sh` has the right discipline (interleaved arms, `/usr/bin/time` capturing wall/CPU/peak RSS, `quant.sf` hashed per row) but replays a RAD, so it excludes mapping, which is the phase an allocator actually moves. I am building the full-quant equivalent (FASTQ in, sweep over threads x selective-alignment/sketch x decoys x mapping-output). It would answer gate 2 for an allocator and it is most of what #1041 needs too. Happy to propose it as its own PR if you would want it in-tree; otherwise it stays on my side and only its numbers show up.

